PR #16, opened 2019-12-06
Live Patching updates

Loquacity:
https://github.com/SUSE/spacewalk/issues/10202


Loquacity:
Added requirement for spacewalk-utils. Ref Server is down, but when it's up I'll review the SLES 15 instructions and replace the CLI steps with web ui.

PR #19, opened 2019-12-09
Use API for mass SP migration

keichwa:
new section: SP mass migration

closes https://github.com/SUSE/spacewalk/issues/6684

### DIFF
--- a/modules/upgrade/pages/client-migration.adoc
+++ b/modules/upgrade/pages/client-migration.adoc
@@ -164,16 +164,21 @@ After the migration is complete, schedule a reboot of the client from the {produ
 
 If you want to migrate a large number of clients to the next SP version, you can use {productname} API calls.
 
-You need to know to which migration target you want to migrate.
+
+
+.Procedure: SP Mass Migration
+. You need to know to which migration target you want to migrate.
 List available migration targets:
++
 ----
 spacecmd api -- system.listMigrationTargets -A 1000010001
 ----
++
 Create a list of system IDs you want to migrate.
-
-For each system ID, call [systemitem]``listMigrationTarget`` and check that it is an available target.
-If the system ID is an available target, call [systemitem]``system.scheduleSPMigration``.
-If it is not an available target, skip the system.
+. For each system ID, call [systemitem]``listMigrationTarget`` and check that it is an available target.
++
+* If the system ID is an available target, call [systemitem]``system.scheduleSPMigration``.
+* If it is not an available target, skip the system.
 
 Adapt this template for your environment:
 


modules/upgrade/pages/client-migration.adoc
```diff
@@ -157,3 +157,36 @@ If the system status reports that the migration has failed, but also shows that
 After the migration is complete, schedule a reboot of the client from the {productname} Server.
 
 ////
+
+
+
+== SP Mass Migration
+
+If you want to migrate many clients to the next SP version, {suse} recommends to use {productname} API calls.
```
Loquacity:
```suggestion
If you want to migrate a large number of clients to the next SP version, you can use {productname} API calls.
```

modules/upgrade/pages/client-migration.adoc
```diff
@@ -157,3 +157,36 @@ If the system status reports that the migration has failed, but also shows that
...
+
+You need to know to which migration target you want to migrate.
```
Loquacity: I think this should probably be a procedure.

modules/upgrade/pages/client-migration.adoc
```diff
@@ -157,3 +157,36 @@ If the system status reports that the migration has failed, but also shows that
...
+List available migration targets:
+----
+spacecmd api -- system.listMigrationTargets -A 1000010001
+----
+Create a list of system IDs you want to migrate.
+
+Now we can iterate over the system IDs and call [systemitem]``listMigrationTarget`` for every system and check if the wanted target is part of the available targets.
```
Loquacity:
```suggestion
For each system ID, call [systemitem]``listMigrationTarget`` and check that it is an available target.
```

modules/upgrade/pages/client-migration.adoc
```diff
@@ -157,3 +157,36 @@ If the system status reports that the migration has failed, but also shows that
...
+If yes, we call [systemitem]``system.scheduleSPMigration``, otherwise we skip the system.
```
Loquacity:
```suggestion
If the system ID is an available target, call [systemitem]``system.scheduleSPMigration``.
If it is not an available target, skip the system.
```

modules/upgrade/pages/client-migration.adoc
```diff
@@ -157,3 +157,36 @@ If the system status reports that the migration has failed, but also shows that
...
+
+Here is a template that you can adapt for your environment:
```
Loquacity:
```suggestion
Adapt this template for your environment:
```

modules/upgrade/pages/client-migration.adoc
keichwa: ok, trying to do my best ;)

PR #24, opened 2019-12-11
remove ubuntu supported features

Loquacity:
Resolves task 1 of https://github.com/SUSE/spacewalk/issues/10239

### DIFF
--- a/modules/client-configuration/pages/clients-ubuntu.adoc
+++ b/modules/client-configuration/pages/clients-ubuntu.adoc
@@ -12,8 +12,7 @@ Canonical does not endorse or support {susemgr}.
 ====
 
 Bootstrapping is supported for starting {ubuntu} clients and performing initial state runs such as setting repositories and performing profile updates.
-
-However, the root user on {ubuntu} is disabled by default, so in order to use bootstrapping, you will require an existing user with [command]``sudo`` privileges for Python.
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/clients-ubuntu.adoc
```diff
@@ -11,32 +11,10 @@ Traditional clients are not supported.
...
-* Bootstrapping
-* Synchronizing [systemitem]``.deb`` channels
-* Assigning [systemitem]``.deb`` channels to clients
-* GPG signing [systemitem]``.deb`` repositories
-* Information displayed in System details pages
-* Package install, update, and remove
-* Package install using [systemitem]``Package States``
-* Configuration and state channels
-
 Bootstrapping is supported for starting {ubuntu} clients and performing initial state runs such as setting repositories and performing profile updates.
```
juliogonzalez: Why is this staying?

Loquacity:
> Driving by I saw other tiny issues:
> 
> * `in oder` is a weasel word. Better avoid it.

Removed

> * the `[ Create Repository ]` procedure is rather terse. Just reading it (w/o seeing the UI) is confusing because as a reader I cannot imagine that you are referencing different things with `[ Create Repository ]`.

This will get addressed when we go over the procedure.

modules/client-configuration/pages/clients-ubuntu.adoc
Loquacity: Because it explains why you need to enable the root user. I've just put those two paras together into one so it makes more sense now.

PR #25, opened 2019-12-12
add feature new deployment with PXE + autoinstallation

mcalmer:
We described in the feature list if we support "re-deployment" but it seems it would be better
to also add "new deployment" to make it more obvious for customers.

This PR add "System deployment (PXE/...)" to the list of features.

Additionally fixed autoyast vs. Kickstart and moved Ubuntu/debian to explicite "not supported" as we should not really have questionmarks in that list.


mcalmer:
@Loquacity that was fast. In this case I will just merge it.

Loquacity:
> @Loquacity that was fast. In this case I will just merge it.

I aim to please ;)

PR #29, opened 2019-12-16
Add note about Ubuntu root access 

Loquacity:
Copy from https://github.com/SUSE/doc-susemanager/pull/909


modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -47,6 +47,34 @@ For more information and instructions, see xref:administration:tshoot-registercl
 
 
 
+== Disabling the FQDNS grain
```
keichwa: strange artifacts here, too.

PR #20, opened 2019-12-09
Update Administration Guide - Monitoring 

cavalheiro:
A partial rewrite and content review of the Monitoring chapter in the Admin guide, according to the latest product features.

PR #32, opened 2019-12-17
monitoring: explain how to compose DSN string

moio:
Feel free to copy edit!

Context from this change comes from internal mailing list:

http://mailman.suse.de/mlarch/SuSE/suse-manager/2019/suse-manager.2019.12/msg00173.html

### DIFF
--- a/modules/administration/pages/monitoring.adoc
+++ b/modules/administration/pages/monitoring.adoc
@@ -217,7 +217,7 @@ image::monitoring_enable_services.png[scaledwidth=80%]
 
 . In the {susemgr} {webui}, open the details page of the system to be monitored, and navigate to the menu:Formulas[Prometheus Exporters] tab.
 . Check the [guimenu]``Enabled`` checkbox for both the Node and the Postgres Exporter.
-. In the [guimenu]``Postgres Exporter`` section, in the [guimenu]``Data Source Namer`` field, enter the path to your data source (for example, [systemitem]``postgresql://user:passwd@localhost:5432/database?sslmode=disable``).
+. In the [guimenu]``Postgres Exporter`` section, in the [guimenu]``Data Source Namer`` field, enter the the string [systemitem]``postgresql://USER:PASSWORD@localhost:5432/database?sslmode=disable``, replacing ``USER`` and ``PASSWORD`` with the values of the ``db_user`` and ``db_password`` from the ``/etc/rhn/rhn.conf`` file. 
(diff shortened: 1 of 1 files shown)


modules/administration/pages/monitoring.adoc
```diff
@@ -217,7 +217,7 @@ image::monitoring_enable_services.png[scaledwidth=80%]
 
 . In the {susemgr} {webui}, open the details page of the system to be monitored, and navigate to the menu:Formulas[Prometheus Exporters] tab.
 . Check the [guimenu]``Enabled`` checkbox for both the Node and the Postgres Exporter.
-. In the [guimenu]``Postgres Exporter`` section, in the [guimenu]``Data Source Namer`` field, enter the path to your data source (for example, [systemitem]``postgresql://user:passwd@localhost:5432/database?sslmode=disable``).
+. In the [guimenu]``Postgres Exporter`` section, in the [guimenu]``Data Source Namer`` field, enter the the string [systemitem]``postgresql://USER:PASSWORD@localhost:5432/database?sslmode=disable``, replacing ``USER`` and ``PASSWORD`` with the values of the ``db_user`` and ``db_password`` from the ``/etc/rhn/rhn.conf`` file.
```
keichwa:
```suggestion
. In the [guimenu]``Postgres Exporter`` section, in the [guimenu]``Data Source Namer`` field, enter the string [systemitem]``postgresql://USER:PASSWORD@localhost:5432/database?sslmode=disable``, replacing ``USER`` and ``PASSWORD`` with the values of ``db_user`` and ``db_password`` from the ``/etc/rhn/rhn.conf`` file. 
```

Loquacity:
This has conflicted with @cavalheiro 's Monitoring edits from PR#20 https://github.com/uyuni-project/uyuni-docs/pull/20 Easy enough to resolve technically, but I'm not sure what the correct information is.

cavalheiro:
The updated version of the monitoring admin guide does not include specific arguments for each exporter, as those are explained in detail in the linked upstream documentation. If we want to include PostgreSQL exporter specifics, it would make sense to include it for others as well. Is this needed?

Loquacity:
OK, in that case, I recommend dropping this PR.

modules/administration/pages/monitoring.adoc
moio: I agree we should always use PostgreSQL, which is the correct name

PR #51, opened 2020-01-24
Add note about monitoring via network boundaries

witekest:
SUSE/spacewalk#9413

### DIFF
--- a/modules/administration/pages/monitoring.adoc
+++ b/modules/administration/pages/monitoring.adoc
@@ -259,7 +259,9 @@ Shipped exporters use these ports:
 * PostgreSQL exporter: 9187
 * Apache exporter: 9117
 
-Another approach is to deploy a Prometheus instance in the network local to the exporters and configure federation. This will allow the main monitoring server to scrape time series from the local Prometheus instance. In this setup only Prometheus API port (9090) has to be opened.
+Alternatively, you can deploy a Prometheus instance in the exporters' local network, and configure federation. 
+This allows the main monitoring server to scrape the time series from the local Prometheus instance. 
+If you use this method, you only need to open the Prometheus API port (9090).
 
 For more information on Prometheus federation, see https://prometheus.io/docs/prometheus/latest/federation/.
 


modules/administration/pages/monitoring.adoc
```diff
@@ -246,3 +246,21 @@ image::monitoring_configure_formula.png[scaledwidth=80%]
 ====
 Monitoring formulas can also be configured for System Groups, by applying the same configuration used for individual systems inside the corresponding group.
 ====
+
+
+
+=== Network boundaries
```
Loquacity:
```suggestion
=== Network Boundaries
```

modules/administration/pages/monitoring.adoc
```diff
@@ -246,3 +246,21 @@ image::monitoring_configure_formula.png[scaledwidth=80%]
...
+
+As mentioned above scrape targets have to be accessible from Prometheus server. In case of clients installed on cloud instances it is often the simplest and  the best approach to add required ports into security group and grant access to the monitoring server. Following ports are used by the shipped exporters:
```
Loquacity:
```suggestion
Scrape targets must be accessible from the Prometheus server. 
For clients installed on cloud instances, the simplest approach is to add the required ports into a security group that has access to the monitoring server. 
Shipped exporters use these ports: 
```

modules/administration/pages/monitoring.adoc
```diff
@@ -246,3 +246,21 @@ image::monitoring_configure_formula.png[scaledwidth=80%]
...
+
+* Node exporter: 9100
+* PostgreSQL exporter: 9187
+* Apache exporter: 9117
+
+Another approach is to deploy a Prometheus instance in the network local to the exporters and configure federation. This will allow the main monitoring server to scrape time series from the local Prometheus instance. In this setup only Prometheus API port (9090) has to be opened.
```
Loquacity:
```suggestion
Alternatively, you can deploy a Prometheus instance in the exporters' local network, and configure federation. 
This allows the main monitoring server to scrape the time series from the local Prometheus instance. 
If you use this method, you only need to open the Prometheus API port (9090).
```

modules/administration/pages/monitoring.adoc
```diff
@@ -246,3 +246,21 @@ image::monitoring_configure_formula.png[scaledwidth=80%]
...
+
+For more information on about Prometheus federation, see https://prometheus.io/docs/prometheus/latest/federation/.
+
+Finally, in some cases it is reasonable to proxy the the requests through the network boundary. Tools like PushProx deploy a proxy and a client on both sides of the network barrier and allow seamless Promethes operation across NAT or other similar network topologies.
```
Loquacity:
```suggestion
You can also proxy requests through the network boundary. 
Tools like PushProx deploy a proxy and a client on both sides of the network barrier and allow Prometheus to work across network topologies such as NAT.
```

PR #53, opened 2020-01-28
Remove auditlog-keeper and icinga references

juliogonzalez:
`modules/administration/pages/icinga.adoc` had this:
```
////
I sent a message to the suse manager ML 2019-05-16 asking about Icinga and have received exactly nothing back. I think we can probably silently drop it from the 4.0 nav. It'll still exist in the older docs, and I won't delete the actual file just yet, in case someone yells at us. LKB - 2019-05-20
////
```
It's dropped from 4.0 and nobody complained. For 4.1 we don't even offer the icinga integration anymore, so I think it's time to let it go.

It also contained references to auditlog-keeper, that is removed as part of this PR (we will not have it at 4.1).


modules/installation/pages/ports.adoc
```diff
@@ -72,8 +72,6 @@ In most cases, you will not need to adjust these ports.
 | Port number | Notes
```
keichwa:
```suggestion
| Port Number | Notes
```
Drive by edit.  I'd capitalize column titles (title style).  But we can do this later.

PR #56, opened 2020-01-29
Ref Guide - Bare Metal Systems

Loquacity:

A note on terminology:

We use "bare metal" through the docs, but the UI uses "bare-metal". So when using the term in prose, I've gone with a space, but when referring directly to UI elements, I've left the hyphen in. I checked out the ngram for the US English corpus, and I think we probably should petition for the UI team to update to the two-word form:

 
image

Additionally, even though the UI uses "bare-metal system" I've gone with "client" so it's consistent with the rest of the docs.

Finally, this patch includes an updated screenshot, but I have chosen not to include it in the text.


keichwa:
:+1: and if possible we want to avoid the term "bare metal" term altogether.  I'm not sure what we agreed to use instead (unprovisioned or physical or a combination of both).  I often do not know what we are talking about when we use this term ;)

Loquacity:
>  and if possible we want to avoid the term "bare metal" term altogether. I'm not sure what we agreed to use instead (unprovisioned or physical or a combination of both). I often do not know what we are talking about when we use this term ;)

Hrm, that's a different proposition. I'm pretty familiar with the term, but I can see the confusion (especially if bare metal systems are actually VMs, so neither bare nor metal :wink: ). It's used all the way through the UI, do you think that using a different term to mean the same thing might be more confusing, not less, though?

keichwa:

> Hrm, that's a different proposition. I'm pretty familiar with the term,
> but I can see the confusion (especially if bare metal systems are
> actually VMs, so neither bare nor metal :wink: ). It's used all the way
> through the UI, do you think that using a different term to mean the
> same thing might be more confusing, not less, though?

In the end, either term is fine for me.  IIRC, someone some time ago
started with getting rid of the "bare metal".  For example, the bare
metal system type is gone from "Systems > System List".  Now there we
have:

"Unprovisioned Systems"

And then, there are "Physical Systems" and "Without System Type".

-- 
Product & Solution Documentation   Maxfeldstraße 5
                                   90409 Nürnberg, Germany

Loquacity:
> the bare metal system type is gone from "Systems > System List". Now there we have: "Unprovisioned Systems" And then, there are "Physical Systems" and "Without System Type".

:woman_facepalming:

PR #57, opened 2020-01-29
Ref Guide - Bootstrap

Loquacity:

### DIFF
--- a/modules/client-configuration/pages/registration-bootstrap.adoc
+++ b/modules/client-configuration/pages/registration-bootstrap.adoc
@@ -56,7 +56,7 @@ For details on each setting, see xref:reference:admin:bootstrap-script.adoc[].
 Alternatively, you can access the bootstrap script over HTTPS.
 Replace ``example.com`` with the host name of your {productname} Server:
 ----
-https://EXAMPLE.COM/pub/bootstrap/bootstrap.sh
+https://<example.com>/pub/bootstrap/bootstrap.sh
 ----
 
 


modules/client-configuration/pages/registration-bootstrap.adoc
```diff
@@ -43,42 +43,34 @@ Attempting to register openSUSE Leap 15 or SLE 15 systems using Python 2 boot
...
 . In the [guimenu]``SUSE Manager Configuration - Bootstrap`` dialog, uncheck the [guimenu]``Bootstrap using Salt`` checkbox if you are installing a traditional client.
 For Salt clients, leave it checked.
-Use default settings and click the btn:[Update] button.
-+
+. The required fields are pre-populated with values derived from previous installation steps.
+For details on each setting, see xref:reference:admin:bootstrap-script.adoc[].
+. Click btn:[Update] create the script.
+. The bootstrap script is generated and stored on the server in the [path]``/srv/www/htdocs/pub/bootstrap`` directory.
+Alternatively, you can access the bootstrap script over HTTPS.
+Replace ``example.com`` with the host name of your {productname} Server:
```
keichwa: I see you downcased this.  Nevertheless, this
```suggestion
Replace ``EXAMPLE.COM`` with the host name of your {productname} Server:
```
Even if you don't accept it we should talk about it.  Just to avoid changing those things back and forth ;)

modules/client-configuration/pages/registration-bootstrap.adoc
```diff
@@ -43,42 +43,34 @@ Attempting to register openSUSE Leap 15 or SLE 15 systems using Python 2 boot
...
 
-[WARNING]
-.Using SSL
+[IMPORTANT]
 ====
-Unchecking [guimenu]``Enable SSL`` in the {webui} or setting `USING_SSL=0` in the bootstrap script is not recommended.
-If you disable SSL nevertheless you will need to manage custom CA certificates to be able to run the registration process successfully.
+Do not disable SSL in your bootstrap script.
+Ensure that [guimenu]``Enable SSL`` is checked in the {webui}, or that the setting `USING_SSL=1` exists in the bootstrap script.
+If you disable SSL, the registration process requires custom SSL certificates.
```
keichwa: Maybe, adding an xref to the section about creating custom SSL certificates here?  IIRC, you recently work on such a section :)

modules/reference/pages/admin/bootstrap-script.adoc
```diff
@@ -1,52 +1,33 @@
 [[s3-sattools-config-bootstrap]]
 = Bootstrap Script
 
-The menu:Manager Configuration[Bootstrap Script] page allows you to generate a bootstrap script that registers the client systems with {productname} and disconnects them from the remote {scc}.
+In the menu:Manager Configuration[Bootstrap Script] section you can generate a custom bootstrap script.
+Bootstrap scripts are used to register clients with {productname}.
+The generated script will be placed in [path]``/srv/www/htdocs/pub/bootstrap/`` on your {productname} Server.
+
+
```
keichwa: Can we avoid such table markup overhead? I also don't like table layout if there are other possibilities.  I'd prefer definition lists such as:

**Option**
Description (multi-line text)
_Default:_ xyz or `value`

**Option2**
Description2 (multi-line text)
_Default:_ xyz or `value`

modules/client-configuration/pages/registration-bootstrap.adoc
Loquacity: Ah yeah, I was uncertain what we'd agreed. I'll take it to email.

modules/client-configuration/pages/registration-bootstrap.adoc
Loquacity: Good idea.

modules/reference/pages/admin/bootstrap-script.adoc
Loquacity: So this is what we were discussing on email about standards for reference material. If we're abiding by DITA standards, then it should be in table format. I know tables are difficult in adoc (and I hate making them as much as anyone else, especially after the feature matrix nastiness last year), but I think they have better readability over long deflists.

keichwa: Usually readability is worse.  At the first glance, they might look nice... But often there are serious white-space issues (for example, column for "description" too small because there are some insanely long option names).  As long as there is only one sorting critearium (option name) table layout is not the best markup solution.

Loquacity: I think I need to find some research on that. Scanning deflists can definitely be harder than scanning tables, especially when we're talking about a series of config options or something.

keichwa: ok, but for now we can go for tables.

modules/client-configuration/pages/registration-bootstrap.adoc
Loquacity: So far, the overwhelming consensus is to use angle brackets. I'll update accordingly.

modules/reference/pages/admin/bootstrap-script.adoc
Loquacity: yep, will let you know what I find.

PR #62, opened 2020-01-30
Ref Guide - Manager Config Orgs

Loquacity:

### DIFF
--- a/modules/reference/pages/admin/manager-config-organizations.adoc
+++ b/modules/reference/pages/admin/manager-config-organizations.adoc
@@ -1,6 +1,6 @@
 [[s3-sattools-config-orgs]]
 = Organizations
 
-The menu:Admin[Manager Configuration >Organizations] section page contains details about organizations in {productname}, and provides links to create and configure organizations and users.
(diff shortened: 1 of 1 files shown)


modules/reference/pages/admin/manager-config-organizations.adoc
```diff
@@ -1,6 +1,6 @@
 [[s3-sattools-config-orgs]]
 = Organizations
 
-The menu:Manager Configuration[Organizations] page contains details about the organizations feature of {productname}, and links for creating and configuring organizations.
+The menu:Admin[Manager Configuration >Organizations] section page contains details about organizations in {productname}, and provides links to create and configure organizations and users.
```
keichwa:
```suggestion
The menu:Admin[Manager Configuration > Organizations] section page contains details about organizations in {productname}, and provides links to create and configure organizations and users.
```
Do you actually want to call these things "section page"?  Several names are possible: tab(?), dialog, page, section, maybe also "section page".  I think it would be good if we agree to use only one or two of them.

Loquacity: No, that's a typo. I was going with "section".

PR #65, opened 2020-01-31
Ref Guide - Restart

Loquacity:

### DIFF
--- a/modules/reference/pages/admin/restart.adoc
+++ b/modules/reference/pages/admin/restart.adoc
@@ -3,4 +3,4 @@
 
 The menu:Admin[Manager Configuration > Restart] section allows you to restart {productname}.
 You will need to do this after making configuration changes.
-It will take four to five minutes for {productname} to become available again after a restart.
+It will take some time for {productname} to become available again after a restart.


modules/reference/pages/admin/restart.adoc
```diff
@@ -1,9 +1,6 @@
...
 
-The menu:Manager Configuration[Restart] page comprises the final step in configuring {productname}.
-
-image::admin_configuration_restart.png[scaledwidth=80%]
-
-Click the btn:[Restart] button to restart {productname} and incorporate all of the configuration options added on the previous screens.
-It will take between four and five minutes for a restart to finish.
+The menu:Admin[Manager Configuration > Restart] section allows you to restart {productname}.
+You will need to do this after making configuration changes.
+It will take four to five minutes for {productname} to become available again after a restart.
```
keichwa: Wondering whether we can take the "four to five minutes" duration for granted?

Loquacity: Yeah, I wondered about that too. Perhaps we should say something like "some time" instead?

keichwa:
On 2020-02-05 01:01, Lana Brindley wrote:

> Yeah, I wondered about that too. Perhaps we should say something like
> "some time" instead?

Yes, that's better.  Or "on standard hardware several minutes".

-- 
Karl Eichwalder

PR #77, opened 2020-02-03
Autogenerate PDF navigation from the HTML Antora list

paususe:
Autogenerate the PDF nav files from the HTML nav files so that we do not need to maintain them in sync by hand.

Output is slightly different from the previous one, mostly due to wrong use of
Antora list levels IMHO. E. g. in the Upgrade guide, every section (update, migrate from 3.x, proxy update, etc) is a subsection of the introduction, which makes no sense at all. 

**Side-by-side review of the PDF books generated by 'master' and by this PR is required. I can clearly see some errors in the depth in the current HTML navigation, which when transposed to the PDFs, are very easy to notice by looking at TOCs.** E. g.

image

### DIFF
--- a/modules/reference/pages/help/api-sample-scripts.adoc
+++ b/modules/reference/pages/help/api-sample-scripts.adoc
@@ -2,4 +2,4 @@
 
 The [guimenu]``API Sample Scripts`` menu item gives you some sample scripts for using the {productname} API in several common languages.
 
-For more sample scripts, see https://github.com/spacewalkproject/spacewalk/.
+For more sample scripts, see https://github.com/uyuni-project/uyuni/.


paususe:
@keichwa yes, it's possible to do that. In fact, it's possible to also avoid duplicating SUSE Manager and Uyuni.

jcayouette:
I will have a look in the morning! Thanks this at first glance looks like a nice upgrade :)

paususe:
Ready for review

Loquacity:
I'm going to defer to @jcayouette to make a decision here. My current understanding is that we were waiting for an Antora solution, but if that solution is too far away or too difficult, then I'm happy to consider an alternative. My main concern is about @jcayouette's time and bandwidth, considering we have a lot on our plates right now, and I want to ensure we avoid double-handling the problem if we can.

paususe:
@Loquacity
IMHO there's no doubt about this PR from the technical point of view: it's way better than what we have, and when antora will have native PDF support, we can move to that.

However, this PR reveals our section levels (depths) are wrong in many cases in the HTML, so the moment this PR is merged, all the PDF books will look different. The reason we didn't notice the wrong section depths in the HTML index before is we used different section levels for PDF and HTML. That requires review and fixing before this PR is merged. For that, you don't need @jcayouette

Loquacity:
I'm not suggesting that it's not technically accurate, or that we don't require it. I'm more concerned with bandwidth: @jcayouette needs to be able to implement and maintain it from a tools perspective, and the entire team needs to be able to make the changes.

paususe:
@Loquacity
This PR is the full implementation, @jcayouette only needs to review it.

The review and fix of the HTML indices is not part of this PR and can be done by you or @keichwa and committed after this PR.

jcayouette:
> @Loquacity
> IMHO there's no doubt about this PR from the technical point of view: it's way better than what we have, and when antora will have native PDF support, we can move to that.
> 
> However, this PR reveals our section levels (depths) are wrong in many cases in the HTML, so the moment this PR is merged, all the PDF books will look different. The reason we didn't notice the wrong section depths in the HTML index before is we used different section levels for PDF and HTML. That requires review and fixing before this PR is merged. For that, you don't need @jcayouette

Please compare the offsets in both versions.
Our Antora output is correct at master along with our PDF lists, however when the pdfs are creating using your sed commands they are offset by +1 and the "sections **" are ignored because they are still in the form of antora list items.

**This**:
```
* Installation Guide
include::modules/installation/pages/install-intro.adoc[leveloffset=+2]
** Requirements (This is being ignored when the PDF is created even though its a section)
include::modules/installation/pages/general-requirements.adoc[leveloffset=+3]
include::modules/installation/pages/hardware-requirements.adoc[leveloffset=+3]
endif::[]
```

**Should Be This** (see the leveloffset):
```
* Installation Guide
include::modules/installation/pages/install-intro.adoc[leveloffset=+1]
** Requirements
include::modules/installation/pages/general-requirements.adoc[leveloffset=+2]
include::modules/installation/pages/hardware-requirements.adoc[leveloffset=+2]
endif::[]
```

Also I think the lists are broken in your produced PDF list because the ifdef is not closed around the list content. You may be handling this in your code perhaps you can clarify here.

```
// PDF Installation Guide
ifdef::backend-pdf[]
= Installation Guide: {productname} {productnumber}
include::./branding/pdf/entities.adoc[]
:toclevels: 3
:doctype: book

// License
include::modules/installation/pages/common_gfdl1.2_i.adoc[leveloffset=+1]

endif::[]
```

Installation guide @master: 
- https://gist.github.com/jcayouette/f680c4b4041c4904dae88db4d9aa2897

Installation guide @paususe :

- https://gist.github.com/jcayouette/4a89e95d5772c59947baaeef42066324

paususe:
I have fixed the part where I was ignoring the AsciiDoc titles (i. e. sections with no xref), now it works too.

With that and a few adjustments to cope with inconsistencies (e.g. repeating the guide name at the top), the output of this PR is equal or better (some pages were missing in the PDF index, they are now in) as it was before.

The reference guide PDF index is different, though. In general, I'd say the new PDF index autogenerated from the Antora HTML index is better than the old one, as the manually-maintained PDF index had the wrong depth for some sections.

jcayouette:
@paususe Testing your updates locally on my system then I will add my review! Thanks

keichwa:
Joseph Cayouette <notifications@github.com> writes:

> LGTM, couple of minor issues I will fix in an upcoming PR. @paususe and
> I have tested output of both Antora and the new Asciidoctor-pdf indexes
> he creates using SED.

Yes, sounds good—thank you all!

-- 
Product & Solution Documentation   Maxfeldstraße 5
                                   90409 Nürnberg, Germany

PR #91, opened 2020-02-17
Editing new Subscription Matcher Chapter - Admin Guide

Loquacity:
https://github.com/SUSE/spacewalk/issues/10588
https://github.com/uyuni-project/uyuni-docs/pull/81

### DIFF
--- a/modules/administration/pages/subscription-matching.adoc
+++ b/modules/administration/pages/subscription-matching.adoc
@@ -22,7 +22,7 @@ You can also download the reports in .csv format, or access them from that comma
 
 By default, the subscription matcher runs daily, at midnight.
 To change this, navigate to menu:Admin[Task Schedules] and click ``gatherer-matcher-default``.
-Change the schedule as required, and click btn:``Update Schedule``.
+Change the schedule as required, and click btn:[Update Schedule].
 
 Because the report can only match current clients with current subscriptions, you might find that the matches change over time.
 The same client will not always match the same subscription.
@@ -49,7 +49,7 @@ When you have created a pin, the subscription matcher favors matching a specific
 
 However, the matcher will not always respect a pin.
 It depends on the subscription being available, and whether or not the subscription can be applied to the client.
-Additionally, pins will be ignored if they result in a match that violates the terms and conditions of the subscription, or if the matcher has determined a more accurate match.
+Additionally, pins will be ignored if they result in a match that violates the terms and conditions of the subscription, or if the matcher detects a more accurate match if the pin is ignored.
 
 To add a new pin, click btn:[Add a Pin], and select the client to pin.
 
--- a/modules/reference/pages/audit/audit-subscription-matching.adoc
+++ b/modules/reference/pages/audit/audit-subscription-matching.adoc
@@ -5,7 +5,6 @@ The menu:Audit[Subscription Matching] section provides reports that match your 
 Subscription matching reports provide information about clients that do not have a subscription, and subscription start and end dates.
 
 
-
 [[subscription-matching-options]]
 [cols="1,1", options="header"]
 .Subscription Matching Options
@@ -25,7 +24,6 @@ Table entries are highlighted if they are due to expire within three months.
 Table entries that have already expired are shown in grayscale.
 
 
-
 For messages relating to subscription matching, navigate to the [guimenu]``Messages`` tab.
 
 [[subscription-matching-status]]
@@ -41,7 +39,6 @@ For Linux-based hosts using ``libvirt``, check that the host is registered, and 
 |===
 
 
-
 To pin clients to a particular subscription, navigate to the [guimenu]``Pins`` tab.
 
 [[pin-status]]
@@ -53,8 +50,6 @@ To pin clients to a particular subscription, navigate to the [guimenu]``Pins`` t
 | Not satisfied           | The client was not successfully matched with a subscription.
 | Pending next run        | Waiting for the next matcher run.
 |===
-
-
 
 
 For more information about subscription matching, see xref:administration:subscription-matching.adoc[].


Loquacity:
@mcalmer can you please review and let me know if I've misunderstood anything? Happy to have a call if you like :)

modules/reference/pages/audit/audit-subscription-matching.adoc
```diff
@@ -4,4 +4,56 @@
...
+
+
+For messages relating to subscription matching, navigate to the [guimenu]``Messages`` tab.
+
+[[subscription-matching-status]]
+[cols="1,1,1", options="header"]
+.Subscription Matching Statuses
+|===
+| Status                  | Description | Action
+| Unsupported Part Number | The detected part number is unknown or unsupported. | Check the part number, and that the client has been matched to the correct subscription.
```
mcalmer: The action for unsupported part numbers is to call support and open a Service Request. In this case we need to add the partnumber to the product.

modules/reference/pages/audit/audit-subscription-matching.adoc
```diff
@@ -4,4 +4,56 @@
...
+| Physical Guest          | A client is reporting as virtual, but could be a physical client. | Check the client hardware data.
+| Guest with Unknown Host | A virtual client has an unknown host. | The client could be a physical client, not virtual.
```
mcalmer: Action should be to check the virtual host manager and configure it correctly to report the host and its guests.
In case the host is a linux system with libvirt, check, that the host is registered as well and has the virtual host system type  (checkbox in the system details)

modules/administration/pages/subscription-matching.adoc
```diff
@@ -1,153 +1,46 @@
...
-Guest with Unknown Host::
-Virtual guest has unknown host, assuming it is a physical system
+If the subscription matcher is not automatically matching a particular client with the correct subscription, you can manually pin them.
+When you have created a pin, the subscription matcher favors matching a specific subscription with a given system or group of systems.
 
-Unknown CPU Count::
-System has an unknown number of sockets, assuming 16.
-You can try fixing this by scheduling hardware refresh for affected system.
+However, the matcher will not always respect a pin.
+It depends on the subscription being available, and whether or not the subscription can be applied to the client.
```
mcalmer: I would add something like this:

A pin will be ignored when it violates the terms and conditions of the matched product, or the algorithm detects a better matching result when ignoring the pin.

mcalmer: There is a small typo. Duplicate 'a'

modules/administration/pages/subscription-matching.adoc
```diff
@@ -1,153 +1,61 @@
 [[subscription-matching]]
 = Subscription Matching
```
keichwa:
```suggestion
```

modules/administration/pages/subscription-matching.adoc
```diff
@@ -1,153 +1,61 @@
...
-{productname} automatically generates up-to-date nightly status reports by matching your {suse} subscriptions with all your registered systems.
-These reports are stored in [path]``/var/lib/spacewalk/subscription-matcher`` and provided in CSV format.
-These CSV files may be opened with a spreadsheet application such as LibreOffice Calc.
-
-If you want to schedule these reports to be produced at different times, or at a certain frequency or schedule a one time completion you can perform this task by editing the Taskomatic settings for the gatherer-matcher located in the {webui} at menu:Main Menu[Admin > Task Schedules > gatherer-matcher-default].
-
-image::audit_subscription_matching_gatherer_matcher_default.png[scaledwidth=80%]
+By default, the subscription matcher runs daily, at midnight.
+To change this, navigate to menu:Admin[Task Schedules] and click ``gatherer-matcher-default``.
+Change the schedule as required, and click btn:``Update Schedule``.
```
keichwa: I'm not sure whether this button markup is also valid asciidoc decoration.  I think our default markup is btn:[xxx].
Broken by design.

modules/administration/pages/subscription-matching.adoc
```diff
@@ -1,153 +1,61 @@
...
 
-image::add-pin.png[scaledwidth=80%]
 
-Within the menu:Subscriptions Available for Selected System[] window click the menu:Save Pin[] button to raise priority for subscription use on the selected system.
+If the subscription matcher is not automatically matching a particular client with the correct subscription, you can manually pin them.
+When you have created a pin, the subscription matcher favors matching a specific subscription with a given system or group of systems.
 
+However, the matcher will not always respect a pin.
+It depends on the subscription being available, and whether or not the subscription can be applied to the client.
+Additionally, pins will be ignored if they result in a match that violates the terms and conditions of the subscription, or if the matcher has determined a a more accurate match.
```
keichwa:
```suggestion
Additionally, pins will be ignored if they result in a match that violates the terms and conditions of the subscription, or if the matcher has determined a more accurate match.
```

modules/reference/pages/audit/audit-subscription-matching.adoc
```diff
@@ -4,4 +4,57 @@
 The menu:Audit[Subscription Matching] section provides reports that match your currently installed clients to your existing product subscriptions.
 Subscription matching reports provide information about clients that do not have a subscription, and subscription start and end dates.
 
+
```
keichwa: I think you can remove empty lines here and later on.

modules/administration/pages/subscription-matching.adoc
Loquacity: Oh, that's just me getting confused. You're right :)

modules/administration/pages/subscription-matching.adoc
Loquacity: Notes on the wording here: I'm trying to avoid 'algorithm' because there's a lot of confusion about what exactly an algorithm is or does (non-technical people seem to consider it as something closer to AI/ML than we do, and I don't think we want to give that impression), so I've stuck with "the matcher" here instead. On "better matching result" I've gone with "a more accurate match", because "better" is really subjective: what makes it better? "More accurate" seems like it's at least a bit more specific than "better". Let me know what you think :)

mcalmer: I am fine with 'more accurate'

PR #103, opened 2020-02-24
Ref Guide - Images

Loquacity:
https://github.com/SUSE/spacewalk/issues/10593

### DIFF
--- a/modules/reference/pages/images/images-image-list.adoc
+++ b/modules/reference/pages/images/images-image-list.adoc
@@ -4,7 +4,10 @@
 
 The menu:Images[Image List] section displays your current operating system images.
 
-Click btn:[Import] to import a new image.
+Click btn:[Import] to import a new Docker image.
+You can only import new images created from a Docker image using this mechanism.
+To import images based on Kiwi instead, see xref:administration:image-management.adoc[].
+
 Click btn:[Refresh] to update the list.
 
 Perform bulk actions by checking images in the list.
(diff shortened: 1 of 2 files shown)


Loquacity:
> I've the feeling that we sometimes lose info. The description seems to cover only the bare UI. If images are actually built and in, more options are available. Esp. you could click the name of an image and see details. I think the same is valid for stores.

I'm being as careful as I can to make sure that everything is captured. Usually it gets shorter just because of being a bit more economical with words/tables, etc. The big problem I have had with this section is that I don't have access to any images or anything to try anything out, so I'm relying on what was already there. Which was, unfortunately, not much. I suspect it's been a problem for some time with this section ;) Possibly need to create an issue to work directly with an SME and improve this section (as well as the related Images section in the Admin Guide).

modules/reference/pages/images/images-image-list.adoc
```diff
@@ -1,33 +1,26 @@
...
+[[image-list-columns]]
+[cols="1,1", options="header"]
+.Image List Columns
+|===
+| Column               | Description
+| Name                 | Name of the image.
+| Version and Revision | Version and revision of the image.
+| Updates              | Any updates that are currently available for the image.
+| Patches and Packages | Any patches or packages that are currently available for the image.
+| Build                | The time and date the image was built.
```
aaannz: Unless there is some change in uyuni, this does not inform about when the image was built, but rather if it is build, scheduled, building or failed.

modules/reference/pages/images/images-image-list.adoc
```diff
@@ -1,33 +1,26 @@
 [[ref.webui.images.list]]
 = Image List
 
-If you click menu:Main Menu[Images > Images], an overview listing of your images appears.
-Several columns provide information about each image:
 
+The menu:Images[Image List] section displays your current operating system images.
 
-image::images_list_images.png[scaledwidth=80%]
+Click btn:[Import] to import a new image.
```
aaannz: Importing of Kiwi images is not supported. It is valid only for Docker based images. Shouldn't it be mentioned here?

aaannz:
Didn't you forgot to push your changes? I don't see anything changed in the source.

Loquacity:
> Didn't you forget to push your changes? I don't see anything changed in the source.

:woman_facepalming:

PR #106, opened 2020-02-26
Ref Guide - Navigation

Loquacity:
https://github.com/SUSE/spacewalk/issues/10594

Move navigation guidance to Install


modules/installation/pages/webui-setup.adoc
```diff
@@ -33,3 +33,126 @@ The [guimenu]``System Groups`` pane shows any system groups you have created, an
 
 The [guimenu]``Recently Registered Systems`` pane shows all clients registered in the past thirty days.
 Click the name of a client in the list to be taken to the menu:Systems[System Details] section for that client.
+
+
+
+== Web Interface Navigation
```
keichwa: I'd rate this as reference documentation.  All the boring stuff is reference, IMO.  Maybe, it's a corner case.

Loquacity: Yeah, I thought a lot about this. It *is* reference info, but there's no UI element to hang it off, so the only way you even see it is if you go to the top level of the Ref Guide on the docs site, which doesn't seem particularly useful for readers. Since I created the Web UI setup chapter in the install guide, I liked the idea of moving it there where it's a bit more relevant. They're likely to see it at the moment that they're first installing/using the Web UI, which is when they will need the most help with it.

PR #112, opened 2020-02-29
Enable public cloud quick start guide

paususe:
Adjust filenames, add Makefile macros and targets and add to the Antora index


paususe:
@jcayouette could you please merge this? (4.1 material)

PR #114, opened 2020-03-03
Fix for for spacewalk-utils split

juliogonzalez:
Reason: https://github.com/uyuni-project/uyuni/pull/1950 (SUSE Manager card at https://github.com/SUSE/spacewalk/issues/9405)

`spacewalk-common-channels` will remain part of `spacewalk-utils` and will be now L3 supported.

### DIFF
--- a/modules/administration/pages/public-cloud-azure.adoc
+++ b/modules/administration/pages/public-cloud-azure.adoc
@@ -94,11 +94,13 @@ For more information on setting DNS zones, see the Azure documentation.
 
 Ensure that your {productname} Server is registered with {scc}.
 
+// I wonder why we do even need spacecmd
 When your server is registered, install these extra packages:
 ----
-zypper -n in spacecmd spacewalk-utils mlocate sysstat
+zypper -n in spacecmd mlocate sysstat
 ----
 // spacecmd will be installed by default next time
+// ^ How is so?
 
 Apply the latest updates and reboot the server:
 ----
--- a/modules/client-configuration/pages/clients-centos.adoc
+++ b/modules/client-configuration/pages/clients-centos.adoc
@@ -11,12 +11,6 @@ You are responsible for arranging access to {centos} base media repositories and
 
 ifeval::[{suma-content} == true]
 
-[IMPORTANT]
-====
-{suse} does not provide support for {centos} operating systems.
-{productname} allows you to manage {centos} clients, but support is not provided.
-====
-
 endif::[]
 
 
@@ -47,15 +41,8 @@ systemctl restart taskomatic
 
 == Channel and Repository Management
 
-The [package]``spacewalk-utils`` package contains a number of command line tools required for client administration, including the [command]``spacewalk-common-channels`` tool.
-
 
 ifeval::[{suma-content} == true]
-
-[IMPORTANT]
-====
-{suse} only provides support for [command]``spacewalk-clone-by-date`` and [command]``spacewalk-manage-channel-lifecycle`` tools.
-====
 
 endif::[]
 
--- a/modules/client-configuration/pages/clients-ubuntu.adoc
+++ b/modules/client-configuration/pages/clients-ubuntu.adoc
@@ -146,14 +146,8 @@ endif::[]
 
 
 ifeval::[{suma-content} == true]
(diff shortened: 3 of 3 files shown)


modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -11,10 +11,10 @@ You are responsible for arranging access to {centos} base media repositories and
 
 ifeval::[{suma-content} == true]
 
-[IMPORTANT]
+[WARNING]
 ====
-{suse} does not provide support for {centos} operating systems.
-{productname} allows you to manage {centos} clients, but support is not provided.
+The ``spacewalk-utils-extras`` package is not supported by {suse}, including the ``spacewalk-common-channels`` tool.
+You will not be supported if your system becomes inoperable by using these tools.
```
juliogonzalez: I copied this from the Ubuntu client tools as it's what we have there.

Question is: SUSE Manager release notes says this: https://www.suse.com/releasenotes/x86_64/SUSE-MANAGER/4.0/#_spacewalk_utils

So as you can see, 3 scripts (now part of `spacewalk-utils`) are now fully supported (meaning L3 supported), while the others are L1 supported.

However we always said at the doc they were not supported. Is it the case? @paususe?

modules/administration/pages/public-cloud-azure.adoc
```diff
@@ -94,6 +94,8 @@ For more information on setting DNS zones, see the Azure documentation.
 
 Ensure that your {productname} Server is registered with {scc}.
 
+// I fail to se why we need spacewalk-utils!!! What script from this package are we
```
juliogonzalez: @jcayouette git blame says you added this. Why?

juliogonzalez:
@Loquacity requesting re-review as the whole story is now different (the script is now L3 supported and part of `spacewalk-utils` so we don't need the warnings anymore

modules/administration/pages/public-cloud-azure.adoc
Loquacity: Looks like it could just be a bit of copypasta?

Loquacity:
@juliogonzalez are we good to merge this now?

juliogonzalez:
Almost. Let me merge the PR with the code first, just in case, and then I will take care of this one.

modules/administration/pages/public-cloud-azure.adoc
juliogonzalez: What should we do about this?

juliogonzalez: Ok, let's do the following. Let's keep this PR open until @jcayouette is back tomorrow, so he can merge this and then package.

jcayouette: Ya its copy pasta!

juliogonzalez: Ack, but don't resolve yet, as I need to remove it, and you need to review again that I am not breaking anything.

modules/administration/pages/public-cloud-azure.adoc
```diff
@@ -94,11 +94,13 @@ For more information on setting DNS zones, see the Azure documentation.
 
 Ensure that your {productname} Server is registered with {scc}.
 
+// I wonder why we do even need spacecmd
```
juliogonzalez: Ping @Loquacity I wonder if you shouldn't create a card to review this `.adoc`.

modules/administration/pages/public-cloud-azure.adoc
juliogonzalez: I found some more problems.

I think doc squad should really review this whole file and see what makes sense. I don't neither see why we need `spacecmd` or the other packages (others maybe because of the Azure images?)

I will not do this as part of this PR. For `spacewalk-utils` split, we are ready.

jcayouette: I will create another issue to track the other changes.

modules/administration/pages/public-cloud-azure.adoc
Loquacity: Yes please :)

PR #124, opened 2020-03-06
Recurring Action Documentation

parlt91:
Documentation for the Recurring Action feature.

This addresses: #116 

Signed-off-by: Pascal Arlt <parlt@suse.com>

PR #128, opened 2020-03-10
Parlt91/recurring action documentation

Loquacity:
This edits the content from https://github.com/uyuni-project/uyuni-docs/pull/124, sets up the navigation for the Admin Guide, and moves the reference content to a new section in the Ref Guide.

### DIFF
--- a/modules/reference/pages/users/user-details.adoc
+++ b/modules/reference/pages/users/user-details.adoc
@@ -50,7 +50,7 @@ For instructions to deactivate your own account, see xref:reference:home/user-ac
 
 [NOTE]
 ====
-If you delete a user account, any schedules that user created will also be deleted.
+If you delete a user account, any recurring actions that user created will also be deleted.
 ====
 
 


modules/administration/pages/recurring-actions.adoc
```diff
@@ -0,0 +1,55 @@
...
+. Choose the frequency of the recurring action:
++
+* [guimenu]``Hourly:`` Type the minute of each hour.
+For example, [parameter]``15`` will run the action at fifteen minutes past every hour.
+* [guimenu]``Daily:`` Select the time of each day.
+For example, [parameter]``01:00`` will run the action at 0100 every day, in the timezone of the {productname} Server.
+* [guimenu]``Weekly:`` Select the day of the week and the time of the day, to execute the action every week at the specified time.
+* [guimenu]``Monthly:`` Select the day of the month and the time of the day, to execute the action every month at the specified time.
+* [guimenu]``Custom Quartz format:`` For more detailed options, enter a custom quartz string.
+For example, to run a recurring action at 0215 every Saturday of every month, enter:
```
keichwa: I think this is Sunday.  My CRONTAB(5) says:
```
day of week    0-7 (0 or 7 is Sunday, or use names)
```
```suggestion
For example, to run a recurring action at 0215 every Sunday of every month, enter:
```

modules/administration/pages/recurring-actions.adoc
```diff
@@ -0,0 +1,55 @@
...
+. OPTIONAL: Toggle the [guimenu]``Test mode`` switch on to run the schedule in test mode.
+. Click btn:[Create Schedule] to save, and see the complete list of existing schedules.
+
+
+
+== Recurring Actions in Organizations
+
+
+Organization Administrators can set and edit recurring actions for all clients in the organization.
+Navigate to menu:Home[MyOrganization > Recurring States] to see all recurring actions that apply to the entire organization.
```
keichwa:
```suggestion
Navigate to menu:Home[My Organization > Recurring States] to see all recurring actions that apply to the entire organization.
```

modules/administration/pages/recurring-actions.adoc
parlt91: `0 15 2 ? * 7` is: `Every Saturday at 02:15`
So it's right as it is right now

modules/reference/pages/users/user-details.adoc
```diff
@@ -48,6 +48,10 @@ To delete a user:
 
 For instructions to deactivate your own account, see xref:reference:home/user-account-account-deactivation.adoc[].
 
+[NOTE]
+====
+If you delete a user account, any schedules that user created will also be deleted.
```
hustodemon: I would avoid the term "schedules", as it is not clear which schedules we mean here. I'd suggest:
- either using "recurring actions" term,
- or referencing the "recurring-actions" section in the ref guide.
- (alternatively: both :smile:  )

hustodemon:
Left one comment, the rest looks great to me!

modules/reference/pages/users/user-details.adoc
Loquacity:
```suggestion
If you delete a user account, any recurring actions that user created will also be deleted.
```

I don't think we need to reference the recurring actions section here, because we're not talking about setting them up, or changing them. Which is also why I moved it into this section, and took it out of the recurring actions section.

PR #130, opened 2020-03-10
Replace proxy #10521 ke

keichwa:
Clean up Replace Proxy description for SUMA and Uyuni.

### DIFF
--- a/modules/installation/pages/proxy-setup.adoc
+++ b/modules/installation/pages/proxy-setup.adoc
@@ -248,7 +248,8 @@ For more information about reactivation keys, see xref:reference:systems/system-
 [IMPORTANT]
 ====
 During the installation of the proxy, clients will not be able to reach the {productname} Server.
-After a proxy system has been deleted from the systems list, all clients connected to this proxy will be (incorrectly) listed as `directly connected` to the {productname} Server.
+After you have a deleted a proxy, the systems list can be temporarily incorrect.
+All clients that were previously connected to the proxy will show as being directly connected to the server instead.
 After the first successful operation on a client, such as execution of a remote command or installation of a package or patch, this information will automatically be corrected.
 This may take some hours.
 ====


modules/installation/pages/proxy-setup.adoc
```diff
@@ -207,45 +207,36 @@ filename: "pxelinux.0"
...
 
-A {susemgrproxy} is dumb in that it does not contain any information about the clients that are connected to it.
-A {susemgrproxy} can therefore be replaced by a new one.
-Naturally, the replacement proxy must have the same name and IP address as its predecessor.
+A proxy does not contain any information about the clients that are connected to it.
+Therefore, a proxy can be replaced by a new one at any time.
+The replacement proxy must have the same name and IP address as its predecessor.
 
-In order to replace a {susemgrproxy} and keeping the clients registered to the proxy leave the old proxy in {productname}.
+Leave the old proxy in place while you prepare the replacement.
```
mantel: Maybe it should be clarified what "leave in place" means. Of course it should be shut down because the new proxy will have the same IP address as the old one and having two machines with the same IP is not really fun :)

modules/installation/pages/proxy-setup.adoc
```diff
@@ -207,45 +207,36 @@ filename: "pxelinux.0"
...
-// ** Copy the activation keys from the previous server.
-// Of course, it is always better to re-create the keys.
+Consider copying important or manually created data to a central place that can also be accessed by the new proxy.
 . Shut down the proxy.
 . Install a new {susemgrproxy}{nbsp}{productnumber}, following xref:install-proxy-unified.adoc[Proxy Installation].
 . In the {productname} {webui} select the newly installed {susemgrproxy} and delete it from the systems list.
-[[step.at.proxy.migration3.replace.react]]
 . In the {webui}, create a reactivation key for the old proxy system: On the System Details tab of the old proxy click [guimenu]``Reactivation``.
-Then click [guimenu]``Generate New Key``, and remember it (write it on a piece of paper or copy it to the clipboard).
+Click [guimenu]``Generate New Key``, and make a note of the new key, as you will need it later.
```
mantel: In general you should be able to use the same key you also used for the previous proxy. There is no general need for a new key.

modules/installation/pages/proxy-setup.adoc
```diff
@@ -207,45 +207,36 @@ filename: "pxelinux.0"
...
+.Procedure: Replacing a Proxy and Keeping the Clients Registered
 . Before starting the actual migration procedure, save the data from the old proxy, if needed.
-Consider copying important data to a central place that can also be accessed by the new proxy.
-// REMARK: 2019-08-23, ke: this is unclear
-// ** Copy the scripts that are still needed.
-// ** Copy the activation keys from the previous server.
-// Of course, it is always better to re-create the keys.
+Consider copying important or manually created data to a central place that can also be accessed by the new proxy.
 . Shut down the proxy.
 . Install a new {susemgrproxy}{nbsp}{productnumber}, following xref:install-proxy-unified.adoc[Proxy Installation].
```
juliogonzalez: Shouldn't we use here:
```
. Install a new proxy{nbsp}{productnumber}, following xref:install-proxy-uyuni.adoc[Uyuni Proxy Installation].
```
Just as you did for Uyuni?

modules/installation/pages/uyuni-proxy-setup.adoc
```diff
@@ -209,32 +209,25 @@ filename: "pxelinux.0"
...
-// ** Copy the scripts that are still needed.
-// ** Copy the activation keys from the previous server.
-// Of course, it is always better to re-create the keys.
+Consider copying important or manually created data to a central place that can also be accessed by the new proxy.
 . Shut down the proxy.
-. Install a new proxy{nbsp}{productnumber}, following xref:install-proxy-unified.adoc[Proxy Installation].
+. Install a new proxy{nbsp}{productnumber}, following xref:install-proxy-uyuni.adoc[Uyuni Proxy Installation].
 . In the {productname} {webui} select the newly installed {susemgrproxy} and delete it from the systems list.
-[[step.at.proxy.migration3.replace.react]]
 . In the {webui}, create a reactivation key for the old proxy system: On the System Details tab of the old proxy click [guimenu]``Reactivation``.
```
juliogonzalez: @mantel do we need this new reactivation key?

mantel: The re-activation key is only needed if you do not want to lose the history of the machine. Else it will become a "new" one with a new ID.

modules/installation/pages/proxy-setup.adoc
keichwa: I push a change for clarity.

modules/installation/pages/proxy-setup.adoc
keichwa: ok, I'll add that clarification.

modules/installation/pages/proxy-setup.adoc
keichwa: I do not think so.  I think the preferred way to install the SUMA Proxy is to use the Unified Installer.

modules/installation/pages/uyuni-proxy-setup.adoc
keichwa: added this info also for uyuni

keichwa:
if you've time please recheck.
@Loquacity I need your review, too :)

modules/installation/pages/proxy-setup.adoc
```diff
@@ -196,56 +197,57 @@ For more information about Cobbler, see xref:client-configuration:cobbler.adoc[C
 [[proxy.pxe.dhcp]]
 === Configure DHCP for PXE through {susemgrproxy}
```
juliogonzalez: For coherence, shouldn't this be `{productname} Proxy` as well?

modules/installation/pages/proxy-setup.adoc
```diff
@@ -196,56 +197,57 @@ For more information about Cobbler, see xref:client-configuration:cobbler.adoc[C
...
 ----
 
 
 
-// REMARK: 2019-08-23, ke: this needs closer checking
-// see https://github.com/SUSE/spacewalk/issues/10521
-// and https://github.com/uyuni-project/uyuni-docs/issues/110
-[[replacing.a.susemgrproxy]]
+[[replace-susemgrproxy]]
 == Replace a {susemgrproxy}
```
juliogonzalez: Same.

modules/installation/pages/uyuni-proxy-setup.adoc
```diff
@@ -209,32 +208,34 @@ filename: "pxelinux.0"
...
 
 
-////
-// REMARK: 2019-08-23, ke: this needs closer checking
-// see https://github.com/uyuni-project/uyuni-docs/issues/110
-[[replacing.a.susemgrproxy]]
-== Replace a {susemgrproxy}
+
+[[replace-uyuniproxy]]
+== Replace a Uyuni Proxy
```
juliogonzalez: `{productname} Proxy` maybe?

keichwa:
On 2020-03-13 17:41, Julio González Gil wrote:
> juliogonzalez approved this pull request.
> 
> Current PR should work, but I strongly suggest deciding what to use in
> all places. Sometimes we have `{productname} Proxy`, sometimes
> `{susemgrproxy}` and sometimes only `Proxy` :-)

Yes, agreed.  I can probably merge all this and then have it in one
file only again.

-- 
Karl Eichwalder

keichwa:
> Current PR should work, but I strongly suggest deciding what to use in all places. Sometimes we have `{productname} Proxy`, sometimes `{susemgrproxy}` and sometimes only `Proxy` :-)

The "problem" is that macro expansion does not always work everywhere...  Nevertheless I'll now try to make better use of it and commit changes.

modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,58 +195,59 @@ For more information about Cobbler, see xref:client-configuration:cobbler.adoc[C
...
+** Install any other needed software
+** Set up TFTP synchronization if the proxy is used for autoinstallation
 
 [IMPORTANT]
-.Proxy Installation and Client Connections
 ====
 During the installation of the proxy, clients will not be able to reach the {productname} Server.
-After a {susemgrproxy} system has been deleted from the systems list, all clients connected to this proxy will be (incorrectly) listed as `directly connected` to the {productname} Server.
-After the first successful operation on a client _such as execution of a remote command or installation of a package or patch_ this information will automatically be corrected.
+After a proxy system has been deleted from the systems list, all clients connected to this proxy will be (incorrectly) listed as `directly connected` to the {productname} Server.
```
Loquacity:
```suggestion
After a proxy system has been deleted, the systems list shows all clients that were connected to the deleted proxy, as now being directly connected to the {productname} Server.
This is incorrect.
```

modules/installation/pages/uyuni-proxy-setup.adoc
```diff
@@ -9,8 +9,7 @@
 == Install the [path]``uyuni_proxy`` pattern
 
 Check that the Proxy pattern is installed correctly.
-You can do this in the {productname} server {webui} by attempting to install the [package]``patterns_uyuni_proxy`` package.
-Alternatively, use this command on the proxy, as root, to check that the [path]``uyuni_proxy`` pattern is installed:
+On the server select the [package]``pattern_uyuni_proxy`` package for installation, or make sure the [path]``uyuni_proxy`` pattern is installed using the following command on the proxy as root:
```
Loquacity:
```suggestion
On the server select the [package]``pattern_uyuni_proxy`` package for installation, or make sure the [path]``uyuni_proxy`` pattern is installed using this command on the proxy as root:
```

modules/installation/pages/uyuni-proxy-setup.adoc
```diff
@@ -209,32 +208,34 @@ filename: "pxelinux.0"
...
+[[replace-uyuniproxy]]
+== Replace a {productname} Proxy
 
 A proxy does not contain any information about the clients that are connected to it.
-Because of this, a proxy can be replaced by a new one at any time.
+Therefore, a proxy can be replaced by a new one at any time.
 The replacement proxy must have the same name and IP address as its predecessor.
 
-Leave the old proxy in place while you prepare the replacement.
+Shut down the old proxy, but leave it installed while you prepare the replacement.
```
Loquacity:
```suggestion
Shut down the old proxy, and leave it installed while you prepare the replacement.
```

modules/installation/pages/uyuni-proxy-setup.adoc
```diff
@@ -209,32 +208,34 @@ filename: "pxelinux.0"
...
 Create a reactivation key for this system and then register the new proxy using the reactivation key.
-If you do not use the reactivation key, you will need to re-register all the clients against the new proxy.
+If you do not use the reactivation key, you will need to re-register all the
+clients against the new proxy.
```
Loquacity:
```suggestion
If you do not use the reactivation key, you will need to re-register all the clients against the new proxy.
```

modules/installation/pages/uyuni-proxy-setup.adoc
```diff
@@ -209,32 +208,34 @@ filename: "pxelinux.0"
...
+
+[NOTE]
+====
+The reactivation key is only needed if you do not want to lose the history of the machine.
+If you do not use a reactivation key, the replacement proxy will become a "new" one with a new ID.
+====
+
+Here is a detailed procedure step-by-step:
```
Loquacity:
```suggestion
```

modules/installation/pages/uyuni-proxy-setup.adoc
```diff
@@ -209,32 +208,34 @@ filename: "pxelinux.0"
...
 . Before starting the actual migration procedure, save the data from the old proxy, if needed.
-Consider copying important data to a central place that can also be accessed by the new proxy.
-// REMARK: 2019-08-23, ke: this is unclear
-// ** Copy the scripts that are still needed.
-// ** Copy the activation keys from the previous server.
-// Of course, it is always better to re-create the keys.
+Consider copying important or manually created data to a central place that can also be accessed by the new proxy.
 . Shut down the proxy.
-. Install a new proxy{nbsp}{productnumber}, following xref:install-proxy-unified.adoc[Proxy Installation].
+. Install a new proxy{nbsp}{productnumber}, following xref:install-proxy-uyuni.adoc[Uyuni Proxy Installation].
```
Loquacity:
```suggestion
. Install a new {productname} Proxy.
For installation instructions, see xref:install-proxy-unified.adoc[Proxy Installation].
```

modules/installation/pages/proxy-setup.adoc
keichwa: Yes, thanks, this is better. But I fear it is still confusing.  I think this is only temporarily and it is only incorrectly displayed or communicated to the user. The configuration itself is correct, and this will be visible to the user after a certain period of time.

Do we need a note or different wording, @Loquacity ?

modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,58 +195,59 @@ For more information about Cobbler, see xref:client-configuration:cobbler.adoc[C
...
+Click [guimenu]``Generate New Key``, and make a note of the new key, as you will need it later.
 For more information about reactivation keys, see xref:reference:systems/system-details/sd-reactivation.adoc[Reactivation Keys].
-. After the installation of the new proxy, perform the following actions (if needed):
-** Copy the centrally saved data to the new proxy system.
-** Install any other needed software.
-** If the proxy is also used for autoinstallation, do not forget to setup TFTP synchronization.
+. After the installation of the new proxy, you might also need to:
+** Copy the centrally saved data to the new proxy system
+** Install any other needed software
+** Set up TFTP synchronization if the proxy is used for autoinstallation
```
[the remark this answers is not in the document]
keichwa: fixed with an external commit.

modules/installation/pages/proxy-setup.adoc
Loquacity: Yeah, maybe. What about this instead? It moves the important information first, adds "temporarily" to make it clear it's not permanent, and splits the sentence up.

```suggestion
After you have a deleted a proxy, the systems list can be temporarily incorrect.
All clients that were previously connected to the proxy will show as being directly connected to the server instead.
```

PR #138, opened 2020-03-12
Fix Makefile

paususe:
There were left-overs from a previous conflict


paususe:
@Loquacity @jcayouette 
Could you please merge this? The docs currently do not build because the Makefile contains left-overs from a previous conflict.

PR #144, opened 2020-03-13
Explain sizing of /var/cache on both the Server and the Proxy

paususe:
Related to https://github.com/SUSE/spacewalk/issues/10863 , in this PR it is explained the rationale behind this mount point size.

### DIFF
--- a/modules/installation/pages/hardware-requirements.adoc
+++ b/modules/installation/pages/hardware-requirements.adoc
@@ -21,7 +21,8 @@ For {smr} hardware requirements, see xref:retail:retail-requirements.adoc[].
 |                          | _Production Server_ Minimum 32{nbsp}GB
 | Disk Space:              | [path]``/`` _(root)_ Minimum 24{nbsp}GB
 |                          | [path]``/var/lib/pgsql`` Minimum 50{nbsp}GB
-|                          | [path]``/var/spacewalk`` Minimum 50{nbsp}GB per SUSE product and 360{nbsp}GB per Red Hat product
+|                          | [path]``/var/spacewalk`` Minimum 50{nbsp}GB per {suse} product and 360{nbsp}GB per Red Hat product
+|                          | [path]``/var/cache`` In accordance to products to be synchronized: 100{nbsp}MB per {suse} product, 1{nbsp}GB per Red Hat product. Double that when the Server is an ISS master.
 | Swap space:               | 3{nbsp}GB
 |===
 
@@ -36,6 +37,7 @@ For {smr} hardware requirements, see xref:retail:retail-requirements.adoc[].
 | Disk Space:              | [path]``/`` Minimum 100{nbsp}GB
 |                          | [path]``/var/lib/pgsql`` Minimum 50{nbsp}GB
 |                          | [path]``/var/spacewalk`` Minimum 50{nbsp}GB per SUSE product and 360{nbsp}GB per Red Hat product
+|                          | [path]``/var/cache`` In accordance to products to be synchronized: 100{nbsp}MB per {suse} product, 1{nbsp}GB per Red Hat product. Double that when the Server is an ISS master.
 | Swap space:               | 3{nbsp}GB
 |===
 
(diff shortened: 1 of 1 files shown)


paususe:
@Loquacity @keichwa could you please review?

modules/installation/pages/hardware-requirements.adoc
```diff
@@ -52,5 +52,15 @@ For {smr} hardware requirements, see xref:retail:retail-requirements.adoc[].
...
+
+{productname} Server stores packages in [path]``/var/spacewalk``. In case it runs out of disk space, the repository synchronization will fail. Based on what and how many client operating systems and repositories you plan to mirror, you shold estimate how much space to assign to [path]``/var/spacewalk`` and provide enough space.
+
+{productname} Proxy stores packages in [path]``/var/cache`` but only as a cache. When not enough space is available under [path]``/var/cache``, {productname} Proxy will remove old, unused packages and replace them with newer packages.
+
+Because of that:
+
+* The larger [path]``/var/cache`` is on the {productname} Proxy or {productname} Retail Branch Server, the less traffic there will be between it and the {productname} Server. 
+* By making [path]``/var/cache`` on the {productname} Proxy at least the same size as [path]``/var/spacewalk`` in the {productname} Server, you could avoid a lot of traffic after the first sync.
+* [path]``/var/cache`` can be rather small on the {productname} Server compared to {productname} Proxy (less than 15G should be enough).
```
Loquacity:
```suggestion
{productname} Server stores packages in the [path]``/var/spacewalk/`` directory. 
Repository synchronization fails if this directory runs out of disk space. 
You can estimate how much space the [path]``/var/spacewalk/`` directory requires based on the number and type of clients and repositories you plan to mirror.

{productname} Proxy caches packages in the [path]``/var/cache/`` directory.
If there is not enough space available in [path]``/var/cache/``, the proxy will remove old, unused packages and replace them with newer packages.

As a result of this behavior:

* The larger [path]``/var/cache/`` directory is on the {productname} Proxy, the less traffic there will be between it and the {productname} Server. 
* By making the [path]``/var/cache/`` directory on the {productname} Proxy at least the same size as [path]``/var/spacewalk/`` in the {productname} Server, you avoid a large amount of traffic after the first synchronization.
* The [path]``/var/cache/`` directory can be small on the {productname} Server compared to {productname} Proxy.
We recommend less than 15{nbsp}GB in this situation.
```

keichwa: Instead of the last sentence (`We recommend less than 15{nbsp}GB in this situation.`), I'd write for clarity:
```
{suse} recommends less than 15{nbsp}GB on the{productname} Server in this situation.
```
And let's avoid "we" phrases.

paususe: I agree to most of @Loquacity 's changes. I'm not sure less than 15 GB are actually "recommended", it's actually "allowed", "enough", etc.
@mcalmer what do you think?

mcalmer: It depends on how many repos you sync. On a ISS master additional space is required as all data gets pre-generated also in /var/cache. 
I think "less than 15 GB" is wrong in some cases. Especially when the customer sync a lot of channels.

paususe: How much additional space would be required on an ISS master? 

Can we give some kind of a metric, e. g. "each channel requires X MB of disk on /var/cache on a normal Server and Y MB additionally if ISS master"?

mcalmer: ~ 80M -100M per channel. RES channel 1G.
Double the number when Server is an ISS master.
(just guessing by looking at some numbers on our server)

paususe: I have updated the text with Lana's and MC's suggestions.

modules/installation/pages/hardware-requirements.adoc
```diff
@@ -52,5 +54,19 @@ For {smr} hardware requirements, see xref:retail:retail-requirements.adoc[].
...
+Repository synchronization fails if this directory runs out of disk space. 
+You can estimate how much space the [path]``/var/spacewalk/`` directory requires based on the number and type of clients and repositories you plan to mirror.
+
+{productname} Proxy caches packages in the [path]``/var/cache/`` directory.
+If there is not enough space available in [path]``/var/cache/``, the proxy will remove old, unused packages and replace them with newer packages.
+
+As a result of this behavior:
+
+* The larger [path]``/var/cache/`` directory is on the {productname} Proxy, the less traffic there will be between it and the {productname} Server. 
+* By making the [path]``/var/cache/`` directory on the {productname} Proxy at least the same size as [path]``/var/spacewalk/`` in the {productname} Server, you avoid a large amount of traffic after the first synchronization.
```
mcalmer: "at least the same size" - Makeing it bigger is totally useless. So I would suggest we remove "at least".
Even same size is not recommended/needed. You typically never use all packages behind the proxy. So you would just spend never used disk space. But this is a decission of the admin.

PR #119, opened 2020-03-05
Adds Debian to nav (for uyuni only), edits.

Loquacity:
Following up on https://github.com/uyuni-project/uyuni-docs/pull/111

Note that I've added it to the nav, and put the Uyuni conditional at that level, rather than in individual sections in the chapter itself. This was for a few reasons:

* The SUMA content seemed incomplete
* It seems to me that to manage Debian clients on SUMA, you need to do at least a few things that are technically unsupported. We should probably avoid this kind of content where possible.
* It just makes everything simpler to have the conditional at a higher level (and we also intend to split the Ubuntu content in a similar way at a later date, too).

You will also note that I've commented out a couple of larger chunks of content late in the chapter, mostly because it's repeating stuff we have elsewhere in the docs suite  (CLM and bootstrapping). I'm happy to include it at a high level with an xref or something, instead.

PR #111, opened 2020-02-28
Debian9 10 onboarding

gnuninu:
Created a new onboarding page for debian clients.

- uploaded pictures in assets/images and referenced them in the page
- Removed sections related with manual repo and channels creation of the debian tools - no longer required
- Pictures modified accordingly

PR #145, opened 2020-03-14
Changed -master to -devel to reflect channels.ini

gnuninu:
- Changed -master to -devel to reflect channels.ini when running spacewalk-common-channels

### DIFF
--- a/modules/client-configuration/pages/clients-debian.adoc
+++ b/modules/client-configuration/pages/clients-debian.adoc
@@ -54,7 +54,7 @@ For {debian}{nbsp}9:
 spacewalk-common-channels -u admin -a amd64-deb debian-9-pool-amd64-uyuni
 spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-main-security-uyuni
 spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-main-updates-uyuni
-spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-uyuni-client-client
+spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-uyuni-client
 ----
 For {debian}{nbsp}10:
 +


keichwa:
Thanks why we have you ;)

In product documentation, I would not talk about "devel" at all ;^)  But ok, we live in these agile times...

modules/client-configuration/pages/clients-debian.adoc
```diff
@@ -54,15 +54,15 @@ For {debian}{nbsp}9:
 spacewalk-common-channels -u admin -a amd64-deb debian-9-pool-amd64-uyuni
 spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-main-security-uyuni
 spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-main-updates-uyuni
-spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-uyuni-client-master
+spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-uyuni-client-client
```
Loquacity:
```suggestion
spacewalk-common-channels -u admin -a amd64-deb debian-9-amd64-uyuni-client
```

modules/client-configuration/pages/clients-debian.adoc
```diff
@@ -54,15 +54,15 @@ For {debian}{nbsp}9:
...
 ----
 For {debian}{nbsp}10:
 +
 ----
 spacewalk-common-channels -u admin -a amd64-deb debian-10-pool-amd64-uyuni
 spacewalk-common-channels -u admin -a amd64-deb debian-10-amd64-main-security-uyuni
 spacewalk-common-channels -u admin -a amd64-deb debian-10-amd64-main-updates-uyuni
-spacewalk-common-channels -u admin -a amd64-deb debian-10-amd64-uyuni-client-master
+spacewalk-common-channels -u admin -a amd64-deb debian-10-amd64-uyuni-client-client
```
Loquacity:
```suggestion
spacewalk-common-channels -u admin -a amd64-deb debian-10-amd64-uyuni-client
```

gnuninu:
thanks guys for reviewing

Proposal based on a personal taste:
Following up on PR #119 and the original #111
Wouldn't be better to have all the steps required to register the client available on the same page?
e.g. under
Uyuni --> Client Configuration Guide --> Other Clients --> Debian Clients

1) Create custom channels
2) Sync channels
3) Clone channels or create lifecycle content
4) Create activation key
5) Bootstrapping from UI and script

I know that this may add redundancy to the overall doc but it may help the users (especially beginners) having everything they need on the same page rather than forcing them to browse over the pages...

My 2 cents

Loquacity:
> thanks guys for reviewing
> 
> Proposal based on a personal taste:
> Following up on PR #119 and the original #111
> Wouldn't be better to have all the steps required to register the client available on the same page?
> e.g. under
> Uyuni --> Client Configuration Guide --> Other Clients --> Debian Clients
> 
> 1. Create custom channels
> 2. Sync channels
> 3. Clone channels or create lifecycle content
> 4. Create activation key
> 5. Bootstrapping from UI and script
> 
> I know that this may add redundancy to the overall doc but it may help the users (especially beginners) having everything they need on the same page rather than forcing them to browse over the pages...
> 
> My 2 cents

Thanks for the feedback. To try and give you an idea of why we're doing it this way, I need to explain an information architecture/data typing method called "Every Page is Page One" (it's been developed by Mark Baker, you can google for it). We don't yet have our docs in quite this form, but it's the long-term goal. EPPO documentation tends to read a little bit like a Wikipedia page: it gives all the information required for a particular topic (in our case, it's usually a concept and a series of related tasks), and extra information is provided through links to other content. This breaks documentation out of a model where you need to read everything in order, or where later chapters are reliant on information in earlier chapters. That's good for us, because we can drop EPPO topics in to a book in almost any location, provided it has working links to relevant information, and we provide a good way for readers to navigate around (browsing through the topics in a list, search, navigation elements, etc).

So, in our documentation, we treat "Registering Debian Clients" as one EPPO topic, "Creating Custom Channels" as another topic, and "Bootstrapping" as another, then we link back and forth between them as required. This means we only need to write and maintain one version of each, so we end up with better quality documentation that can be maintained by fewer writers. It also means we're not giving that information over and over again (possibly in slightly different forms) to readers who might not need it. So if you're a reader who has registered other clients, and now you're doing a Debian client, you probably don't need to understand how to create an Activation Key, but just in case you do, you can easily skip over to that information. It's the technical documentation equivalent of falling into a Wikipedia rabbit hole :wink: 

If you want some more info about EPPO, you can watch this short video I recorded at a conference a few years ago: https://www.youtube.com/watch?v=VKaRJ-_vHig

Hope that helps!

gnuninu:
> Every Page is Page One

Thanks @Loquacity for the explanation, very interesting indeed.
I will take a look at this concept.
I kind of noticed that there was some sort of intelligence built-in in the doc code that i am still not entirely able to grasp :) 

Let's take the ubuntu page as an example.
https://www.uyuni-project.org/uyuni-docs/uyuni/client-configuration/clients-ubuntu.html

After "Root Access" I would personally point the users to an "how to create an activation key" page
https://www.uyuni-project.org/uyuni-docs/uyuni/client-configuration/clients-and-activation-keys.html

Is this something correct based on the design you have in mind?
If yes what's the best way to reference it, with a simple link perhaps?

Loquacity:
If you want to put it there, you would need to mention something about the Activation Key first, possibly in a new section. I would be more inclined to add this line up in the GPG keys section, after we mention activation keys:
`For more information on Activation Keys, see xref:client-configuration:clients-and-activation-keys.adoc[].`
And possibly a link to the config channels section, too, once I write it ;)

PR #151, opened 2020-03-17
Add webui instructions to SLES-ES

Loquacity:
https://github.com/SUSE/spacewalk/issues/10805

PR #152, opened 2020-03-17
Unify placeholders in the examples

calancha:
* modules/client-configuration/pages/clients-sleses.adoc (Add Base Media)
Use <os_name> placeholder in the examples.
Use <SERVER-FQDN> instead of EXAMPLE.COM in the examples.

### DIFF
--- a/modules/client-configuration/pages/clients-sleses.adoc
+++ b/modules/client-configuration/pages/clients-sleses.adoc
@@ -170,15 +170,15 @@ The parent channel will not contain any packages.
 
 . On the {productname} Server, at the command prompt, as root, copy the base media image to the [path]``/tmp/`` directory.
 . Create a directory to contain the media content.
-Replace [command]``<os_name>`` with either ``rhel``  or ``centos``:
+Replace [command]``<os_name>`` with either ``sleses6``,  ``sleses7`` or ``sleses8``:
 +
 ----
-mkdir -p /srv/www/htdocs/pub/rhel
+mkdir -p /srv/www/htdocs/pub/<os_name>
 ----
 . Mount the image:
 +
 ----
-mount -o loop /tmp/<iso_filename> /srv/www/htdocs/pub/rhel
+mount -o loop /tmp/<iso_filename> /srv/www/htdocs/pub/<os_name>
 ----
 
 When the image is mounted, you can synchronize the base media.


keichwa:
I'd keep example.com.  Maybe, we should lowercase it again.  Whatever we decide to use, we should document it.  If we don't do this, we will change it again and again.

Loquacity:
> I'd keep example.com. Maybe, we should lowercase it again. Whatever we decide to use, we should document it. If we don't do this, we will change it again and again.

We previously agreed on lowercase and angle brackets. I like to revert to example.com where it makes sense, as it's a IANA reserved address with guidance about what to do if you hit it (it's also considered "best practice" in technical writing guidance).

jcayouette:
Agree on example.com

modules/client-configuration/pages/clients-sleses.adoc
```diff
@@ -175,13 +175,13 @@ Mount the {rhel} or {centos} 6, 7 or 8 Media:
 Replace [command]``<os_name>`` with either ``rhel``  or ``centos``:
```
paususe: ... or ``sleses`` (Expanded Support), or ``ol`` (Oracle Linux), and version should also be specified. E. g. ``sleses8``.

Given that this page is for SLES ES, I wouldn't menion RHEL or CentOS but say something like:

```suggestion
Replace [command]``<os_name>`` with either ``sleses6``, ``sleses7``  or ``sleses8``.
```

We should have this same content in the RHEL and CentOS doc pages.

modules/client-configuration/pages/clients-sleses.adoc
```diff
@@ -197,15 +197,15 @@ The following steps require you to have mounted the base media for the selected
 . At the command prompt as root enter the following to sync the **BaseOS** channel:
 +
 ----
-spacewalk-repo-sync -c channel_name -u https://EXAMPLE.COM/pub/rhel/BaseOS/
+spacewalk-repo-sync -c channel_name -u https://<SERVER-FQDN>/pub/<os_name>/BaseOS/
```
paususe:
```suggestion
spacewalk-repo-sync -c channel_name -u https://susemanager.example.com/pub/<os_name>/BaseOS/
```

modules/client-configuration/pages/clients-sleses.adoc
```diff
@@ -197,15 +197,15 @@ The following steps require you to have mounted the base media for the selected
...
 ----
 
 . Wait for the sync to complete.
 
 . At the command prompt as root enter the following to sync the **AppStream** channel:
 +
 ----
-spacewalk-repo-sync -c channel_name -u https://EXAMPLE.COM/pub/rhel/AppStream/
+spacewalk-repo-sync -c channel_name -u https://<SERVER-FQDN>/pub/<os_name>/AppStream/
```
paususe:
```suggestion
spacewalk-repo-sync -c channel_name -u https://susemanager.example.com/pub/<os_name>/AppStream/
```

modules/client-configuration/pages/clients-sleses.adoc
```diff
@@ -215,7 +215,7 @@ spacewalk-repo-sync -c channel_name -u https://EXAMPLE.COM/pub/rhel/AppStream/
 At the command prompt as root enter the following:
 +
 ----
-spacewalk-repo-sync -c channel_name -u https://EXAMPLE.COM/pub/<os_name>/
+spacewalk-repo-sync -c channel_name -u https://<SERVER-FQDN>/pub/<os_name>/
```
paususe:
```suggestion
spacewalk-repo-sync -c channel_name -u https://susemanager.example.com/pub/<os_name>/
```

Loquacity:
I agree with Pau's suggestions.

modules/client-configuration/pages/clients-sleses.adoc
calancha: Thank you for the suggestions.
It turned out that the command line part discussed here was dropped in PR #151 (*).  Therefore, I am going to address just the existent part in the code base, i.e.,
- Use <os_name> in the examples
- Refer to sleses instead of centOS/RHEL.

(*) In case we wish to reinclude the command line instructions, then I suggest we open a new issue to address that.

PR #155, opened 2020-03-18
DHCP Formula for UEFI

Loquacity:
https://github.com/SUSE/spacewalk/issues/10956

### DIFF
--- a/modules/salt/pages/formula-dhcpd.adoc
+++ b/modules/salt/pages/formula-dhcpd.adoc
@@ -24,7 +24,7 @@ The DHCPd formula is used to configure the DHCP service on the branch server.
 . In the [guimenu]``Next Server`` field, enter the hostname or IP address of the branch server (for example: [systemitem]``192.168.1.5``).
 . In the [guimenu]``Filename`` field, if you are booting a client using PXE, type the path to the PXE bootloader.
 Otherwise, keep the default value of [systemitem]``/boot/pxelinux.0``.
-. In the [guimenu]``Filename Efi`` field, if you are booting a UEFI client using PXE, type the path the PXE bootloader.
+. In the [guimenu]``Filename Efi`` field, if you are booting a UEFI client using PXE, type the path to the PXE bootloader.
 Otherwise, keep the default value of [systemitem]``/boot/shim.efi``.
 . In the [guimenu]``Filename Http`` field, if you are booting a UEFI client using HTTP, type [systemitem]``http://branchserver/saltboot/boot/shim.efi``.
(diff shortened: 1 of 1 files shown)


modules/salt/pages/formula-dhcpd.adoc
```diff
@@ -22,6 +22,10 @@ The DHCPd formula is used to configure the DHCP service on the branch server.
 . In the [guimenu]``Broadcast Address`` field, enter the broadcast IP address for the branch network (for example: [systemitem]``192.168.1.255``).
 . In the [guimenu]``Routers`` field, enter the IP address to be used by routers in the branch server network (for example: [systemitem]``192.168.1.5``).
 . In the [guimenu]``Next Server`` field, enter the hostname or IP address of the branch server (for example: [systemitem]``192.168.1.5``).
-. In the [guimenu]``Filename`` field, keep the default value of [systemitem]``/boot/pxelinux.0``.
+. In the [guimenu]``Filename`` field, if you are booting a client using PXE, type the path to the PXE bootloader.
+Otherwise, keep the default value of [systemitem]``/boot/pxelinux.0``.
+. In the [guimenu]``Filename Efi`` field, if you are booting a UEFI client using PXE, type the path the PXE bootloader.
```
keichwa:
```suggestion
. In the [guimenu]``Filename Efi`` field, if you are booting a UEFI client using PXE, type the path to the PXE bootloader.
```

jcayouette:
I am seeing two failures in our xrefs:

``` 
path: modules/client-configuration/pages/registration-bootstrap.adoc | xref: administration:custom-ssl.adoc
path: modules/reference/pages/admin/bootstrap-script.adoc | xref: administration:custom-ssl.adoc
```

modules/salt/pages/formula-dhcpd.adoc
```diff
@@ -22,6 +22,10 @@ The DHCPd formula is used to configure the DHCP service on the branch server.
...
+Otherwise, keep the default value of [systemitem]``/boot/shim.efi``.
```
nadvornik: The default values will work for 90% of installations, so it is probably better to say
"If unsure, keep the default value ..."

Loquacity: Heh, that feels like we're talking down to readers a bit (as in, "you probably won't understand this, it's very technical"). Does the "otherwise" not cut it?

nadvornik: Maybe you can use different words, but it should be clear that the default values should not be changed in typical installation.

PR #170, opened 2020-03-25
Deploy terminals by USB

Loquacity:
https://github.com/SUSE/spacewalk/issues/11045

### DIFF
--- a/modules/retail/pages/retail-deploy-terminals.adoc
+++ b/modules/retail/pages/retail-deploy-terminals.adoc
@@ -11,7 +11,7 @@ When you have the {susemgr} Server and Branch Server set up, you are ready to de
 
 Each procedure is detailed in this section.
 
-If you do not want to boot terminals from the network, you can create a live USB image and deploy terminals using a removable USB storage device.
+If you are not able to boot terminals from the network, you can create a live USB image and deploy terminals using a removable USB storage device.
 
 If you have many terminals, and would prefer this to be handled with a script, see xref:retail:retail-mass-config.adoc[].
 


modules/retail/pages/retail-deploy-terminals.adoc
```diff
@@ -46,14 +51,16 @@ For example:
 ----
 HWTYPE:POSVendor-Terminal1
 ----
-+
+. In the {susemgr} {webui}, navigate to menu:Systems[System Groups], and click the btn:[Create Group] button.
+. In the [guimenu]``Create System Group`` dialog, create a new system group, using the hardware type group name you created earlier.
```
keichwa:
```suggestion
. In the [guimenu]``Create System Group`` dialog, create a new system group, using the hardware type group name you determined in step one of this procedure.
```

modules/retail/pages/retail-deploy-terminals.adoc
```diff
@@ -46,14 +51,16 @@ For example:
...
+
+
```
keichwa:
```suggestion
```

modules/retail/pages/retail-deploy-terminals.adoc
```diff
@@ -120,7 +128,64 @@ image::retail_imagesync_formula.png[scaledwidth=80%]
...
 
 . Power on your POS terminals.
 . The branch server will bootstrap the terminals according to the data you have provided.
+
+
+
+== Deploy Terminals with a Removable USB Device
+
+If you do not want to boot terminals from the network, you can create a live USB image and deploy terminals using a removable USB storage device.
+This is useful if you are unable or unwilling to connect some or all of your terminals to the network, or if you do not have a local {smr} branch server providing network services.
```
keichwa:
```suggestion
This is useful if you cannot connect your terminals to the network, or if you do not have a local {smr} branch server providing network services.
```
The user will know why USB might be a good option.

modules/retail/pages/retail-deploy-terminals.adoc
```diff
@@ -120,7 +128,64 @@ image::retail_imagesync_formula.png[scaledwidth=80%]
...
+----
+salt <branch_server> image_sync_usb.create <usb image name> <size in MiB>
+----
+. Insert the USB device into the {smr} Server machine, and copy the image to the new location:
++
+----
+salt <branch_server> cmd.run "dd if=<usb image name> of=<path to usb device on branch>"
+----
+
+
```
keichwa:
```suggestion
```

modules/retail/pages/retail-deploy-terminals.adoc
aaannz: This is not entirely connect. Terminals still need access to the network because the USB stick contains only boot image part usually served by PXE. The system image with OS and custom retail applications still needs to be fetched from SUMA server.

I think that use case for this feature are ATMs. They have network connection, but they do not have local branch which would provide PXE and tftp services. So they are booted using this USB stick (which may contain VPN configuration) and connect to some remote branch in the datacenter already through salt.

modules/retail/pages/retail-deploy-terminals.adoc
```diff
@@ -120,7 +128,64 @@ image::retail_imagesync_formula.png[scaledwidth=80%]
...
+You can check this by editing the Saltboot formula in the {smr} {webui}.
+For more information about the Saltboot formula, see xref:salt:formula-saltboot.adoc[].
+
+
+
+.Procedure: Deploying Images to the Terminals using USB
+
+. Power on the POS terminal.
+. Insert the USB device into the terminal.
+. If not automatically done, mount the USB drive to begin bootstrapping.
```
aaannz: This USB is bootable, so the order should be:
```adoc
. Insert the USB device into the terminal.
. Power on the POS terminal.
. Make sure booting from USB device is enabled and boot from USB device.
```

modules/retail/pages/retail-deploy-terminals.adoc
```diff
@@ -120,7 +128,64 @@ image::retail_imagesync_formula.png[scaledwidth=80%]
...
+. On the {smr} Server, at the command prompt, as root, create the POS image:
++
+----
+salt <branch_server> image_sync_usb.create <usb image name> <size in MiB>
+----
+. Insert the USB device into the {smr} Server machine, and copy the image to the new location:
++
+----
+salt <branch_server> cmd.run "dd if=<usb image name> of=<path to usb device on branch>"
+----
```
aaannz: This is not correct. The snipped was meant to show that user can trigger usb stick build from server, however the resulting image is still on branch server.
So unless user transfers the image from branch server to suma server, usb device needs to be available on branch server.
I think this section brings only more confusion and probably best is to remove this procedure and leave only the first one where the image is created on branch server.

modules/retail/pages/retail-deploy-terminals.adoc
```diff
@@ -120,7 +128,64 @@ image::retail_imagesync_formula.png[scaledwidth=80%]
...
+
+You can create the bootable USB device on the branch server directly, or on the {smr} Server.
+
+
+
+.Procedure: Creating a Bootable USB Device on the Branch Server
+. On the {smr} branch server, at the command prompt, as root, create the POS image:
++
+----
+salt-call image_sync_usb.create <usb image name> <size in MiB>
```
aaannz: I forgot to add that the size should be at least 300MB.

Loquacity:
> too many single sentence paras.

Oh, but I'm so good at them! :stuck_out_tongue_winking_eye:

PR #175, opened 2020-03-27
monitoring ports

Loquacity:
https://github.com/SUSE/spacewalk/issues/9413#issuecomment-604418355

@witekest please review :)


modules/installation/pages/ports.adoc
```diff
@@ -1,5 +1,5 @@
 [[ports]]
-= Required Network Ports
+= Network Ports
```
keichwa: I think we must add port `9090`.

witekest: I'd keep it as it is or mark as _optional_. We normally don't deploy Prometheus server on the client. Federation is a special case which possibly will be covered when working on cluster awareness story.

modules/installation/pages/ports.adoc
```diff
@@ -146,4 +146,7 @@ In most cases, you will not need to adjust these ports.
 | 22          | Inbound   | SSH      | Required for ssh-push and ssh-push-tunnel contact methods.
 | 80          | Outbound  |          | Used to reach the server or proxy.
 | 5222        | Outbound  | TCP      | Required to push OSAD actions to the server or proxy.
+| 9100        | Outbound  | TCP      | Required for Prometheus node exporter
+| 9117        | Outbound  | TCP      | Required for Prometheus Apache exporter
+| 9187        | Outbound  | TCP      | Required for Prometheus PostgreSQL
```
witekest: Nit: Points missing at the ends of the sentences.

cavalheiro:
Following up on a suggestion from Don, it would be important to add that Alertmanager runs on port 9093, and that if you run it on a separate host (different from the one where you run Prometheus) that this port needs to be open as well.

modules/installation/pages/ports.adoc
Loquacity: > I think we must add port `9090`.

What's on port 9090?

Loquacity: > I'd keep it as it is or mark as _optional_. We normally don't deploy Prometheus server on the client. Federation is a special case which possibly will be covered when working on cluster awareness story.

This is the heading for the entire section. I realised when discussing it in the PHP that this whole page lists ports that are, for the most part, optional. We even have a note stating that at the top: 

> You will not need to open all of these ports.
> Some ports only need to be opened if you are using the service that requires them.

So I removed "Required" because, basically, they aren't.

modules/installation/pages/ports.adoc
Loquacity: To be fair, we should probably remove the periods from all of them, because they're not full sentences, but then I think @keichwa would probably come after me :wink:

keichwa:
On 2020-03-30 03:01, Lana Brindley wrote:

> To be fair, we should probably remove the periods from all of them,
> because they're not full sentences, but then I think @keichwa would
> probably come after me :wink:

Could happen.  But if you'll remove all of them in that column it would
probably be fine.

-- 
Karl Eichwalder

keichwa:
On 2020-03-30 02:58, Lana Brindley wrote:

> -= Required Network Ports
> += Network Ports
> 
>> I think we must add port `9090`.
> 
> What's on port 9090?

I now wee that this is just a proposal where Prometheus could list.
I do not whether we also have to mention the configurable Proetheus
port.  Maybe, we already do.

-- 
Karl Eichwalder

modules/installation/pages/ports.adoc
cavalheiro: Port 9090 is where the Prometheus UI runs - it is exposed by the Prometheus server itself.

PR #176, opened 2020-03-27
Menu changes

Loquacity:
https://github.com/uyuni-project/uyuni/issues/271

This is pretty much a quick-and-dirty attempt at cleaning this up, so a couple of things to note:

* The Reference Guide work is ongoing, and will clean many of these pages up before 4.1
* I have only looked at the Reference Guide. I have no doubt there are other changes required throughout the suite. Given the difficulty of grepping effectively for these, I am going to leave them and we will find them organically as we are working on other things. But if you notice any particularly egregious errors in the meantime, let us know.

### DIFF
--- a/modules/reference/pages/software/software-channel-list-menu.adoc
+++ b/modules/reference/pages/software/software-channel-list-menu.adoc
@@ -6,7 +6,7 @@
 [[ref.webui.software.channel.list.overview]]
 == Channels Overview
 
-The menu:Main Menu[Software > Channels] menu item is the first to appear.
+The menu:Software[Channels] menu item is the first to appear.
 A software channel provides packages grouped by products or applications to simplify the selection of packages to be installed on a system.
 
 There are two types of software channels: base channels and child channels.


modules/reference/pages/software/software-channel-list-menu.adoc
```diff
@@ -1,12 +1,12 @@
...
-= Channel List Menu
+= Channels Menu
 
 
 
 [[ref.webui.software.channel.list.overview]]
 == Channels Overview
 
-The menu:Main Menu[Software > Channels List] menu item is the first to appear.
+The menu:Main Menu[Software > Channels] menu item is the first to appear.
```
keichwa: we no longer want `main menu` as the starting point, iirc:
```suggestion
The menu:Software[Channels] menu item is the first to appear.
```

Loquacity:
> Mmmmm, I am not sure about these changes. Looking at https://uyuni-master-srv.mgr.suse.de/rhn/YourRhn.do it seems most of these changes (but not all) were already in and they are reverted now from what they should be. Maybe you took a look at some outdated screenshots? I'd say keep the ref server as the unique source of truth. In there we have the latest and uptodate code/labels version.
> 
> Note: the pattern change we made and that required the docs to be updated was, for instance:
> 
> * Patches > Patches
>   and we moved to:
> * Patches > Patch List
> 
> I am sorry you spent some time to prepare this PR.
> 
> Bottom line: if there is not hurry for this, I'd say we can go through a complete check and cleanup of the menu labels all at once. Maybe it is a contribution a developer can do as well. Just my 2c

OK, let's just delete this. For the most part, these kinds of changes will be updated organically anyway.

PR #180, opened 2020-03-31
Formula with forms

Loquacity:
https://github.com/SUSE/spacewalk/issues/11060

### DIFF
--- a/modules/salt/pages/formulas-custom.adoc
+++ b/modules/salt/pages/formulas-custom.adoc
@@ -4,9 +4,10 @@
 
 Some formulas are provided by default with {productname}.
 Other official formulas can be installed as RPM packages.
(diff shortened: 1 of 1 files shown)


Loquacity:
> I think the changes go in the right direction but I still do not like the end result.
> 
> The main header is "Custom Salt formulas", yet the next section is "Install Official Salt Formulas". That does not make any sense. I think we should split this page in two clearly separated pages:
> 
> 1. How to install official Salt formulas (I don't like that wording, btw, I'd rather use something like "Install SUSE Manager formulas")
> 2. How to write your own formula (or even install a third-party formula) and create your own form.

Yeah, I was thinking much the same thing, it's a bit confused. I'll merge this one for now, and track a re-write in a new issue, though, because it's probably going to need more important from @aaannz.

PR #181, opened 2020-03-31
add highstate definition to salt terminology

Loquacity:
https://github.com/SUSE/spacewalk/issues/10576

### DIFF
--- a/modules/salt/pages/salt-terminology.adoc
+++ b/modules/salt/pages/salt-terminology.adoc
@@ -51,7 +51,7 @@ For more information on grains, see https://docs.saltstack.com/en/latest/topics/
 
 Highstate::
 This term is used when you apply all outstanding states to all targeted clients at the same time.
-The highstate must be applied regularly when doing broad changes to systems, including enabling and disabling formulas.
+The highstate must be applied when doing changes to systems, including enabling and disabling formulas.
 
 
 Pillar::


modules/salt/pages/salt-terminology.adoc
```diff
@@ -29,46 +29,53 @@ Multiple Salt environments are not supported.
...
 salt '*' grains.items
 ----
-
++
 For more information on grains, see https://docs.saltstack.com/en/latest/topics/grains/.
 
 
+Highstate::
+This term is used when you apply all outstanding states to all targeted clients at the same time.
+The highstate must be applied regularly when doing broad changes to systems, including enabling and disabling formulas.
```
keichwa:
```suggestion
The highstate must be applied when doing changes to systems, including enabling and disabling formulas.
```
I'd say it's not important whether a change is broad or small, and why appyling highstate if no state is outstanding?  But you can say something such as:
If you apply the highstate regularly, you are on the safe side: forgotten outstanding events will then be applied.

Loquacity:
> I'm not sure whether we actually need all the new newlines ;)

So I added all the ``+`` to make the deflist work properly. Previously, only the first para was indented under the definition:

image

keichwa:
On 2020-04-01 02:45, Lana Brindley wrote:
>> I'm not sure whether we actually need all the new newlines ;)
> 
> So I added all the ``+`` to make the deflist work properly.
> Previously, only the first para was indented under the definition:

Yeah.  I had such a feeling ;)

-- 
Karl Eichwalder

PR #182, opened 2020-03-31
fix success msg

keichwa:
close https://github.com/SUSE/spacewalk/issues/11122


keichwa:
Ornela, I unfortunately cannot add you as a reviewer.  Trying this way, @0rnela

PR #189, opened 2020-04-07
LSD Guide - Editing

Loquacity:
https://github.com/SUSE/spacewalk/issues/11144

### DIFF
--- a/modules/large-deployments/pages/operation-reqs.adoc
+++ b/modules/large-deployments/pages/operation-reqs.adoc
@@ -12,7 +12,7 @@ Monitor the server as you scale to identify problems early.
 
 // Might be worth alphabetising these. LKB 2018-08-09
 
-=== Salt Client Onboarding Rate
+== Salt Client Onboarding Rate
 
 The rate at which {productname} can onboard clients is limited and depends on hardware resources.
 Onboarding clients at a faster rate than {productname} is configured for will build up a backlog of unprocessed keys.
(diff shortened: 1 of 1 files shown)


Loquacity:
Thanks @admd I've just done another editing pass. I'd like to merge this in the next 24 hours if possible, so we can get it into the beta 3 package.

modules/large-deployments/pages/hub-namespaces.adoc
```diff
@@ -1,23 +1,36 @@
...
-* the `hubKey` can be obtained via 3 different login methods.
-For more detail, see authentication modes.
-* individual Server IDs can be obtained via `client.hub.listServerIds(hubKey)` (see examples section)
-* the `unicast` namespace assumes all methods receive hubKey and serverId as their first two parameters, then any other parameter as specified by the regular Server API.
-* the `multicast` namespace assumes all methods receive hubKey, a list of Server IDs, then lists of per-Server parameters as specified by the regular Server API.
-Return value will be an array, indexed per Server, of the results of individual Server calls.
\ No newline at end of file
+Hub supports three different namespaces:
+
+* The ``hub`` namespace is used to target the Hub Server.
```
chiaradiamarcelo, line 12: this is not entirely true. The "Hub" namespaces is used to target the Hub XMLRPC API server.
(there are not many methods actually under this namespace, maybe 3 or 4) 

For targetting the Hub directly you don't use any special namespace, you just use it as if it were a regular susemanager API 

EDIT: we can just leave it like this for Beta3 and fix it after if we are gonna release it

PR #190, opened 2020-04-07
update information about CentOS 6 SSL error

rjmateus:
SSL errors described in the troubleshooting occurs in sles11 and also on centos6. Documentation updated accordingly.


modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -90,20 +90,20 @@ Therefore you must not mount [filename]``/tmp`` with the [option]``noexec`` opti
 
 == SSL errors
```
keichwa, line 91:
```suggestion
== SSL Errors
```

PR #197, opened 2020-04-09
VMWare cert tshoot

Loquacity:
https://github.com/SUSE/spacewalk/issues/10924


Loquacity:
@mcalmer waiting on your review so we can merge this.

Loquacity:
> Not sure if the certificate in anchors/ needs a special name. Maybe we can write to use suffix '.crt' or '.pem'.
> But the rest looks good.

Who could we ask about that?

mcalmer:
> > Not sure if the certificate in anchors/ needs a special name. Maybe we can write to use suffix '.crt' or '.pem'.
> > But the rest looks good.
> 
> Who could we ask about that?

I know that `.crt` and `.pem` is working. I do not know if `.0` would work. So I would always suggest to use a suffic like .crt or .pem.

PR #206, opened 2020-04-15
fix and enhance extract of /etc/salt/master.d/susemanager.conf

keichwa:
close https://github.com/SUSE/spacewalk/issues/11217

### DIFF
--- a/modules/salt/pages/salt-file-locations.adoc
+++ b/modules/salt/pages/salt-file-locations.adoc
@@ -3,7 +3,7 @@
 
 
 There are several ways to set up the Salt file structure.
-Here it is described how Salt is supported and set up as part of {productname} Server.
+This section describes how Salt is supported and set up as part of {productname} Server.
 The main configuration file is [path]``/etc/salt/master.d/susemanager.conf``.
 
 [NOTE]


keichwa:
Do we need more reviewers?

modules/salt/pages/salt-file-locations.adoc
```diff
@@ -52,7 +62,8 @@ Do not edit or add custom Salt data to this directory.
 
 [path]``/usr/share/salt-formulas/states``::
 [path]``/usr/share/susemanager/formulas/states``::
-These directories are shipped and updated with {productname} or additional extensions. They inclucde states for salt formulas.
+These directories are shipped and updated with {productname} or additional extensions.
+They inclucde states for salt formulas.
```
Loquacity, line 66:
```suggestion
They include states for Salt formulas.
```

Loquacity:
> Do we need more reviewers?

@mcalmer maybe?

mcalmer:
I am not really sure, what we want to achieve with this?

This config file is managed by suse manager.  When we change something here, it will overwrite all changes made by the customer. We should say that. The old version will be stored as susemanager.conf.rpmsave and the customer is responsible for merging changes back.

But it would be better if they would use a new configfile. Multiple files are 'merged'. This should be used whenever possible. 

Finally not all config options we use are documented here. What was the criteria to select them?

keichwa:
The idea probably is to provide a little bit of reference documentation here: where are the default Salt master configuration files, which files is the user allowed to change, where are the default and custom files that keep the Salt clients up and running (states, pillars, etc.).

Here we do not want to cover all and everything but just the basics.

mcalmer:
> The idea probably is to provide a little bit of reference documentation here: where are the default Salt master configuration files, which files is the user allowed to change, where are the default and custom files that keep the Salt clients up and running (states, pillars, etc.).
> 
> Here we do not want to cover all and everything but just the basics.

Ok, maybe we should write somewhere that you should prevent changes in this file "when possible". In most cases customer could add another config file in that directory. There are some cases where you must change a value we already set in susemanager.conf. If this is the case, the customer must change it in our file. But he should be aware that an update of spacewalk-setup package may overwrite his changes. 
The file is marked as `%config` . As soon as "we" change the content of that file in our git repo, the update of the package will replace it with our new version. If we do not touch the file, it will not be touched on the customer system as well. So not every package update will replace it, but it may happen.

More exact information under which conditions a customer must do the changes in our file and when he can do it in an extra file, can be provided by the salt team.

keichwa:
thanks @mcalmer I added some basic warnings.  Hope I got it right.

modules/salt/pages/salt-file-locations.adoc
```diff
@@ -2,12 +2,23 @@
...
-This diagram shows the Salt file structure, as it is used by the {productname} Server.
-The files are listed in the [path]``/etc/salt/master.d/susemanager.conf`` configuration file.
+There are several ways to set up the Salt file structure.
+Here it is described how Salt is supported and set up as part of {productname} Server.
+The main configuration file is [path]``/etc/salt/master.d/susemanager.conf``.
 
-----
+[NOTE]
+====
+To avoid trouble in the future better do not touch [path]``/etc/salt/master.d/susemanager.conf``.
```
Loquacity, line 11:
```suggestion
Do not edit the [path]``/etc/salt/master.d/susemanager.conf`` configuration file.
```

modules/salt/pages/salt-file-locations.adoc
```diff
@@ -2,12 +2,23 @@
...
+This file belongs to the [package]``spacewalk-setup`` package and is marked as [literal]``%config``.
+When {suse} updates [package]``spacewalk-setup`` [path]``susemanager.conf`` can get overwritten and customization can get lost.
```
Loquacity, line 13:
```suggestion
When {suse} updates the [package]``spacewalk-setup`` package, the [path]``susemanager.conf`` file is overwritten, and any customization is lost.
```

modules/salt/pages/salt-file-locations.adoc
```diff
@@ -2,12 +2,23 @@
...
+Whenever possible it is recommended to add own configuration file to the [path]``/etc/salt/master.d/`` directory, and thus preventing the update process from deleting own settings from the main [path]``susemanager.conf`` configuration file.
```
Loquacity, line 14:
```suggestion
Instead, add your own configuration file to the [path]``/etc/salt/master.d/`` directory.
This prevents the update process from deleting your settings from the main [path]``susemanager.conf`` configuration file.
```

modules/salt/pages/salt-file-locations.adoc
```diff
@@ -2,12 +2,23 @@
...
+====
+
+Here {suse} lists some settings from [path]``/etc/salt/master.d/susemanager.conf`` that will help with finding configuration options:
```
Loquacity, line 17:
```suggestion
Some settings from [path]``/etc/salt/master.d/susemanager.conf`` that can help with finding configuration options:
```

modules/salt/pages/salt-file-locations.adoc
```diff
@@ -2,12 +2,23 @@
 = Salt File Locations and Structure
 
 
-This diagram shows the Salt file structure, as it is used by the {productname} Server.
-The files are listed in the [path]``/etc/salt/master.d/susemanager.conf`` configuration file.
+There are several ways to set up the Salt file structure.
+Here it is described how Salt is supported and set up as part of {productname} Server.
```
Loquacity, line 6:
```suggestion
This section describes how Salt is supported and set up as part of {productname} Server.
```

keichwa:
thanks a million, Lana!

PR #211, opened 2020-04-17
bootstrapping with SSH

Loquacity:
https://github.com/SUSE/spacewalk/issues/11235

### DIFF
--- a/modules/reference/pages/systems/bootstrapping.adoc
+++ b/modules/reference/pages/systems/bootstrapping.adoc
@@ -1,7 +1,7 @@
 [[ref.webui.systems.bootstrapping]]
 = Bootstrapping [Salt]
 
-The [guimenu]``Bootstrapp Minions`` page allows you to bootstrap Salt clients from the {webui}.
+The [guimenu]``Bootstrap Minions`` page allows you to bootstrap Salt clients from the {webui}.
 
 .Bootstrapping
 


Loquacity:
@jcayouette Note on the changelog entry: I figure if we add it in the PR, then it will merge when the PR merges, and we don't need to make a separate change (which should make it easier overall). I guess the real problem will be conflicts, but (at least in theory) they should be easy enough to manage. Let's see how it goes.

.changelog
```diff
@@ -7,3 +7,4 @@
 
 * Added note about reserved characters in Proxy passwords(bsc#1134836)
 * Syntax and presentation updates to Salt Guide
+* Added Bootstrapping with SSH content
```
hustodemon, line 10: I'd be specific, that this is about SSH key (bootstrapping via SSH (with password) was already supported before).
```suggestion
* Added Bootstrapping with SSH key content
```

modules/client-configuration/pages/salt-client-proxy.adoc
```diff
@@ -19,7 +19,12 @@ For information about creating a bootstrap repository, see xref:client-configura
 By default, the SSH  port is [systemitem]``22``.
 . In the [guimenu]``User`` field, type the username to log in to the client.
 By default, the username is [systemitem]``root``.
-. In the [guimenu]``Password`` field, type password to log in to the client.
+. In the [guimenu]``Authentication Method`` field, select the authentication method to use for bootstrapping the client.
++
+* For password authentication, in the [guimenu]``Password`` field, type password to log in to the client.
+* For SSH Private key authentication, enter the private key and the associated passphrase.
+The key is only stored for as long as the bootstrapping process takes to complete.
+The key is deleted after bootstrapping is successful.
```
hustodemon, line 27: Nitpick: the key is also deleted, when the bootstrapping is unsuccessful :-), so either I'd suggest using something more general ("..the key is deleted after bootstrapping"), or I'd even leave this sentence out (the line 26 already says that).

modules/reference/pages/systems/bootstrapping.adoc
```diff
@@ -20,8 +20,20 @@ User::
...
+// I've added these new bits in the existing format for now, but this whole section will get overhauled with the Ref Guide updates. LKB --2020-04-17
+Authentication Method::
+Select either ``Password`` or ``SSH Private Key``.
+
 Password::
-Input the clients login password.
+For password authentication, enter the client's login password.
+
+SSH Private Key::
+For SSH authentication, copy the SSH private key.
```
hustodemon, line 31:
```suggestion
For SSH key authentication, copy the SSH private key.
```
As mentioned in the changelog, I'd be explicit about the key here (the "SSH authentication" happens also in the password case).

modules/reference/pages/systems/bootstrapping.adoc
```diff
@@ -20,8 +20,20 @@ User::
...
+The key is only stored for as long as the bootstrapping process takes to complete.
+The key is deleted after bootstrapping is successful.
```
hustodemon, line 33: Same as above ^^^.

modules/reference/pages/systems/bootstrapping.adoc
```diff
@@ -20,8 +20,20 @@ User::
...
+
+SSH Private Key Passphrase::
+For SSH authentication, copy the passphrase for the private key.
```
hustodemon, line 36: I'd say "enter" - the passphrase is basically a password, which users type in.
```suggestion
For SSH authentication, enter the passphrase for the private key.
```

modules/reference/pages/systems/bootstrapping.adoc
```diff
@@ -30,6 +42,7 @@ Disable SSH Strict Key Host Checking::
 This check box is selected by default.
 This allows the script to auto-accept host keys without requiring a user to manually authenticate.
 
+// Aside from the terrible formatting here, does this need the tech preview notice removed? --LKB 2020-04-17
```
hustodemon, line 45: Wow, I didn't know that we still have "minion-less" salt clients as a tech preview. Is this true @renner ?

Loquacity: I suspect it's more that we failed to take the note out some time ago, honestly ...

modules/reference/pages/systems/bootstrapping.adoc
```diff
@@ -30,11 +41,9 @@ Disable SSH Strict Key Host Checking::
 This check box is selected by default.
 This allows the script to auto-accept host keys without requiring a user to manually authenticate.
 
+
```
keichwa, line 44:
```suggestion
```

modules/reference/pages/systems/bootstrapping.adoc
```diff
@@ -30,11 +41,9 @@ Disable SSH Strict Key Host Checking::
...
 Manage System Completely via SSH (Will not Install an Agent)::
 +
```
keichwa, line 46:
```suggestion
```
I think + i not needed here...

PR #226, opened 2020-04-23
Upgrade Guide - Arch Review

Loquacity:
PDF Version here: 
[suse_manager_upgrade_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/4520060/suse_manager_upgrade_guide.pdf)

No content changes, with the exception of updating/writing a few intros.

image
image


Loquacity:
Some exposition to give background on why I made the changes:

When we originally designed this book, it was supposed to be organised by component first (Server, Proxy, Client, DB), then upgrade type second (major, minor/SP, MU). This has largely remained true but, in the HTML at least, the levels had gone missing, so there was no section nesting to visually separate the components. So I fixed that as a first step.

When I looked through the content as it previously was, it struck me that we don't use consistent naming for any of the processes in this book. Not just upgrade/update/migrate (which we've previously discussed), but major and minor versions, SP and MU upgrades, etc etc. In the original introduction, we had a pretty good breakdown and explanation of the X.Y.Z schema, so I decided to use that as the touchstone. I updated that slightly to make it clearer, then renamed the various sections, split a couple where appropriate, and created new intro levels for each component.

So the end result is that we have four clear top level sections: Server, Proxy, Client, DB (& Troubleshooting)
Once you pick which component you want to upgrade, you can pick the type of upgrade you want: X, Y, or Z (for the DB, it's version 10 or version 12, which technically X but I thought that might be taking it a little *too* far :wink: ).

Let me know what you think.

Loquacity:
Oh, and an addendum: this is going to break a whole lot of xrefs, but I'm being lazy and waiting for the tests to tell me about them.

keichwa:
The clever strategy is that we now label everything as "upgrade".  Yes, that's a good idea! We should do this throughout the texts, too. This means the "Server - Major Version" / "Server - X Upgrade" section better starts with "Upgrading SUSE Manager Server ..." instead of "Migrating SUSE Manager Server ..." etc. pp.

keichwa:
The other thing is that Uyuni is different with the version numbers and for Proxy/Client the x.y.z concept does not make that much sense ;)  I still believe we'd better offer dedicated proxy and client guides where we also discuss upgrading those systems.  But for the moment, we can live with the general Installation and Upgrade Guides.

modules/upgrade/pages/client-intro.adoc
```diff
@@ -0,0 +1,10 @@
+[[client-migration]]
+= Upgrade the Clients
+
+
+Clients use the versioning system of their underlying operating system.
+For most clients using {suse} operating systems, you can perform minor upgrade within the {productname} {webui}.
```
keichwa, line 6: I think we can safely say:
```suggestion
For clients using {suse} operating systems, you can perform upgrades within the {productname} {webui}.
```
This would cover package and SP upgrades.

modules/upgrade/pages/client-intro.adoc
```diff
@@ -0,0 +1,10 @@
...
+
+To upgrade clients running SLE{nbsp}12 to SLE{nbsp}15, the upgrade is automated, but you will need to do some preparation steps before you begin.
```
keichwa, line 8: Maybe, we could generalize this?  It's major version upgrade and should also be valid for Leap upgrades.

modules/upgrade/pages/client-x.adoc
```diff
@@ -1,30 +1,13 @@
...
-Upgrading traditional clients from SLE{nbsp}12 with the latest Service Pack to SLE{nbsp}15 is automated, but you will need to do some preparation steps before you can upgrade.
-
-Upgrading traditional clients from earlier SLE{nbsp}15 Service Packs to later SLE{nbsp}15 Service Packs is fully automated.
-
-For supported {sle}{nbsp}15 upgrade paths, see https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-upgrade-paths.html
-
-
-
-== Upgrade Clients from SLE{nbsp}12 to SLE{nbsp}15
+= Client - X Upgrade
```
keichwa, line 2: In the nav, we have "Major Version" instead of "X Upgrade".  This also applies for all the other sections.

modules/upgrade/pages/client-x.adoc
```diff
@@ -115,43 +100,6 @@ PXE booting is not used during this process.
 Approximately three minutes after the job was fetched, the client will go down for reboot.
 
 
-////
```
keichwa, line 118: I think we want to enable this now because we now must cover the SLE 15 to SLE 15 SP1 upgrade etc. Maybe, to be moved in a client-y file.

Loquacity:
> The clever strategy is that we now label everything as "upgrade". Yes, that's a good idea! We should do this throughout the texts, too. This means the "Server - Major Version" / "Server - X Upgrade" section better starts with "Upgrading SUSE Manager Server ..." instead of "Migrating SUSE Manager Server ..." etc. pp.

Yeah, I haven't done anything other than some minor intro edits. I think this is a good idea, but this PR is probably big enough already :wink:

modules/upgrade/pages/client-x.adoc
Loquacity: Yeah, and it's "X Upgrade" in the PDF. I'll update the nav to read "X Upgrade", etc.

modules/upgrade/pages/client-x.adoc
Loquacity: Sure

keichwa:
proxy-y-z.adoc lacks https://github.com/uyuni-project/uyuni-docs/commit/58f2ddfcd30aa75959aaeac99407db2765177850 .  It is also not in current master, this probably means something went wrong with my squash and merge action...

Loquacity:
> The other thing is that Uyuni is different with the version numbers and for Proxy/Client the x.y.z concept does not make that much sense ;) I still believe we'd better offer dedicated proxy and client guides where we also discuss upgrading those systems. But for the moment, we can live with the general Installation and Upgrade Guides.

Yeah, I think when we split the Install Guide, we would be well-served to split the Upgrade Guide as well. Future problem!

modules/upgrade/pages/client-intro.adoc
Loquacity: Yes, agreed

Loquacity:
> proxy-y-z.adoc lacks [58f2ddf](https://github.com/uyuni-project/uyuni-docs/commit/58f2ddfcd30aa75959aaeac99407db2765177850) . It is also not in current master, this probably means something went wrong with my squash and merge action...

I have no idea what's going on there.

keichwa:
> I have no idea what's going on there.

I did not create a proper PR :-(  now it is ready to be approved...

Issue #204, opened 2020-04-14
Proxy update guide wrong?


atorlee:
Going through https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/upgrade/proxy-update.html I noticed that one is supposed to execute the spacewalk-proxy command on the SUSE Manager Server: 

--- cut here ---
On the SUSE Manager Server where the proxy is registered as a client, stop the proxy service:
spacewalk-proxy stop
[...]
On the SUSE Manager Server where the proxy is registered as a client, start the proxy service:
spacewalk-proxy start
--- cut here ---

I believe the spacewalk-proxy command is to be executed on the SUSE Manager Proxy

Loquacity:
Hi @atorlee thanks for the bug report! We're looking into it now :)

PR #228, opened 2020-04-23
stop and start proxy service on the proxy

keichwa:
https://github.com/uyuni-project/uyuni-docs/issues/204


keichwa:
thanks you both!

PR #233, opened 2020-04-24
In-Place Client Upgrades

Loquacity:
https://github.com/SUSE/spacewalk/issues/11272

### DIFF
--- a/modules/upgrade/pages/client-y-z.adoc
+++ b/modules/upgrade/pages/client-y-z.adoc
@@ -51,7 +51,7 @@ Use these variables in your autointallation profiles for automating in-place upg
 
 
 
-.Example: Variables for use in Autoinstallation Profiles
+.Example: Variables for Use in Autoinstallation Profiles
 ----
 registration_key=1-15sp1-demo-test
 org=1
@@ -61,7 +61,7 @@ distro_label=15sp1-demo-test
 
 
 
-.Example: Entry for use in Autoinstallation Profiles
+.Example: Entry for Use in Autoinstallation Profiles
 ----
  <listentry>
      <ask_on_error config:type="boolean">true</ask_on_error>


Loquacity:
Email sent asking for review, so I'm going to leave this out here a little longer to see if we can get further feedback.

modules/upgrade/pages/client-y-z.adoc
```diff
@@ -0,0 +1,106 @@
...
+. You can use variables in the profile to distinguish between the different lifecycle environments.
+
+
+For example autoinstallation profiles, see xref:upgrade:upgrade-script-example.adoc[].
+
+Use these variables in your autointallation profiles for automating in-place upgrades:
+
+
+
+.Example: Variables for use in Autoinstallation Profiles
```
keichwa, line 54:
```suggestion
.Example: Variables for Use in Autoinstallation Profiles
```

modules/upgrade/pages/client-y-z.adoc
```diff
@@ -0,0 +1,106 @@
...
+----
+registration_key=1-15sp1-demo-test
+org=1
+channel_prefix=15sp1-demo-test
+distro_label=15sp1-demo-test
+----
+
+
+
+.Example: Entry for use in Autoinstallation Profiles
```
keichwa, line 64:
```suggestion
.Example: Entry for Use in Autoinstallation Profiles
```

modules/upgrade/pages/client-y-z.adoc
```diff
@@ -0,0 +1,106 @@
...
+
+== Upgrade
+
+
+When you have prepared the server for the upgrade, you can provision the clients.
+
+
+
+.Procedure: Provisioning the Clients
+. In the {productname} {webui}, navigate to menu:Systems[System List], and select the clients you want to provision to add them to the system set manager.
```
keichwa, line 84: This "to ... to ... to" is not that easy to understand.  Here is an attempt to split the sentence:
```suggestion
. In the {productname} {webui}, navigate to menu:Systems[System List], and select the clients you want to provision.
Selecting them you add them to the system set manager.
```

Loquacity:
> Only minor things.
> 
> I think we should just use one term: either "lifecycle project" or "lifecycle environment".

Those are two different things, though?

modules/upgrade/pages/client-y-z.adoc
Loquacity: That doesn't quite work either. What about:

> Selecting the clients adds them to the system set manager.


??

mcalmer:
> > Only minor things.
> > I think we should just use one term: either "lifecycle project" or "lifecycle environment".
> 
> Those are two different things, though?

Correct, a project has multiple environments.

modules/upgrade/pages/client-y-z.adoc
keichwa: I'm not sure about the right wording, but I think it works as follows:

1/ You select the clients (you tick them)
2/ Then all selected clients are ready to be managed by system set manager routines

Maybe better just ignore my suggestion.  I think the initial is technically correct and we can keep it. ;)

keichwa:
On 2020-04-28 01:52, Lana Brindley wrote:

> Those are two different things, though?

I'm not an expert with these things.  You are probably right.

-- 
Karl Eichwalder

modules/upgrade/pages/client-y-z.adoc
Loquacity: I think any technical differences are pretty minor here. I merged it all into one step because we describe it in detail in the SSM chapter, and it's the kind of task you are likely to do over and over again.

PR #241, opened 2020-04-28
create section for download endpoint

keichwa:
I think the endpoint configuration deserve a section of its own.


modules/salt/pages/salt-states.adoc
```diff
@@ -121,6 +121,9 @@ salt -I 'group_ids:GID' state.apply
 ----
 
 
+
+== Download Endpoint ==
```
Loquacity, line 125:
```suggestion
== Download Endpoint
```

Your docbook is showing :wink:

PR #247, opened 2020-05-04
Upgrade check pg #11380 keichwa

keichwa:
Is this all what we need to fix the reported bug?
https://bugzilla.suse.com/show_bug.cgi?id=1170711
https://github.com/SUSE/spacewalk/issues/11380


modules/upgrade/pages/upgrade-overview.adoc
```diff
@@ -13,8 +13,8 @@ It is possible to automate some of the upgrades, but others need to be performed
 This guide is not intended to be read cover to cover.
 Instead, navigate to the component you want to upgrade, then identify the versions you are upgrading from and to.
 ====
-
-
+
```
keichwa, line 16:
```suggestion

```

Loquacity:
If I'm not mistaken, the bug says we need to be more clear about requiring a pg upgrade. You've added the two pages that were missing from the nav (thanks!), but neither of those pages actually seem to mention pg directly. Can we maybe add an admonition or something to call out the pg upgrade, with an xref to the appropriate chapter?

keichwa:
On 2020-05-05 02:35, Lana Brindley wrote:
> If I'm not mistaken, the bug says we need to be more clear about
> requiring a pg upgrade. You've added the two pages that were missing
> from the nav (thanks!), but neither of those pages actually seem to
> mention pg directly.

I think the important step is that we mention the migration script in
these files (fingers crossed that the formatting will make it through):

```
. When rebooted the spacewalk services are not running until you have 
migrated
the PostgreSQL database to version{nbsp}12.
. Log in on the text console as root, and run the database migration 
script:
+
----
/usr/lib/susemanager/bin/pg-migrate-10-to-12.sh
----
```

> Can we maybe add an admonition or something to
> call out the pg upgrade, with an xref to the appropriate chapter?

This is a good idea.  I'll add the xref!

-- 
                                    Maxfeldstraße 5
                                    90409 Nürnberg, Germany

keichwa:
It's now ready for review, please.

PR #260, opened 2020-05-06
QSPC Tech Review

Loquacity:
Feedback on one.suse.com

### DIFF
--- a/modules/quickstart-public-cloud/pages/clients.adoc
+++ b/modules/quickstart-public-cloud/pages/clients.adoc
@@ -7,7 +7,7 @@ In the {productname} {webui}, navigate to menu:Systems[Bootstrapping], then fill
 Make sure you use stable FQDNs for the ``Host`` field, or {productname} will not be able to find your host when your Public Cloud gives you a different short-lived FQDNS.
 
 Public cloud images usually do not allow SSH login with username and password, but only SSH with a certificate.
-If you want to use bootstrap from the {webui}, you will need to enable SSH login with user and password.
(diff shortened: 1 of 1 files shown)


modules/quickstart-public-cloud/pages/clients.adoc
```diff
@@ -3,24 +3,14 @@
...
 
-Also, please note public cloud images usually do not allow SSH login with user and password but only SSH certificate login. If you want to use the bootstrap from the UI, you will need to enable SSH login with user and password:
-<describe the required steps>
-
-{productname} will support bootstrapping from the UI with SSH certificates in a future release.
-
-Alternatively, you can use a bootstrap script.
-
-When the bootstrap process has completed, your client will be listed at menu:Systems[System List].
+Public cloud images usually do not allow SSH login with username and password, but only SSH with a certificate.
```
keichwa, line 9:
```suggestion
Public cloud images usually do not allow SSH login with user and password, but only SSH with a certificate.
```
At other places nearby we seem to use just "user"

Loquacity: but a user is a person, whereas a username is a piece of information that can be entered into a field?

keichwa:
On 2020-05-07 02:13, Lana Brindley wrote:

> but a user is a person, whereas a username is a piece of information
> that can be entered into a field?

I fear you are right.  Maybe, we should consolidate in this directory.
We probably must check the web UI, too.

-- 
                                    Maxfeldstraße 5
                                    90409 Nürnberg, Germany

PR #265, opened 2020-05-07
client reqs

Loquacity:
https://github.com/SUSE/spacewalk/issues/10344

This needs an SME, @mcalmer can you help?

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -29,3 +29,4 @@ Updated Client Configuration Guide and Reference. (bsc#1164451)
 * Added admonition about changing auth method to SSO (bsc#1170818)
 * Documented gitfs in Salt Guide
 * Added client hardware requirements to Installation Guide
+* Updated Red Hat Client chapter in Client Configuration Guide (bsc#1163516)
--- a/modules/administration/pages/image-management.adoc
+++ b/modules/administration/pages/image-management.adoc
@@ -255,12 +255,12 @@ An Image Profile that can be reused is published at https://github.com/SUSE/mana
 
 [NOTE]
 ====
-The [option]``ARG`` parameters ensure that the built image is associated with the desired repository version served by {productname}.
+The [option]``ARG`` parameters ensure that the built image is associated with the desired repository served by {productname}.
 The [option]``ARG`` parameters also allow you to build image versions of {sles} which may differ from the version of {sles} used by the build host itself.
 
 For example: The [comand]``ARG repo`` parameter and the [command]``echo`` command pointing to the repository file, creates and then injects the correct path into the repository file for the desired channel version.
 
-The repository version is determined by the activation key that you assigned to your image profile.
+The repository is determined by the activation key that you assigned to your image profile.
 ====
 
 [NOTE]
@@ -299,6 +299,14 @@ RUN /root/add_packages.sh
 # After building remove the repository path from image
 RUN rm -f /etc/zypp/repos.d/susemanager:dockerbuild.repo
 ----
+
+//TODO: Replace the "custom-system-info" link
+.Using Custom Info Key-value Pairs as Docker Buildargs
+You can assign custom info key-value pairs to attach information to the image profiles.
+Additionally, these key-value pairs are passed to the Docker build command as `buildargs`.
+
+For more information about the available custom info keys and creating additional ones, see xref:reference:systems/custom-system-info.adoc[].
+
 
 [[at.images.docker.building]]
 === Build an Image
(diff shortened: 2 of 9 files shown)


Loquacity:
> I usually manage clients with less hardware resources. It would also be good if we'd add other supported architectures.
> 
> [Edited: creating a separate issue.]

I really just plucked numbers out of the air, so i'm looking for guidance on what's reasonable here. Arch is in the table above this one.

keichwa:
On 2020-05-07 07:44, Lana Brindley wrote:

> I really just plucked numbers out of the air, so i'm looking for
> guidance on what's reasonable here. Arch is in the table above this
> one.

yes, that's how read it.  i guess in the end we'll say something such
as: if your client runs fine with its workload the Salt/SUSE Manager
overhead is minimal.  Of course, our product gurus will have to check
this.

-- 
                                    Maxfeldstraße 5
                                    90409 Nürnberg, Germany

modules/installation/pages/client-requirements.adoc
```diff
@@ -21,15 +21,33 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
...
 | {opensuse} Leap 15.1 | {x86}_64                       | icon:times[role="danger"]      | icon:check[role="green"]
-| {ubuntu} 16.04 | {x86}_64                             | icon:times[role="danger"]      | icon:check[role="green"]
+| {ubuntu} 20.04 | {x86}_64                             | icon:times[role="danger"]      | icon:check[role="green"]
 | {ubuntu} 18.04 | {x86}_64                             | icon:times[role="danger"]      | icon:check[role="green"]
+| {ubuntu} 16.04 | {x86}_64                             | icon:times[role="danger"]      | icon:check[role="green"]
+|===
+
+
+
+For most clients, minimum hardware is required.
```
mcalmer, line 42: I would suggest to explain more what the management stack need on top of the normal OS and workload.

CPU: the OS needs at least 1. And I think we do not need more just because we want to manage it with salt.
RAM: the OS has minimal and recommended values. Maybe you can check the recommended values of the OSes but I think we do not need to increase it. Be careful with "minimal" values. AFAIK the SLES minimal does not really allow to "update" the client as the metadata of the update channels cause OOM exceptions.
SWAP: just standard recommendation. It is a safety value in case the RAM is not big enough. You can also put RAM+SWAP together as virtual memory. This would also cover s390x
Disk: this is difficult as we do not know what the customer want to run on this. So here I would suggest to say something about the size of the management stack. But even this is not easy. So maybe we can say 200MB on top of a minimal OS installation or something like this.

paususe: > I would suggest to explain more what the management stack need on top of the normal OS and workload.

In fact, I would *only* say that about client requirements. Something in the lines of "whatever your workload needs, plus 200 MB RAM for Salt or X MB RAM for the traditional stack. No additional CPU core required unless your client already struggles with the workload".

modules/installation/pages/client-requirements.adoc
```diff
@@ -21,15 +21,34 @@ For details on supported product versions, see https://www.suse.com/lifecycle.
...
+
+[[clients.hw.reqs]]
+[cols="1,1", options="header"]
+.Client Additional Hardware Requirements
+|===
+| Hardware               | Recommended
+| CPU:                   | 1 64-bit CPU core ({x86_64})
+| RAM                    | 2{nbsp}GB
+| Swap space             | 2{nbsp}GB
+| Disk Space:            | 200{nbsp}MB
```
mcalmer, line 53: I am not 100% happy with this, as it mix minimal values with additional. It give the impression that only our mamagement stack requires 2GB RAM + 2GB Swap.

Suggestion:

```suggestion
.Client Hardware Requirements
|===
| Hardware               | Size
| CPU:                   | minimal 1 64-bit CPU core ({x86_64}) for the system
| RAM                    | minimal 2{nbsp}GB for the system
| Swap space             | recommended 2{nbsp}GB
| Disk Space:            | + 200{nbsp}MB for the management stack
```

Alternative we guess a bit:
- drop CPU (as any value fit)
- RAM: +512MB for the management stack
- drop SWAP space
- Disk Space: 200MB for the management stack

I have no idea if 512MB just for the management stack is sufficient or still over commited.

Loquacity: Heh, this is harder than I expected it to be :wink: 
I'll take another look on Monday and see what we can come up with. I don't think there's any rush to get this in.

PR #243, opened 2020-04-29
Updates for Ubuntu 20.04

Loquacity:
**WARNING:** Do not merge this PR until #267 (or something to disable/remove Ubuntu 20.04 on Uyuni) is merged!!!

https://github.com/SUSE/spacewalk/issues/11361

PR #267, opened 2020-05-08
Disable Ubuntu 20.04 for Uyuni, and some minor fixes

juliogonzalez:
**WARNING**: Do not merge https://github.com/uyuni-project/uyuni-docs/pull/243 until this PR (or something to disable/remove Ubuntu 20.04 on Uyuni) is merged!!!

I generated both docs, and looked good, but please have a review as well.

As you see, I commented where I could, and added `if`s where I couldn't.

I really think this situation is better than rolling back, specially because there are some places where you can't just remove lines. Enabling this when we have Ubuntu 20.04 on Uyuni will be really quick.

### DIFF
--- a/modules/client-configuration/pages/clients-ubuntu.adoc
+++ b/modules/client-configuration/pages/clients-ubuntu.adoc
@@ -3,7 +3,13 @@
 
 This section contains information about registering Salt clients running {ubuntu} operating systems.
 
+// Remove the ifs and the uyuni part when we add the Ubuntu 20.04 there (Leap 15.2 server is require first)
+ifeval::[{suma-content} == true]
 {productname} supports {ubuntu} 16.04 LTS, 18.04 LTS, and 20.04 LTS clients using Salt.
+endif::[]
+ifeval::[{uyuni-content} == true]
+{productname} supports {ubuntu} 16.04 LTS and 18.04 LTS clients using Salt.
+endif::[]
 Traditional clients are not supported.
 
 [IMPORTANT]
@@ -165,13 +171,11 @@ endif::[]
 ifeval::[{suma-content} == true]
 .Procedure: Alternative Method for Adding {ubuntu} Channels and Repositories
 . At the command prompt on the {productname} Server, as root, install the [systemitem]``spacewalk-utils`` package:
-+
 ----
 zypper in spacewalk-utils
 ----
 . Add the {ubuntu} channels.
 For {ubuntu} 16.04:
-+
 ----
 spacewalk-common-channels \
 ubuntu-1604-amd64-main \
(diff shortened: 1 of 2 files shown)


modules/client-configuration/pages/clients-ubuntu.adoc
```diff
@@ -165,13 +171,11 @@ endif::[]
 ifeval::[{suma-content} == true]
 .Procedure: Alternative Method for Adding {ubuntu} Channels and Repositories
 . At the command prompt on the {productname} Server, as root, install the [systemitem]``spacewalk-utils`` package:
-+
```
juliogonzalez, line 168: I removed them, at the generated HTML doc the `+` was shown, and I guess that is not wanted.

modules/client-configuration/pages/clients-ubuntu.adoc
```diff
@@ -247,18 +247,18 @@ ubuntu-1804-amd64-universe-updates-uyuni \
...
-spacewalk-common-channels \
-ubuntu-2004-pool-amd64-uyuni \
-ubuntu-2004-amd64-main-uyuni \
-ubuntu-2004-amd64-main-updates-uyuni \
-ubuntu-2004-amd64-main-security-uyuni \
-ubuntu-2004-amd64-universe-uyuni \
-ubuntu-2004-amd64-universe-updates-uyuni \
-ubuntu-2004-amd64-uyuni-client
-----
+// Uncomment the following lines we add the Ubuntu 20.04 there (Leap 15.2 server is require first)
```
juliogonzalez, line 250: Commented as agreed.

modules/client-configuration/pages/supported-features-ubuntu.adoc
```diff
@@ -28,7 +30,7 @@ The icons in this table indicate:
 | Operating system packages | icon:times[role="danger"] | icon:times[role="danger"] | icon:times[role="danger"]
 | Registration | ``Salt`` | ``Salt`` | ``Salt``
 | Install packages | ``Salt`` | ``Salt`` | ``Salt``
-| Apply patches | icon:question[role="gray"] | icon:question[role="gray"] | icon:question[role="gray"]
```
juliogonzalez, line 31: Not sure what the character before `Apply` was, but it wasn't a whitespace:

image

modules/client-configuration/pages/clients-ubuntu.adoc
Loquacity: Removing them will mean the code block pops out of the step block, though, and then renumbering will start again. All that said, they shouldn't be showing either :woman_shrugging:

juliogonzalez: Let's merge, and maybe you can have a look on your branch :-)

PR #278, opened 2020-05-14
add system locking chapter

Loquacity:
https://github.com/SUSE/spacewalk/issues/11374

~Draft for now until I can access the reference server.~

@mbologna am I close? :wink:

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -6,4 +6,5 @@
 # Last packaged: 2020-05-19
 
 - Add System Locking chapter to Client Configuration Guide
+- Update OS support notes in Client Configuration Guide
 - Small edits to the Large Scale Deployments Guide (bsc#1171389)
--- a/modules/client-configuration/pages/supported-features-centos.adoc
+++ b/modules/client-configuration/pages/supported-features-centos.adoc
@@ -4,12 +4,14 @@
 
 This table lists the availability of various features on {CentOS} clients.
 
-[WARNING]
+
+[IMPORTANT]
 ====
-{centos} is not currently an officially supported client operating system.
-It may or may not be supported in a future version of {productname}.
-However, {centos} with an {es} subscription is supported.
+The operating system you run on a client is supported by the organization that supplies the operating system.
+{centos} is supported by the {centos} community.
 ====
+
+
 
 The icons in this table indicate:
 
@@ -24,8 +26,8 @@ The icons in this table indicate:
 .Supported Features on {centos} Operating Systems
 |===
 | Feature | {centos}{nbsp}6 | {centos}{nbsp}7 | {centos}{nbsp}8
-| Client | icon:check[role="green"] | icon:check[role="green"] | ``Salt``
-| Operating system packages  |  icon:times[role="danger"] (plain CentOS) / icon:check[role="green"] (with {es}) | icon:times[role="danger"] (plain CentOS)/icon:check[role="green"] (with {es}) | icon:times[role="danger"] (plain CentOS)/``Salt`` (with {es})
+| Client | icon:question[role="gray"] (plain CentOS) / icon:check[role="green"] (with {es}) | icon:question[role="gray"] (plain CentOS) / icon:check[role="green"] (with {es}) | icon:question[role="gray"] (plain CentOS) / ``Salt`` (with {es})
+| System packages |  {centos} Community / {suse} (with {es}) | {centos} Community / {suse} (with {es}) | {centos} Community / {suse} (with {es})
 | Registration | icon:check[role="green"] | icon:check[role="green"] | ``Salt``
 | Install packages | icon:check[role="green"] | icon:check[role="green"] | ``Salt``
 | Apply patches (requires CVE ID) | icon:check[role="green"] (third-party service required for errata) | icon:check[role="green"] (third-party service required for errata) | ``Salt`` (third-party service required for errata)
--- a/modules/client-configuration/pages/supported-features-debian.adoc
+++ b/modules/client-configuration/pages/supported-features-debian.adoc
@@ -11,6 +11,12 @@ ifeval::[{suma-content} == true]
 ====
 endif::[]
 
+[IMPORTANT]
+====
+The operating system you run on a client is supported by the organization that supplies the operating system.
+{debian} is supported by the {debian} community.
+====
+
 The icons in this table indicate:
 
 * icon:check[role="green"] the feature is available on both traditional and Salt clients
@@ -25,7 +31,7 @@ The icons in this table indicate:
 |===
 | Feature | {debian}{nbsp}9 | {debian}{nbsp}10
 | Client | icon:check[role="green"] | icon:check[role="green"]
-| Operating system packages | icon:times[role="danger"] | icon:times[role="danger"]
+| System packages | {debian} Community | {debian} Community
 | Registration | ``Salt`` | ``Salt``
 | Install packages | ``Salt`` | ``Salt``
 | Apply patches | icon:question[role="gray"] | icon:question[role="gray"]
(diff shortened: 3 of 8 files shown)


modules/client-configuration/pages/system-locking.adoc
```diff
@@ -0,0 +1,54 @@
...
+This is useful for clients running production software, or to prevent accidental changes.
+You can disable the system lock when you are ready to perform actions.
+
+System locks are implemented differently on traditional and Salt clients.
+
+
+
+== System Locks on Traditional Clients
+
+When a traditional client is locked, no actions can be scheduled, and a padlock icon is displayed next to the name of the client in the menu:System[System List].
```
mbologna, line 15: Remote commands can still be issued an executed even when a traditional system is locked

modules/client-configuration/pages/system-locking.adoc
```diff
@@ -0,0 +1,54 @@
...
+To stop automated patch updates, navigate to the [guimenu]``System Details`` page for the client, and on the [guimenu]``Properties`` tab, uncheck [guimenu]``Auto Patch Update``.
+====
+
+// All these webui things need to be checked, ref server is down right now. --LKB 20200514
+
+
+
+== System Locks on Salt Clients
+
+When a Salt client is locked, or put into blackout mode, no actions can be scheduled, and a yellow banner is displayed on the [guimenu]``System Details`` page.
```
mbologna, line 36:
```suggestion
When a Salt client is locked, or put into [Salt blackout mode](https://docs.saltstack.com/en/latest/topics/blackout/index.html), no actions can be scheduled, and a yellow banner is displayed on the [guimenu]``System Details`` page. Additionally, any Salt execution command is disabled (except for the ones explicitly allowed in the system lock formula configuration).
```

modules/client-configuration/pages/system-locking.adoc
```diff
@@ -0,0 +1,54 @@
...
+In this mode, actions can be scheduled for the locked client using the {webui} or the API, but the actions will fail.
+
+[NOTE]
+====
+The locking mechanism is not available for Salt SSH clients.
+====
+
+To lock a Salt client using the {productname} {webui}, navigate to the [guimenu]``System Details`` page for the client you want to lock.
+Navigate to the [guimenu]``Formulas`` tab, enable the system lock formula, and apply the highstate.
```
mbologna, line 45:
```suggestion
Navigate to the [guimenu]``Formulas`` tab, enable the system lock formula, and tick the checkbox "Lock system". Finally, save the formula (applying the highstate is not required for this formula).
```

modules/client-configuration/pages/vhm-caasp.adoc
```diff
@@ -18,20 +20,7 @@ If you are using ``cloud-init``, we recommended that you use a bootstrap script
 For more on bootstrapping, see xref:client-configuration:registration-bootstrap.adoc[].
 
 When you have added the {caasp} nodes to {productname}, the registered system will be locked automatically.
```
mbologna, line 22:
```suggestion
When you have added the {caasp} nodes to {productname}, the registered system will automatically apply the Salt formula for system locking.
```

modules/client-configuration/pages/system-locking.adoc
Loquacity: image
It actually tells you to apply the highstate in the UI :confused:

mbologna: Exactly:
image

In other words:

In the first step, you associate the formula with the system. But it's not locked! Association requires a highstate.
In the second step, you lock the system. And that does not require the highstate execution.

Loquacity: ah ha, so the steps go like this:
* Apply the formula
* Apply highstate
* Configure formula (no need to apply highstate again)

??

mateiw: It's a bit confusing. The first message should actually be "... apply the Highstate **if required**."
So the steps should be:
1. Enable the formula
2. Go the the formula and configure it
3. Save it and apply the Highstate only if the message asks you to do that. In the case of the "System Lock" formula it's not actually required.

Loquacity: It doesn't hurt to apply the highstate, though, right? Why don't we just add it as an instruction regardless?

modules/client-configuration/pages/system-locking.adoc
```diff
@@ -14,43 +14,45 @@ System locks are implemented differently on traditional and Salt clients.
...
-Navigate to the menu:Formulas[System Lock] tab, check [guimenu]``Lock system``, and click btn:[Save].
-On this page, you can also specify specific Salt modules to enable them while the Salt client is locked.
-You do not need to apply the highstate.
 
+
+.Procedure: System Locking a Salt Client
+. In the {productname} {webui}, navigate to the [guimenu]``System Details`` page for the client you want to lock.
+. Navigate to the [guimenu]``Formulas`` tab, check the system lock formula, and click btn:[Save].
+. Navigate to the menu:Formulas[System Lock] tab, check [guimenu]``Lock system``, and click btn:[Save].
+On this page, you can also specify specific Salt modules to enable them while the Salt client is locked.
```
keichwa, line 48: unclear IMO.  what's "them"?  "specific" is superfluous (or not helpful enough).

modules/client-configuration/pages/system-locking.adoc
```diff
@@ -14,43 +14,45 @@ System locks are implemented differently on traditional and Salt clients.
...
+. If instructed to by the {webui}, apply the highstate.
```
keichwa, line 49:
```suggestion
. If instructed by the {webui}, apply the highstate.
```

modules/client-configuration/pages/system-locking.adoc
mbologna: Yes, it does not hurt but it would be useless - IMHO we should not add that as an instruction.

modules/client-configuration/pages/system-locking.adoc
Loquacity: OK, so this is interesting. In my original form, I've split the infinitive verb "to apply". Removing the "to" would make the sentence grammatically incorrect, but what we can do is remove the split, which would make it something like "If instructed _to apply_ the highstate by the {webui}, do it". Which, quite frankly, sounds awful. I'm reminded of Orwell's advice to "break any of these rules before doing anything truly barbarous" :joy: I think the only way I can make this sound any good without the split infinitive is to make it more repetitive, like: "If the {webui} instructs you to apply the highstate, apply the highstate." :woman_shrugging:

keichwa:
On 2020-05-19 02:46, Lana Brindley wrote:
> @Loquacity commented on this pull request.

> OK, so this is interesting.

Ok, I now see the dilemma.  Maybe, we can avoid the word
"instructed"?

"If there are changes apply the highstate."

or "... new configuration settings ... "

I'm sure you'll find a good solution :)

-- 
                                    Maxfeldstraße 5
                                    90409 Nürnberg, Germany

Loquacity:
> On 2020-05-19 02:46, Lana Brindley wrote: @Loquacity commented on this pull request. OK, so this is interesting.
> Ok, I now see the dilemma. Maybe, we can avoid the word "instructed"? "If there are changes apply the highstate." or "... new configuration settings ... " I'm sure you'll find a good solution :)

Haha, you have more faith in me than I do in myself, sometimes :wink: 

So we're trying to say that "if you see a banner saying you need to apply the highstate, then you need to apply the highstate. If you don't see a banner, there is no need to apply the highstate."

So, how about something like this:

`When you have made your changes, you might need to apply the highstate.
In this case, a banner in the {webui} will notify you.`

:thinking:

mbologna:
> When you have made your changes, you might need to apply the highstate. In this case, a banner in the {webui} will notify you.

Sounds good to me

PR #282, opened 2020-05-19
Ref Guide - Channel List

Loquacity:
https://github.com/SUSE/spacewalk/issues/10598

@keichwa I will edit the channels.adoc file in the client config separately!

### DIFF
--- a/modules/reference/pages/software/software-menu.adoc
+++ b/modules/reference/pages/software/software-menu.adoc
@@ -1,6 +1,6 @@
 [[ref.webui.software]]
 = Software Menu
 
-The [guimenu]``Software`` section allows you to view and manage software channels and packages on your clients.
+The [guimenu]``Software`` section allows you to view and manage software channels, repositories, and packages.
 
 For more information about software channels, see xref:client-configuration:channels.adoc[].


modules/reference/pages/software/software-channel-list-menu.adoc
```diff
@@ -1,47 +1,29 @@
...
-+
+[[channel-list-filters]]
+[cols="1,1,1", options="header"]
+.Channel List Filters
+|===
+| Filter | Description
+| All         | All channels available to your organization.
+| SUSE        | Channels provided by {suse}.
+| Popular     | Channels most subscribed to by clients in your organization.
+| My Channels | Software channels that belong to your organization, including custom channels.
```
keichwa, line 24: I think this and the description in the webui is confusing.  "belong" is too broad and these were previously known as custom channels (see the file name of the URL).  Thus these don't "include" but "are" custom channels.

modules/reference/pages/software/software-menu.adoc
```diff
@@ -1,4 +1,6 @@
 [[ref.webui.software]]
 = Software Menu
 
-The pages in the menu:Main Menu[Software] category enable you to view and manage software channels and packages associated with your systems.
+The [guimenu]``Software`` section allows you to view and manage software channels and packages on your clients.
```
keichwa, line 4:
```suggestion
The [guimenu]``Software`` section allows you to view and manage software channels, repositories, and packages.
```
With these menus you manage channels on the server, IMO.  This, later on, helps with managing software on the clients.

modules/reference/pages/software/software-channel-list-menu.adoc
Loquacity: Yeah, when I edit the bit in the Admin Guide, I'm going to put more thought into this, I think, and will update this page accordingly.

keichwa: ok, makes sense.

PR #290, opened 2020-05-21
Ref Guide - Package Search

Loquacity:
https://github.com/SUSE/spacewalk/issues/10598


modules/reference/pages/software/software-package-search.adoc
```diff
@@ -1,31 +1,46 @@
...
+| Keyword     | Description | Example
+| name        | Search package names | ``name:SUSE``
+| version     | Search for a package version | ``version:15``
+| filename    | Search within package file names | ``filename:sles``
+| description | Search within the long description | ``description:java``
+| summary     | Search within the one-line summary | ``summary:java``
+| arch        | Search for a package architecture | ``arch:x86_64``
+|===
+
+For example, if you wanted to search all {sle} packages that include `java` in the description and the summary, use this search:
```
keichwa, line 39:
```suggestion
For example, if you want to search all {sle} packages that include `java` in the description and the summary, use this search:
```
but maybe there is a reason why past tense is better here.

Loquacity: nope, should be present. Good spot!

PR #299, opened 2020-05-28
Add info about local repos

Loquacity:
https://github.com/SUSE/spacewalk/issues/10866


Loquacity:
@keichwa @jcayouette @paususe can I get a review on this please?

Loquacity:
> Wondering whether we should say that this a Salt-only feature or add a note about traditional clients.

I can just specify Salt? What do you think?

PR #310, opened 2020-06-04
Admonition Classes 

Namanbhatia7:
Adding additional classes to admonition blocks to modify the three types of admonition used separately to all modules


Loquacity:
Good start!

Loquacity:
Hi @Namanbhatia7 you've picked up a whole lot of other commits here. You will need to revert them.

PR #323, opened 2020-06-12
Styled Tables and Pre block

Namanbhatia7:
Modified the current design of tables and pre blocks.

Made a PR against gsoc-uyuni-them branch.


Loquacity:
Hi Naman! Any chance you can upload some images for me?

Namanbhatia7:
table

Namanbhatia7:
pre

This is pre block

Loquacity:
> pre
> 
> This is pre block

I like this. Would it apply to all code blocks (including procedures), and then the procedures just have a title?

Loquacity:
On the tables: I really like the use of colour in the header bar. One thing, though, it looks like the vertical line is missing between entries in the header? Is that intentional?

jcayouette:
Looks great Namen! I am not sure how I feel about the missing separator in the table header. It looks clean without, but I wonder if there are some weird cases where it could be confusing. 

Also just to record it here (as per our conversation on gitter) It would be great to add some color themed list numbers to our procedure blocks. I think these will help make the steps pop a bit for a reader.

itemone
itemtwo

Namanbhatia7:
@Loquacity @jcayouette I left the separator in the header as it would look much more cleaner. In many  websites, they leave separator either in table contents or table header. I prefer using separator in table content. This way it looks nice and clean.

Namanbhatia7:
> > pre
> > This is pre block
> 
> I like this. Would it apply to all code blocks (including procedures), and then the procedures just have a title?

Yes, this would apply to all pre blocks. But procedures design is different from pre block. I have sent you a Screenshot on gitter. Once finalized, I will make a PR and upload the SS on github also.

PR #345, opened 2020-06-25
Add required URLs

Loquacity:
# Description

Peter Lees noted that we don't list the required URLs for SUMA for configuring firewalls. Added this info to the ports chapter in the Install Guide.

# Target branches

- [ ] Manager-3.2
- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni

# Links

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -6,4 +6,5 @@
 
 # Last packaged: 2020-07-02
 
-- Add GPG information about Oracle clients to SUMA (bz#1173520)
+- Add GPG information about Oracle clients to SUMA (bz#1173520
+- Add required URLs to Installation Guide


pwl0lwp:
Not sure if this is applicable to Uyuni upstream, since it's potentially specific to SUSE subscribers ...?

Loquacity:
> Not sure if this is applicable to Uyuni upstream, since it's potentially specific to SUSE subscribers ...?

Hrm, that's an interesting question. I would need to check with the devs on what URLs Uyuni needs. Perhaps @mcalmer can answer?

mcalmer:
I just wrote a mail to suse-manager ML and listed more hostnames and the conditions when they are needed.
In principle we use use for 3rd party OSes the same URLs as in spacewalk-common-channels . They would belong to uyuni as well.

SCC and updates.suse.com is only used when you register against SCC , which is not required for uyuni. But it is possible.

PR #349, opened 2020-06-29
remove older ubuntu common-channels procedure for suma

Loquacity:
# Description

Removes the spacewalk-common-channels procedure from the Ubuntu-old chapter in the Client Cfg Guide for SUMA (the Uyuni one remains). Also updates the add_channel_cli snippet to make the channel names more generic.

# Target branches

- [ ] Manager-3.2
- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/11819

### DIFF
--- a/modules/client-configuration/pages/snippets/add_channels_cli.adoc
+++ b/modules/client-configuration/pages/snippets/add_channels_cli.adoc
@@ -4,5 +4,7 @@
 ----
 spacewalk-common-channels \
 <base_channel_name> \
-<child_channel_name>
+<child_channel_name_1> \
+<child_channel_name_2> \
+... <child_channel_name_n>
 ----


modules/client-configuration/pages/snippets/add_channels_cli.adoc
```diff
@@ -4,6 +4,5 @@
 ----
 spacewalk-common-channels \
 <base_channel_name> \
-<client_channel_name> \
-<updates_channel_name> \
+<child_channel_name> \
```
keichwa, line 7:
```suggestion
<child_channel_name>
```
Do not escape the final newline.

calancha:
I find useful adding a note like this before the snippet:

```
[NOTE]
====
The mandatory channels do not contain Ubuntu upstream packages. The repositories and channels for synchronizing upstream content can
be must be configured either manually or using [command]``spacewalk-common-channels``.
====
```

A very similar Note appeared in previous versions of the doc (https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/clients-ubuntu-old.html); I guess it might be deleted because the original note says you only can add the upstream manually; that's why I have rephrased it to mention that you can do it manually or via script.

I also find more clear how it was presented the snippet in the link above: it gives a concrete example of the channels you must add.  It would be nice if you go here from the abstract to the concrete in the following way:
1. First, show the snippet as you propose here
2. And right after you can say;
for instance, in the case of Ubuntu 18.04 you would call:
```spacewalk-common-channels \
ubuntu-1804-amd64-main \
ubuntu-1804-amd64-main-updates \
ubuntu-1804-amd64-main-security
```

modules/client-configuration/pages/snippets/add_channels_cli.adoc
```diff
@@ -4,6 +4,5 @@
 ----
 spacewalk-common-channels \
 <base_channel_name> \
-<client_channel_name> \
-<updates_channel_name> \
-----
\ No newline at end of file
+<child_channel_name>
```
juliogonzalez, line 7: What does this mean?

juliogonzalez: Please share a render of this, I don't really understand how this change is going to work.

Loquacity:
> I find useful adding a note like this before the snippet:
> 
> ```
> [NOTE]
> ====
> The mandatory channels do not contain Ubuntu upstream packages. The repositories and channels for synchronizing upstream content can
> be must be configured either manually or using [command]``spacewalk-common-channels``.
> ====
> ```
> 
> A very similar Note appeared in previous versions of the doc (https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/clients-ubuntu-old.html); I guess it might be deleted because the original note says you only can add the upstream manually; that's why I have rephrased it to mention that you can do it manually or via script.

I'm not sure I fully understand the message here, though. As the content stands, they add the channels following either the SUMA (wizard) or Uyuni (spacewalk-common-channels) instructions depending on which version they're looking at, do a sync, and they have packages. If we add a note saying that adding the channels won't give them packages, isn't that just confusing?

> 
> I also find more clear how it was presented the snippet in the link above: it gives a concrete example of the channels you must add. It would be nice if you go here from the abstract to the concrete in the following way:
> 
> 1. First, show the snippet as you propose here
> 2. And right after you can say;
>    for instance, in the case of Ubuntu 18.04 you would call:
> 
> ```
> ubuntu-1804-amd64-main \
> ubuntu-1804-amd64-main-updates \
> ubuntu-1804-amd64-main-security
> ```
> 

That's the point of the table above the snippet. It gives the exact channel names that are required. So you follow the generic instructions, and sub in the details from the table for your specific version. One of the main problems we had in the previous layout was that the instructions were slightly different on every page, and we mentioned channel names in more than one place. This restructure was designed to limit the channel names to one easily updateable location, and have all the procedures generic and maintained in a single source. 

> Including all required channels in the concrete example. This connects with the important remark that we make below
> "You need all the new channels fully synchronized, including Universe (Universe contains important dependencies for Salt), before bootstrapping any Ubuntu client."
> The nice thing about the concrete example is that a customer can copy and paste and it will include all necessary stuff (you won't miss any important child this way); it's also trivial to modify such a 18.04 snippet touse it for the 16.04 case.

I agree that would be nice to have it copyable, but we don't have that level of logic available to us in the docs right now.

modules/client-configuration/pages/snippets/add_channels_cli.adoc
Loquacity: > What does this mean?

I just realised that specifying "tools channel" or whatever didn't always apply, so i genericised it to "child channels". That way whatever names we use for the child channels in the table (tools, universe, security, etc etc) will apply correctly.

Loquacity:
[suse_manager_client-configuration_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/4855556/suse_manager_client-configuration_guide.pdf)

juliogonzalez:
Wait, I got confused by the screenshot. Maybe things are fine.

modules/client-configuration/pages/snippets/add_channels_cli.adoc
juliogonzalez: `child_channel_name` seems to show you only need to add one channel, when in fact you need to add some.

Also, given the table above, that doesn't mention anything about "child channels" I wonder if this is clear enough.

As a minimum I'd change `<child_channel_name>` to `<child_channel_name1> <child_channel_name2> ... <child_channel_nameN>` or similar

Loquacity: That's a fair comment. We do use the parent/child wording in concepts about channels, though, so I think (hope?) that wouldn't be too confusing. I like the idea of adding a couple, though, I'll do that now.

PR #350, opened 2020-06-29
add no-flush option

Loquacity:
# Description

Add no-flush option to bootstrap chapter in client cfg. Also, it turns out we didn't have an entity for s390x.

# Target branches

- [ ] Manager-3.2
- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/11788

### DIFF
--- a/modules/client-configuration/pages/bootstrap-repository.adoc
+++ b/modules/client-configuration/pages/bootstrap-repository.adoc
@@ -136,7 +136,7 @@ mgr-create-bootstrap-repo -c SLE-15-x86_64 --with-parent-channel sle-product-sle
 
 
 If you are creating bootstrap repositories that include multiple different architectures, you need to be careful that all architectures are updated correctly.
-For example, the {x86_64} and {s390x} architectures for {slea} use the same bootstrap repository URL at ``/srv/www/htdocs/pub/repositories/sle/15/1/bootstrap/``.
+For example, the {x86_64} and {ibmz} architectures for {slea} use the same bootstrap repository URL at ``/srv/www/htdocs/pub/repositories/sle/15/1/bootstrap/``.
 
 Because the ``flush`` option is enabled by default, when you attempt to generate the bootstrap repository for multiple architectures, only one architecture would be generated.
 To avoid this, use the ``--no-flush`` option at the command prompt when creating additional architectures.


keichwa:
The SLE writers currently have these entity defs:
```
<!ENTITY x86           "x86">
<!ENTITY amd64         "AMD64">
<!ENTITY zseries       "IBM&nbsp;Z">
<!ENTITY ipseries      "POWER">
<!ENTITY ppc           "POWER">
<!ENTITY power         "POWER">
<!ENTITY intel64       "Intel&nbsp;64">
<!ENTITY x86-64        "&amd64;/&intel64;">
<!ENTITY arm           "Arm">
<!ENTITY armreg        "Arm*">
<!ENTITY armv7         "Armv7">
<!ENTITY armv8         "Armv8">
<!ENTITY aarch64       "AArch64">
<!ENTITY arm64         "&arm;&nbsp;&aarch64;">
```
The s390 architecture is now called "IBM Z".  Of course, in many strings and labels we still have something with s390.

Loquacity:
> The SLE writers currently have these entity defs:
> 

We should probably have all of these in, even though we don't use all of them.

> 
> The s390 architecture is now called "IBM Z". Of course, in many strings and labels we still have something with s390.

True, but I think in repos and stuff it's still called s390x.

keichwa:
yes, but in these phrases I'd expect to read "IBM Z" (and not s390x):

> For example, the {x86_64} and {s390x} architectures for {slea} use the same bootstrap repository URL at ``/srv/www/htdocs/pub/repositories/sle/15/1/bootstrap/``.

Might be a corner case.

Loquacity:
> yes, but in these phrases I'd expect to read "IBM Z" (and not s390x):
> 
> > For example, the {x86_64} and {s390x} architectures for {slea} use the same bootstrap repository URL at `/srv/www/htdocs/pub/repositories/sle/15/1/bootstrap/`.
> 
> Might be a corner case.

Ah, yes, that's a good point.

Loquacity:
> IIUC the new message about --flush will should for both Uyuni and SUSE Manager. 

Yes, it will appear in both.

> Only question is: shouldn't we have warnings on the client pages for each OS? I am not so sure customers/users are going to check the bootstrap page

I think we have two options here, and it depends on how likely it is that readers will come across the problem:

* If readers are going to hit the problem *every time* they register a client then we can add an admonition to check the bootstrap section.
* If readers are going to hit the problem *occasionally* then we can add a note to the troubleshooting section.

Which do you think is most likely? 

> 
> Also **make sure** this is part of GMC! It's already causing confusion on QA.

Yes, it will get there.

juliogonzalez:
> Which do you think is most likely?

I'd put my money on the first case... because I am sure it's what would happen to me :-)

@mcalmer do you have an opinion?

Loquacity:
OK, let's merge this as-is, then I'll come back and add the extra in once we've decided.

mcalmer:
Everything should be automated and no user has a reason to mess around with it.
That QA must know it as they have to use this switch every time is special to them and not the normal customer usecase.

juliogonzalez:
> Everything should be automated and no user has a reason to mess around with it.
Maybe for SUSE Manager, but don't we still need manual creation for Uyuni because of the parent custom channels from spacewalk-common-channels?

PR #351, opened 2020-06-29
add extra content to ssl certs intro

Loquacity:
# Description

Adds @mcalmer 's exposition about SSL certs to the SSL certs intro (lightly edited). Still to do: merge this section and the section above. They're pretty similar already, but the new content is more specific to SUMA, which is a good thing.

# Target branches

- [ ] Manager-3.2
- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni

# Links

Fixes:
https://github.com/SUSE/spacewalk/issues/11728
https://bugzilla.suse.com/show_bug.cgi?id=1171632#c24

### DIFF
--- a/modules/administration/pages/ssl-certs.adoc
+++ b/modules/administration/pages/ssl-certs.adoc
@@ -3,42 +3,23 @@
 
 {productname} uses SSL certificates to ensure that clients are registered to the correct server.
 
-When you create an SSL certificate for your server, the certificate includes the host name of the server it is associated with.
-The certificate is then deployed on the server, and when clients attempt to connect, they check the certificate details.
-If the certificate contains the name of the server they are attempting to connect to, the connection will be successful.
-The client will also check if the certificate has been signed by a trusted certificate authority.
-
-All certificates must be signed by a certificate authority (CA), in order for them to be considered valid, and for clients to be able to successfully match against them.
-When an organization signs its own certificate, the certificate is considered self-signed.
-A self-signed certificate is straight-forward to set up, and does not cost any money, but they are considered less secure.
-
-For additional security, you can arrange a third party CA to sign your certificates.
-Third party CAs perform checks to ensure that the information contained in the certificate is correct.
-They will usually charge an annual fee for this service.
-Using a third party CA makes certificates harder to spoof, and will provide additional protection for your installation.
-
-By default, {productname} uses a self-signed certificate.
-If you have certificates signed by a third party CA, you can import them to your {productname} installation.
-
-
-
-== SSL Certificates in {productname}
-// I'm adding this as a separate section for now, to make this easier, but I think we can probably merge these two sections together. --LKB 2020-06-29
-
-Every {productname} Server that uses SSL requires an SSL server certificate.
-Provide the path to the server certificate either in the environment variable ``SERVER_CERT`` during setup or with ``--from-server-cert`` when using rhn-ssl-tool directly.
-
-Every client that uses SSL to register to the {productname} Server checks that it is connecting to the right server by validating against the server certificate.
+Every client that uses SSL to register to the {productname} Server checks that it is connecting to the right server by validating against a server certificate.
 This process is called an SSL handshake.
 
-During the SSL handshake, the client will check that the hostname in the server certificate matches.
+During the SSL handshake, the client will check that the hostname in the server certificate matches what it expects.
 The client also needs to check if the server certificate is trusted.
-To do this, it looks for the issuer of the server certificate.
+
+Every {productname} Server that uses SSL requires an SSL server certificate.
+Provide the path to the server certificate using the ``SERVER_CERT`` environment variable during setup, or with the ``--from-server-cert`` option of the [command]``rhn-ssl-tool`` command.
 
 Certificate authorities (CAs) are certificates that are used to sign other certificates.
-The root CA is a special kind of CA that is signed with itself.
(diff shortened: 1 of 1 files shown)


modules/administration/pages/ssl-certs.adoc
```diff
@@ -20,5 +20,52 @@ Using a third party CA makes certificates harder to spoof, and will provide addi
...
+Your root CA certificate is stored using the ``CA_CERT`` environment variable.
+
+In order for SSL authentication to work correctly, the client must trust the root CA.
+This means that the root CA must be installed on every client.
+
+The default method of SSL authentication is for {productname} to use self-signed certificates.
+In this case, {productname} has generated all the certificates, and the root CA has signed the server certificate directly.
+
+An alternative method is to use an intermediate CA.
+In this case, the intermediate CA signs the root CA.
```
mcalmer, line 50: I think the oposite. The Root CA signs the Intermediate CA and the Intermediate CA signs other iCA and finally the last one signs the server certificate.

modules/administration/pages/ssl-certs.adoc
```diff
@@ -20,5 +20,52 @@ Using a third party CA makes certificates harder to spoof, and will provide addi
...
+The intermediate CA can then sign any number of other intermediate CAs, one of which can sign the server certificate.
+This is referred to as a chained certificate.
+
+If you are using intermediate CAs in a chained certificate, the root CA is installed on the client, and during the SSL handshake, the server presents the server certificate.
```
mcalmer, line 54: While this is correct and a prefered way, SUSE Manager does it different.
As the setup tools are not written for iCAs we do a trick and install all CAs as trusted on the clients.
In this case the server does not present the chain to the client.

mcalmer: Oh, i just see, that this is written 2 sections below.
So we should this rephrase as a typical way to configure this.

modules/administration/pages/ssl-certs.adoc
```diff
@@ -20,5 +20,52 @@ Using a third party CA makes certificates harder to spoof, and will provide addi
 By default, {productname} uses a self-signed certificate.
 If you have certificates signed by a third party CA, you can import them to your {productname} installation.
 
+
+
+== SSL Certificates in {productname}
+// I'm adding this as a separate section for now, to make this easier, but I think we can probably merge these two sections together. --LKB 2020-06-29
+
+Every {productname} Server that uses SSL requires an SSL server certificate.
+Your server certificate is stored using the ``SERVER_CERT`` environment variable.
```
mcalmer, line 29: We need to take care that the context of this chapter is correct.
Otherwise nobody understand about what environment variables we talk here.

AFAIK this is only used for the setup at new installations. For the process of replacing certs, we have commandline options which are named similar but not exactly the same.

modules/administration/pages/ssl-certs.adoc
Loquacity: The way I read this is that the client has the root CA, the server has the server cert, and we haven't yet defined where the intermediate certs are. Then we discuss that they can be either on the client or on the server.

modules/administration/pages/ssl-certs.adoc
Loquacity: Ah yeah, my bad.

modules/administration/pages/ssl-certs.adoc
Loquacity: Heh, of course :roll_eyes:

modules/administration/pages/ssl-certs.adoc
Loquacity: reworded slightly. let me know if that's better or not.

modules/administration/pages/ssl-certs.adoc
mcalmer: Maybe we should name both. First would be to express that the variable contain a path to the file.

```suggestion
Provide the path to the server certificate either in the environment variable ``SERVER_CERT`` during setup or with ``--from-server-cert`` when using rhn-ssl-tool directly.
```

Similar for `CA_CERT` and `--from-ca-cert` .

modules/administration/pages/ssl-certs.adoc
```diff
@@ -20,5 +20,51 @@ Using a third party CA makes certificates harder to spoof, and will provide addi
...
+In this case, the root CA signs the intermediate CA.
+The intermediate CA can then sign any number of other intermediate CAs, and the final one signs the server certificate.
+This is referred to as a chained certificate.
+
+If you are using intermediate CAs in a chained certificate, the root CA is installed on the client, and the server certificate is installed on the server.
+During the SSL handshake, clients must be able to verify the entire chain of intermediate certificates between the root CA and the server certificate, so they must be able to access all the intermediate certificates.
+
+There are two main ways of achieving this.
+In {productname}, by default, all the intermediate CAs are installed on the client.
+However, you could also install them on the server.
```
mcalmer, line 59:
```suggestion
However, you could also configure your services on the server to provide them to the client.
```

Loquacity:
Thanks for the edits. I'm going to go ahead and merge these two sections into something more coherent now.

Loquacity:
[uyuni_administration_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/4855660/uyuni_administration_guide.pdf)

image

PR #354, opened 2020-06-30
Fix a coupple of broken links

calancha:
# Description

Fix a couple of broken links

- modules/client-configuration/pages/clients-ubuntu-old.adoc
- modules/client-configuration/pages/clients-ubuntu.adoc

# Targets

- [x] Manager-4.1
- [x] Uyuni


calancha:
@Loquacity Since this is orthogonal to the issues we have discussed last days, then I prefer to handle it separately here.

PR #358, opened 2020-07-02
Since instead of From?

vmoutoussamy:
Since instead of FRom?

# Description

# Target branches

- [ ] Manager-3.2
- [ ] Manager-4.0
- [ ] Manager-4.1
- [ ] Uyuni

# Links

### DIFF
--- a/modules/installation/pages/install-intro.adoc
+++ b/modules/installation/pages/install-intro.adoc
@@ -11,7 +11,7 @@ This book provides guidance on installing {productname}.
 
 ifeval::[{suma-content} == true]
 
-In SLES{nbsp}15{nbsp}SP1 and later, {susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
+{susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
 This is the default method of installation.
 
 endif::[]


modules/installation/pages/install-intro.adoc
```diff
@@ -11,7 +11,7 @@ This book provides guidance on installing {productname}.
 
 ifeval::[{suma-content} == true]
 
-From SLES{nbsp}15{nbsp}SP1, {susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
+Since SLES{nbsp}15{nbsp}SP1, {susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
```
Loquacity, line 14: How about this?

```suggestion
In SLES{nbsp}15{nbsp}SP1 and later, {susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
```

vmoutoussamy:
Yes, I prefer your suggestion : ) however this made me think... is it possible to install SUMA4.1 with 15SP1? I don't think so right? 
So maybe the "15SP1" part could cause confusion? I think this is a general information statement but can't people misunderstand it and think that 4.1 can be installed with "15SP1 and later"? 

Sorry for the trouble, but I thought it is worth to mention it.

Loquacity:
> Yes, I prefer your suggestion : ) however this made me think... is it possible to install SUMA4.1 with 15SP1? I don't think so right?
> So maybe the "15SP1" part could cause confusion? I think this is a general information statement but can't people misunderstand it and think that 4.1 can be installed with "15SP1 and later"?
> 
> Sorry for the trouble, but I thought it is worth to mention it.

Hrm, I think they can, actually. But we can check.

Loquacity:
> > Yes, I prefer your suggestion : ) however this made me think... is it possible to install SUMA4.1 with 15SP1? I don't think so right?
> > So maybe the "15SP1" part could cause confusion? I think this is a general information statement but can't people misunderstand it and think that 4.1 can be installed with "15SP1 and later"?
> > Sorry for the trouble, but I thought it is worth to mention it.
> 
> Hrm, I think they can, actually. But we can check. Or we can just remove the conditional bit an say "SUSE Manager Server and Proxy are installed using the unified installer" (or similar).

modules/installation/pages/install-intro.adoc
```diff
@@ -11,7 +11,7 @@ This book provides guidance on installing {productname}.
 
 ifeval::[{suma-content} == true]
 
-From SLES{nbsp}15{nbsp}SP1, {susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
+In SLES{nbsp}15{nbsp}SP1 and later, {susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
```
Loquacity, line 14:
```suggestion
{susemgr} Server and Proxy are available as base products, and can be installed with the SLES Unified Installer.
```

Loquacity:
@vmoutoussamy If you're happy with this now, let's merge it.

PR #373, opened 2020-07-11
Provisioning for salt clients is supported

mcalmer:
# Description

Provisioning for salt clients is supported since a while now.
We can remove this Note

# Target branches

- [ ] Manager-3.2
- [x] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni


Loquacity:
Has this already been removed in 4.1?

mcalmer:
No, but I checked 4.1 docs and it seems this section is different there and has not this problem.
Maybe you can check, but I think we do not need to port it.

Loquacity:
> No, but I checked 4.1 docs and it seems this section is different there and has not this problem.
> Maybe you can check, but I think we do not need to port it.

Yeah, ISTR changing it not so long ago. Glad you agree!

PR #383, opened 2020-07-20
Maintenance Windows

Loquacity:
# Description

- Adds new Maintenance Windows chapter to Admin Guide
- Renames old MW chapter to maintenance-window-tasks, and makes it a sub-chapter of the new MW chapter
- Some reorg of the Admin Guide to get like with like
- Adds new MW chapter to Reference Guide

# Target branches

- [ ] Manager-3.2
- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/11254

### DIFF
--- a/modules/administration/pages/maintenance-windows.adoc
+++ b/modules/administration/pages/maintenance-windows.adoc
@@ -138,5 +138,12 @@ The restricted actions are:
 * Service pack migrations
 * Cluster operations
 
+[NOTE]
+====
+For Salt clients, it is possible to run remote commands directly at any time by navigating to menu:Salt[Remote Commands].
+This applies whether or not the Salt client is in a maintenance window.
+For more information about remote commands, see xref:administration:actions.adoc[].
+====
+
 Unrestricted actions are minor actions that are considered safe and are unlikely to cause problems on the client.
 If an action is not restricted it is, by definition, unrestricted, and can be be run at any time.


Loquacity:
image
[suse_manager_administration_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/4945540/suse_manager_administration_guide.pdf)

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,25 @@
+[[maintenance-windows]]
+= Maintenance Windows
+
+The maintenance windows feature in {productname} allows you to schedule actions to occur during your maintenance window period ahead of time, and to create repeating tasks.
+
```
hustodemon, line 5: I think that it'd be good, if we clarified, what the terms *Calendar* and *Schedule* are. These are some facts that should describe them:

- *Calendar* is basically an [iCalendar](https://en.wikipedia.org/wiki/ICalendar) file imported to SUMA
- *Schedule* uses events defined in *Calendar* to define/generate maintenance windows
- *Schedule* can be assigned to *System*s
- A *System* with assigned *Schedule* can't run "restricted" action outside of maintenance windows
- Restricted actions are actions, that modify the system significantly (and could potentially break it) - e.g. package installation, upgrade, service pack migration, highstate application. Unrestricted actions are fairly safe actions - e.g. package profile update, hardware refresh, subscribing to software channels.

diagram-schedule-1

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+.Procedure: Creating a New Schedule
+. Step the first
+. ...
+. Final step
+
+
+
+.Procedure: Performing a Maintenance Update
```
hustodemon, line 22: As a 3rd step, I'd say we should tell the user to assign the *Schedule* to a *System*(or *System*s). When the user does this, it means that they can't schedule the "restricted" actions anytime, but only inside of the maintenance windows defined by the *Schedule* (and *Calendar*).

As a 4th step, we could tell them to perform a package upgrade (any other "restricted" action would do, but I think a package upgrade is a good example) inside of maintenance window. The user just go to the UI for package upgrade (`System details -> Software -> Packages -> Upgrade`), selects the package(s) to upgrade and then selects the maintenance window from the UI.
(I'd also avoid the term *maintenance update*, since this is something that users have associated with the SUMA releases. The more precise caption would be "Performing a Package Upgrade inside a Maintenance Window).

modules/reference/pages/schedule/maintenance-windows.adoc
```diff
@@ -0,0 +1,14 @@
+[[schedule-maint-windows]]
+= Maintenance Windows
+
+The menu:Schedule[Maintenance Windows] section allows you to manage your maintenance window schedules.
+To create a maintenance window, you need to have an existing [path]``.ical`` file that contains the calendar.
```
hustodemon, line 5:
```suggestion
To create a maintenance calendar, you need to have an existing [path]``.ical`` file that contains the calendar.
```

hustodemon: There are 2 methods of creating a *Calendar*: one is (as you have described) by uploading the iCalendar file. The other is by providing a link (`http`, `https`) and SUMA will download the iCalendar data from it.

modules/reference/pages/schedule/maintenance-windows.adoc
```diff
@@ -0,0 +1,14 @@
...
+Create a new schedule, and provide a link to the calendar file.
```
hustodemon, line 6:
```suggestion
Create a new schedule, and select the calendar.
```
When creating a *Schedule*, user just selects the *Calendar* that they created above, they don't provide links anymore.

modules/administration/pages/maintenance-window-tasks.adoc
```diff
@@ -1,5 +1,5 @@
-[[maintenance-window]]
-= Maintenance Window
+[[maintenance-window-tasks]]
```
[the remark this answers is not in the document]
Loquacity: heh, yes, it was a request that came out of the interviews we did when we were planning the IA restructure.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
...
+Unrestricted actions include:
+
+* Package profile update
+* Hardware refresh
+* Subscribing to software channels
+
+
+Before you begin, you must create an [path]``ical`` file for upload, or link to an [path]``ical`` file to create the calendar.
+You can create [path]``ical`` files in your preferred calendaring tool, such as KOrganizer.
+// I don't know if this is correct. ---LKB 2020-07-21
```
hustodemon, line 34: That sounds good.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+
+
+.Procedure: Creating a New Maintenance Calendar
+. In the {productname} {webui}, navigate to menu:Schedule[Maintenance Windows > Calendar], and click btn:[Create].
+. In the [guimenu]``Calendar Name`` section, type a name for your calendar.
+. Either provide a URL to your [path]``ical`` file, or upload the file directly.
+. Click btn:[Create Schedule] to save your calendar.
```
hustodemon, line 42: Hm, I think the button says `Create Calendar` (could be that an incorrect version was deployed on the test server).

ss-cal

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
...
+// Well, that;s confusing :\ -- LKB 2020-07-21
+
+
+
+.Procedure: Creating a New Schedule
+. In the {productname} {webui}, navigate to menu:Schedule[Maintenance Windows > Calendar], and click btn:[Create].
+. In the [guimenu]``Schedule Name`` section, type a name for your schedule.
+. OPTIONAL: If your [path]``ical`` file contains more than one calendar, and you want to schedule all of the calendars, check [guimenu]``Multi``.
```
hustodemon, line 50: I'm not sure if "if your file contains more than one calendar" is fitting here, ical file contains events, not calendars. I'd suggest to rephrase in something like: "single" schedule uses all events defined in the calendar to define maintenance windows, "multi" schedule uses only  events, that have their *summary* equal to the schedule name (see the example picture above).

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
...
+. Select the calendar to assign to this schedule.
+. Click btn:[Update Schedule] to save your schedule.
+
+
+
+.Procedure: Assigning a Schedule to a Client
+. In the {productname} {webui}, navigate to menu:Systems[Systems List], select the client to be assigned to a schedule, locate the [guimenu]``Properties`` panel and click btn:[Edit These Properties].
+Alternatively, you can assign clients through the system set manager.
```
hustodemon, line 58: Maybe I'd elaborate on the place of the feature in the SSM (it's a bit hidden there): it's in `<go to ssm> -> Misc -> Maintenance Windows`.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
...
+. In the [guimenu]``Edit System Details`` page, locate the [guimenu]``Maintenance Schedule`` field, and select the name of the schedule to be assigned.
+. Click btn:[Update Properties] to assign the maintenance schedule.
+. OPTIONAL: To remove a maintenance schedule, check the btn:[Cancel affected actions] option.
```
hustodemon, line 61: The "cancel affected actions" function can be used in situations, when you have some (restricted) actions for a system that are scheduled for the future and you want to assign a schedule to that system. 

Let's consider this example:
- you have a system that is not assigned to any maint. schedules
- you schedule a highstate application to happen in the future, e.g. `2020-12-12 10:00 (Saturday)`
- now you want to assign a maintenance schedule to this system, but the maintenance schedule only contains windows on weekdays
- obviously, the schedule is not compatible with the existing scheduled action (highstate application), so if you check "cancel affected actions", SUMA will cancel the scheduled highstate execution of that highstate upon assigning the schedule
- (otoh, if you leave that unchecked and you assign that schedule, SUMA will report a warning, that assigning that schedule is not allowed)

Therefore I'd rephrase into something like this:
```suggestion
. OPTIONAL: Checking the btn:[Cancel affected actions] cancels already scheduled actions that collide with the maintenance schedule.
```

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
+[[maintenance-windows]]
+= Maintenance Windows
+
+The maintenance windows feature in {productname} allows you to schedule actions to occur during a scheduled maintenance window period.
+Maintenance windows can be scheduled to suit your calendar.
+When you have created a schedule, you can select tasks that can only be performed during a scheduled maintenance window.
```
hustodemon, line 6: I'm not sure I completely understand this sentence, but I think it describes the fact, that the feature restricts action execution to the maintenance windows (and does not allow executing some types of actions outside of them). I think it's important to describe this "restrictiveness" of the maintenance windows, but I find it a bit confusing.

Could we maybe merge this to the first sentence of the paragraph and rephrase to roughly something like this:

The maintenance windows feature in {productname} allows you to schedule actions to occur during a scheduled maintenance window period, and prevents you from executing actions outside of these periods.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
...
+
+Maintenance windows require both a calendar, and a schedule.
+The calendar defines the date and time of your maintenance window events, including recurring events, and must be in [path]``ical`` format.
+The schedule uses the events defined in the calendar to create the maintenance windows.
+You must create an [path]``ical`` file for upload, or link to an [path]``ical`` file to create the calendar, before you can create the schedule.
```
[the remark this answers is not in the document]
Loquacity: added a new section to handle multi.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,61 @@
...
+* Client upgrade
+* Service pack migration
+* Highstate application (for Salt clients)
+
+Unrestricted actions are minor actions that are considered safe and are unlikely to cause problems on the client.
+Unrestricted actions include:
+
+* Package profile update
+* Hardware refresh
+* Subscribing to software channels
```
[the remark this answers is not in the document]
Loquacity: Let's leave it out for now. If we get a customer request we can add it in later on.

modules/administration/pages/maintenance-windows.adoc
Loquacity: sometimes my guesses turn out ok! :tada:

modules/administration/pages/maintenance-windows.adoc
Loquacity: That came from a video, I haven't tested much yet. Will come back to this.

modules/administration/pages/maintenance-windows.adoc
Loquacity: Yep, added a new section to handle this. So in this version I'll just have a one-line 'optional' and then a proper explanation and procedure later on.

modules/administration/pages/maintenance-windows.adoc
Loquacity: Ah, that makes sense now! I've added this into an admonition below the procedure.

modules/administration/pages/maintenance-windows.adoc
Loquacity: FYI: the test server is showing 404 as of 2020-07-22 17:00 UTC. Will have to come back to testing later.

Loquacity:
Note that this isn't fully tested yet, but I think we're OK to open it up to a few more eyeballs :eyes:

Loquacity:
> Thanks for the commits @Loquacity, I think we're almost done. I just left a bunch of comments.
> 
> Moreover, there is one thing that should be explicitly documented (according to the tracking issue): explaining the difference between maintenance windows and system locking (another feature). I think it should come either at the beginning of the section, or just before the "procedures". I wanted to give you some draft of this explanation today, but I'm affraid, I won't make it. I'll definitely send you some stuff tomorrow!

OK, I've just added an admonition in with an xref for now. Let me know what you want in :)

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,103 @@
+[[maintenance-windows]]
+= Maintenance Windows
+
+The maintenance windows feature in {productname} allows you to schedule actions to occur during a scheduled maintenance window period.
+When you have assigned a maintenance window to a client, the client is prevented from executing some actions outside of the specified period.
```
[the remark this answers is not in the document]
Loquacity: OK, that makes a lot of sense. I've reworded again, LMK.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,103 @@
...
+Each event contains a ``summary`` field.
+If you want to create multiple maintenance windows for one calendar, you can specify events for each using the ``summary`` field.
+
+For example, you might like to create a schedule for production servers, and a different schedule for testing servers.
+In this case, you would specify ``SUMMARY: Production Servers`` on events for the production servers, and ``SUMMARY: Testing Servers`` on events for the testing servers.
+
+image::maint_windows_multi.png[scaledwidth=80%]
+
+There are two types of schedule: single, or multi.
+If your calendar contains events that apply to more than one schedule, then you must select ``multi``, and ensure you name the schedule according to the ``summary`` field you used in the calendar file.
```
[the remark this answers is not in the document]
Loquacity: Yeah, so this was my thinking for those three paras:

1: Defines calendar events, introduces the `summary` field.
2: Example of use of the `summary` field
IMAGE
3: Discusses using this in practice

I think if we move para 3 up to be para 1 or 2, it tells you how to do something before you have a clear idea of what the thing is.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+
+
+.Procedure: Creating Multiple Schedules
```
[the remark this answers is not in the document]
Loquacity: Yeah, I was kind of hoping I could sub in "multiple" for "multi" because I have a pathological aversion to abbreviations :joy:

modules/administration/pages/maintenance-windows.adoc
Loquacity: Still 404 at 2020-07-23 0600 UTC

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,111 @@
...
+* Client upgrade
+* Service pack migration
+* Highstate application (for Salt clients)
+
+Unrestricted actions are minor actions that are considered safe and are unlikely to cause problems on the client.
+Some examples of unrestricted actions are:
+
+* Package profile update
+* Hardware refresh
+* Subscribing to software channels
```
mcalmer, line 36: Should we add a complete list of action which are restricted and which not?
AFAIK Pau wanted to have it.

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+
+Before you begin, you must create an [path]``ical`` file for upload, or link to an [path]``ical`` file to create the calendar.
+You can create [path]``ical`` files in your preferred calendaring tool, such as KOrganizer.
```
mcalmer, line 40: Korganizer is not the best example here. Should we better name outlook/office365 and google calendar?

modules/administration/pages/maintenance-windows.adoc
Loquacity: So @hustodemon and I thought that because the list is quite long, we would need to add it at the end of the section or something, rather than here. Happy to do that, if @paususe wants it.

modules/administration/pages/maintenance-windows.adoc
Loquacity: I was going for open source, but sure :)

mcalmer: Yes, but customers using not open source clients :-(
So we designed this to work with outlook calendars and also tried google as most people will have it.
I would still name KOrganizer, but also the other.

paususe: Agree with MC's comment. As much as I like KOrganizer (former KDE developer here!), it's not that popular. I would rewrite naming the most popular e-mail clients:
"... such as Microsoft Outlook, Google Calendar or Mozilla Thunderbird"

modules/administration/pages/maintenance-windows.adoc
hustodemon: I can give you the list. @paususe do you want to have the full list?

modules/administration/pages/maintenance-windows.adoc
hustodemon: Ouch! Working on it now.

hustodemon: (anyway, the correct button is `Create Calendar`, so we should put this one to the docs. if you discover, it says `Create Schedule` in some future version of uyuni, then it's a bug and (i will eat my hat))

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,118 @@
...
+. In the {productname} {webui}, navigate to menu:Schedule[Maintenance Windows > Schedules], and click btn:[Create].
+. In the [guimenu]``Schedule Name`` section, type a name for your schedule.
+. OPTIONAL: If your [path]``ical`` file contains events that apply to more than one schedule, check [guimenu]``Multi``.
+. Select the calendar to assign to this schedule.
+. Click btn:[Create Schedule] to save your schedule.
+
+
+
+.Procedure: Assigning a Schedule to a Client
+. In the {productname} {webui}, navigate to menu:Systems[Systems List], select the client to be assigned to a schedule, locate the [guimenu]``System Properties`` panel and click btn:[Edit These Properties].
```
keichwa, line 64:
```suggestion
. In the {productname} {webui}, navigate to menu:Systems[Systems List], select the client to be assigned to a schedule, locate the [guimenu]``System Properties`` panel, and click btn:[Edit These Properties].
```

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,118 @@
...
+[NOTE]
+====
+When you assign a new maintenance schedule to a client, it is possible that the client might already have some restricted actions scheduled, and that these might now conflict with the new maintenance schedule.
+If this occurs, the {webui} will display an error and you will not be able to assign the schedule to the client.
+To resolve this, check the btn:[Cancel affected actions] option when you assign the schedule.
+This will cancel any previously scheduled actions that conflict with the new maintenance schedule.
+====
+
+
+When you have created your maintenance windows, you can schedule restricted actions, such as package upgrades, to be performed during the maintenance window.
```
hustodemon, line 78: Just me feeling (feel free to ignore): I'm a bit tentative about the  "When you have created your maintenance windows"  phrase, could we use someting like "Now you can schedule restricted actions, such as package upgrades, to be performed during the maintenance window."?

Loquacity: The only problem that I have with your suggestion is that we're assuming the reader knows what "now" refers to. We know that readers jump around documentation, so we can't assume they have both read the entire preceding content and performed all the preceding steps. By using the "when you ... you can" construction (which is becoming standard in our docs), we're giving some more signposts for the reader. I appreciate that it can sound a little clunky when you've read the whole thing though :wink:

Loquacity:
OK, so we just need to know whether or not @paususe wants the full list of actions. I'm assuming it's still the case that we don't want to merge this until *after* Uyuni 2020.07 goes out, though?

modules/administration/pages/maintenance-windows.adoc
paususe: Yes please, full list.

Loquacity: OK. I'll need that list then, please @hustodemon :slightly_smiling_face:

mcalmer: @Loquacity I made a list of actions which can only be executed when a client is in maintenance mode (in a maintenance window). I tried to make the list understanable by humans and grouped things togther . We will see if it is correct enough.

* package operations (install/update/remove)
* patch update
* reboot
* config management changing tasks
* State Highstate
* autoinstallation (reinstall)
* remote command
* SP Migration
* Cluster operations

Loquacity: lol, I thought the list was long? :joy:

modules/administration/pages/maintenance-windows.adoc
```diff
@@ -0,0 +1,142 @@
...
+The restricted actions are:
+
+* Package operations (for example, installing, updating, or removing packages)
+* Patch updates
+* Rebooting a client
+* Rolling back transactions
+* Configuration management changing tasks
+* Applying a highstate (for Salt clients)
+* Autoinstallation and reinstallation
+* Remote commands
```
mcalmer, line 137: Maybe we should add a special note about remote command and salt.
We restrict only `actions` which can be scheduled. 

But for salt clients we have an additional page where you can run remote commands for it.
Salt => Remote Command  - here you can run a command directly. This works also for clients which are not in maintenance mode.

PR #388, opened 2020-07-21
Uyuni proxy upgrade keichwa

keichwa:
# Description

upgrade uyuni proxy to 2020.07 with OS update

# Target branches

- [ ] Manager-3.2
- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni

# Links

### DIFF
--- a/modules/upgrade/pages/proxy-uyuni.adoc
+++ b/modules/upgrade/pages/proxy-uyuni.adoc
@@ -87,7 +87,10 @@ spacewalk-proxy stop
 . In the [guimenu]``Child Channels`` pane, select the {opensuse-version} child channels.
 . Click btn:[Next], and [guimenu]``Confirm Software Channel Change`` with btn:[Confirm].
 . Click menu:Software[Packages > Upgrade], and select all the packages to be updated on the proxy, and then apply the selection.
-. Re-register with the reactivation key.
+. Re-register with the reactivation key using the [command]``rhnreg_ks`` command-line utility.
+The system will be re-registered with the same id, history, and groups.
++
+// and channels (unless the system's base channel changes).
 . On the {productname} Proxy, start the proxy service:
 +
 ----


modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+// there is no chance to fix any issues during migration,
+// make sure you have a backup before continuing. If you are
+// running Uyuni server on a virtual machine, it is advisable
+// to create a snapshot before performing the migration!
+
+. Replace and refresh software repositories:
++
+----
+mv /etc/zypp/repos.d /etc/zypp/repos.d.old
```
juliogonzalez, line 31: Huh? what's this :-?

paususe: It's creating a backup of the old repos directory but this is wrong. What if the user has some other repositories enabled?

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -0,0 +1,51 @@
...
+mkdir /etc/zypp/repos.d
+zypper ar -n "Main Repository" http://download.opensuse.org/distribution/leap/15.2/repo/oss repo-oss
+zypper ar -n "Main Update Repository" http://download.opensuse.org/update/leap/15.2/oss repo-update
+zypper ar -n "Non-OSS Repository" http://download.opensuse.org/distribution/leap/15.2/repo/non-oss repo-non-oss
+zypper ar -n "Update Repository (Non-Oss)" http://download.opensuse.org/update/leap/15.2/non-oss/ repo-update-non-oss
+zypper ar -n "Uyuni Server Stable" https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable/images/repo/Uyuni-Server-POOL-x86_64-Media1/ uyuni-server-stable
```
paususe, line 37: It should be 

```suggestion
zypper ar -n "Uyuni Proxy Stable" https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable/images/repo/Uyuni-Proxy-POOL-x86_64-Media1/ uyuni-proxy-stable
```

juliogonzalez: Could work, but I don't think this procedure makes sense at all as it is. Aren't the proxies always clients?

modules/upgrade/pages/proxy-uyuni.adoc
juliogonzalez: On a proxy, it should not. This is not what worries me most. The whole script is what worries me. I think it's the wrong way to do the uprade.

juliogonzalez: Funny enough it seems we tell people to install the Uyuni Proxy not as a client: https://www.uyuni-project.org/uyuni-docs/uyuni/installation/install-proxy-uyuni.html

I don't know why this is this way. To me it seems strange, but ok. For now it is how it is. So the script seems to make sense.

juliogonzalez: Ah *now* I understand it. You tell them to it that way, but then you say this: https://www.uyuni-project.org/uyuni-docs/uyuni/installation/uyuni-proxy-registration.html (BTW the channel name for `spacewalk-common-channels` is not `uyuni-proxy-stable-leap-151` bot `uyuni-proxy-stable-leap`)

So in the end this migration script is still wrong, as the proxies will be clients. If you run this script, yeah, the migration will work, but then the Server will still think the proxy is a Leap 15.1, so I guess it will crate a mess.

This what I suspect that needs to be done if you want to migrate:

1. Add the Leap 15.2 channels with `spacewalk-common-channels opensuse_leap15_2  opensuse_leap15_2-non-oss  opensuse_leap15_2-non-oss-updates  opensuse_leap15_2-updates  opensuse_leap15_2-uyuni-client uyuni-proxy-stable-leap-152`
2. Sync them: `spacewalk-repo-sync -p opensuse_leap15_2-x86_64` (this will sync the parent and all children).
3. Adjust the proxies using SSM or one by one to use the Leap 15.2 channels you just added and synced. Remember to apply the changes.
4. Perform the migration (this is the "gray" area for me, maybe this where they need to run `spacewalk-proxy stop; zypper ref; zypper -n dup --allow-vendor-change; shutdown -r now` directly on the proxies).

And that should be all.

BTW I just noticed we don't have `uyuni-server-stable-leap-152`, so time to create it.

juliogonzalez: https://github.com/uyuni-project/uyuni/pull/2432

juliogonzalez: OK, I thought a little about all this mess. Let's do the following:

1. I will tag Uyuni tomorrow, and will no accept more submissions to the master branch (meanwhile I will prepare a MU as well)
2. @keichwa and @mantel can test the migration of the proxy tomorrow and Thursday and have the doc ready by end of Thursday
3. @jcayouette prepares the package and the PR for the website on Thursday by the end of the day, or Friday, early in the morning.
4. I start the release on Friday morning, as soon as the PR and the package for the doc are ready.

Sounds fair enough? That gives some more time to test the migration without that much pressure.

Please ACK tomorrow morning if we have an agreement.

keichwa:
Julio González Gil <notifications@github.com> writes:

> Sounds fair enough? That gives some more time to test the migration without that much pressure.
>
> Please ACK tomorrow morning if we have an agreement.

ACK, I think this is a good proposal.

-- 
Karl Eichwalder

modules/upgrade/pages/proxy-uyuni.adoc
mantel: I answered too quickly yesterday. While this script probably would (sort of) work, the clean way to upgrade the proxy would be to initiate an autoyast upgrade from the server. This way all the information regarding the proxy would stay consistent. We even had this documented some time ago; biggest issue is providing a working autoyast file which is a bit tricky.

mantel: > So in the end this migration script is still wrong, as the proxies will be clients. If you run this script, yeah, the migration will work, but then the Server will still think the proxy is a Leap 15.1, so I guess it will crate a mess.

Maybe this could even be fixed up by refreshing the information from the proxy, but yes, the only really clean way is to perform the upgrade via autoyast so the server is aware of any modifications done to the proxy.

juliogonzalez: I think that providing manual instructions should be fine as well (if it is easier) as long as they are accurate. Something like what I described above.

mantel: > I think that providing manual instructions should be fine as well (if it is easier) as long as they are accurate. Something like what I described above.

Yes, especially as it is much easier than to fiddle around with autoyast files. We just need to confirm that the changes in the proxy's system can be updated to the new state afterwards.

keichwa:
can you please check whether I'm on the right track.  I mostly reused what we have for upgrading/migrating SUMA Proxy, and I tried to adjust it as needed.

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -0,0 +1,52 @@
...
+. On the {productname} Proxy, stop the proxy service:
++
+----
+spacewalk-proxy stop
+----
+
+. In the {productname} Server {webui}, navigate to menu:Systems[Proxy] and click the name of the proxy.
+. Click menu:Software[Software Channels], and select the openSUSE Leap {opensuse-version} channel.
+. Click menu:Software[Packages], and select all the packages to be updated on the proxy, and then apply the selection.
+. On the {productname} Proxy, start the proxy service:
```
juliogonzalez, line 45: Does it work? If so, fine with me.

modules/upgrade/pages/proxy-autoyast-uyuni.adoc
```diff
@@ -0,0 +1,142 @@
...
+If the server reports that there are updates available for the proxy, do not apply them before completing the cleanup.
+====
+
+
+
+.Procedure: Cleaning Up
+. On the {productname} Server {webui}, navigate to the [guimenu]``System List``, select the proxy, and navigate to the menu:Software[Software Channels] subtab.
+. Clear the old channels.
+. In the [guimenu]``Base Channel`` field, select `openSUSE Leap 15.1 (x86_64)``.
+. In the [guimenu]``Child Channels`` field, select all the recommended channels.
```
juliogonzalez, line 142: Does this procedure work? Otherwise I'd not document it.

juliogonzalez:
> can you please check whether I'm on the right track. I mostly reused what we have for upgrading/migrating SUMA Proxy, and I tried to adjust it as needed.

Please test at least the manual procedure, ask help from @mantel if needed, and feel free to remove autoyast for now if you can't verify it.

modules/upgrade/pages/proxy-autoyast-uyuni.adoc
```diff
@@ -0,0 +1,142 @@
+[[proxy-x]]
```
keichwa, line 1:
```suggestion
[[proxy-uyuni-ay]]
```

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -0,0 +1,98 @@
...
+----
+spacewalk-proxy stop
+----
+
+. In the {productname} Server {webui}, navigate to menu:Systems[Proxy] and click the name of the proxy.
+. Click menu:Software[Software Channels], and as the base channel select the openSUSE Leap {opensuse-version} channel that is listed in the [systemitem]``Customs Channels`` list.
+. In the [guimenu]``Child Channels`` pane, select the {opensuse-version} child channels.
+. Click btn:[Next], and [guimenu]``Confirm Software Channel Change`` with btn:[Confirm].
+. Click menu:Software[Packages > Upgrade], and select all the packages to be updated on the proxy, and then apply the selection.
+. Re-register with the reactivation key.
```
keichwa, line 90:
```suggestion
. Re-register with the reactivation key using the [command]``rhnreg_ks`` command-line utility.
The system will be re-registered with the same id, history, and groups.
// and channels (unless the system's base channel changes).
```

keichwa:
> To me it looks good. Was autoyast procedure tested as well? If so: great! If not, I'd avoid publishing it for now.

the ay procedure is currently not included in the generated docs.  I commented the include statement in the nav.

PR #391, opened 2020-07-26
Supportability and compatibility fixes

paususe:
# Description

Add Ubuntu 20.04 to the summary client matrix and fix the statement about Scientific Linux (we have never been compatible) and RH Satellite Compatibility (gone for years)

# Target branches

- [x] Manager-3.2
- [x] Manager-4.0
- [x] Manager-4.1
- [x] Uyuni

A similar statement is needed for 4.0 and 3.2: we have never supported Scientific Linux, and RH Satellite compatibility is gone for years. Please create them, I don't know what is the right branch.

### DIFF
--- a/modules/ROOT/pages/index-uyuni.adoc
+++ b/modules/ROOT/pages/index-uyuni.adoc
@@ -7,11 +7,10 @@
 {productname} enables you to manage large sets of Linux systems and keep them up to date, with automated software management, asset management, and system provisioning.
 {productname} allows you to maintain a high level of security while effectively managing system life-cycle requirements.
 
-{productname} uses _Salt_ to provide event-driven configuration and management control.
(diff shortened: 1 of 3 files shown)


Loquacity:
:heavy_check_mark: backported to 4.0: https://github.com/uyuni-project/uyuni-docs/commit/3e26081741bded569d07af76a95ba426868eee14

Not sure if it's worth updating in 3.2. @jcayouette ??

paususe:
> heavy_check_mark backported to 4.0: [3e26081](https://github.com/uyuni-project/uyuni-docs/commit/3e26081741bded569d07af76a95ba426868eee14)
> 
> Not sure if it's worth updating in 3.2. @jcayouette ??

There will be no more 3.2 MUs, so it would be web-only.

PR #396, opened 2020-07-28
SD Remote Commands

Loquacity:
# Description

modules/reference/pages/systems/system-details/sd-remote-command.adoc

Moves old content from Ref Guide to extend existing content in the Actions chapter of the Admin Guide.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Partial https://github.com/SUSE/spacewalk/issues/11449

### DIFF
--- a/modules/administration/pages/actions.adoc
+++ b/modules/administration/pages/actions.adoc
@@ -110,22 +110,16 @@ You can see scheduled actions from action chains by navigating to menu:Schedule[
 You can configure clients to run commands remotely.
 This allows you to issue scripts or individual commands to a client, without having access to the client directly.
 
-For Salt clients, remote commands are run from the [path]``/tmp/`` directory on the client.
-To ensure that remote commands work accurately, do not mount ``/tmp`` with the [parameter]``noexec`` option.
-
-[WARNING]
-====
-All commands run from the [guimenu]``Remote Commands`` page are executed as {rootuser} on clients.
-Wildcards can be used to run commands across any number of systems.
-Always take extra care to check your commands before issuing them.
-====
+This feature is automatically enabled on Salt clients, and you do not need to perform any further configuration.
+For traditional clients, the feature is enabled if you have registered the client using a bootstrap script and have enabled remote commands.
+You can use this procedure to enable it manually, instead.
 
 Before you begin, ensure your client is subscribed to the appropriate tools child channel for its installed operating system.
 For more information about subscribing to software channels, see xref:client-configuration:channels.adoc[].
 
 
-// This seems really hacky. Is it still correct? --LKB 2020-07-28
-.Procedure: Configuring Clients to Accept Remote Commands
+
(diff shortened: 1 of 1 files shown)


Loquacity:
@mcalmer I have a question in modules/administration/pages/actions.adoc L127. Can you tell me if this is still a correct procedure? (Or point me to someone who might know?)

modules/administration/pages/actions.adoc
```diff
@@ -117,9 +120,47 @@ Wildcards can be used to run commands across any number of systems.
...
+----
+mkdir -p /etc/sysconfig/rhn/allowed-actions/script
+----
++
+. Create an empty file called [path]``run`` in the new directory.
+This file grants the {productname} Server permission to run remote commands:
++
+----
+touch /etc/sysconfig/rhn/allowed-actions/script/run
+----
```
mcalmer, line 146: This is technically what happens when you call:

$> mgr-actions-control --enable-run

Loquacity: so should we document that instead?

modules/administration/pages/actions.adoc
```diff
@@ -117,9 +120,47 @@ Wildcards can be used to run commands across any number of systems.
 Always take extra care to check your commands before issuing them.
 ====
 
+Before you begin, ensure your client is subscribed to the appropriate tools child channel for its installed operating system.
+For more information about subscribing to software channels, see xref:client-configuration:channels.adoc[].
+
+
+// This seems really hacky. Is it still correct? --LKB 2020-07-28
```
keichwa, line 127: I think when client is registered with the server, these packages should be installed from the server (while bootstrapping or later).  With the activation key, with the Web UI, or API.

modules/administration/pages/actions.adoc
keichwa: Doesn't this happen automatically if you enable remote command with the bootstrap script.  Of course, it would not hurt if we document in the Ref Guide what happens behind the scenes.  But I think we shouldnot advise the user to run those setup steps manually.
This is just my opinion.  Let's wait what mc will tell us :)

modules/administration/pages/actions.adoc
mcalmer: tradidional clients: requires extra packages and you need to explicit enable remote-command execution.

salt clients: everything is there and possible as soon as you have salt-minion installed.

modules/administration/pages/actions.adoc
mcalmer: Yes, this should happen when you use the bootstrap script and have it configured. But I think it is a good idea to document the packages and the command as well. Just in case somebody wants to enable it later.

Please add also that this is only required for traditional clients. Nothing of it is needed when you use a salt client.

Issue #225, opened 2020-04-22
autoyast.html in SUMA 4.0/4.1 docs not linked in sidebar


sknorr:
The SUMA 4.0 and 4.1 documentation contains this file path:

* https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/client-configuration/autoyast.html

* https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/autoyast.html

This AutoYaST section does not seem to be part of the Guide structure in the sidebar. Is that expected?

sknorr:
Ftr, I have no stakes in this issue, I just randomly stumbled across this file because it was linked from somewhere.

Lana just explained to me that this might be the precursor to files being deleted. I am a bit confused about that practice then -- because taking it out of the sidebar tree does not look intentional or explicit to me. Maybe a note saying "hey, do you think this document is helpful? please file a bug report or it will be removed next week" (paraphrasing, obviously) might be a more explicit way of achieving this..?

Loquacity:
So I think this case might actually be an error. Investigating now.

Loquacity:
> Lana just explained to me that removal from the sidebar tree might be the precursor to files being deleted. I am a bit confused about that practice -- because taking it out of the sidebar tree does not look intentional or explicit to me, and there is no discernible difference to an actual bug. Maybe a note saying "hey, do you think this document is helpful? please file a bug report or it will be removed next week" (paraphrasing, obviously) might be a more explicit way of achieving this..?

This is a result of the major content reorganisations we've been doing over the past couple of releases, and also a function of how our search tool works. Usually, we determine that the content needs to be moved elsewhere in the suite, so we comment out the page in the nav, leave a comment to explain why and where the content has moved to (datestamped), but leave the page published. This means that when we come across the page next, we can make an educated guess about whether or not to delete the page, based on how much time has elapsed, and whether or not anyone has yelled. Our search tool indexes the entire suite, not just the pages linked in the nav, so it will turn up results that have been delinked. That means that we have been able to use this method as a sort of "soft" deletion, which has been useful with all the content restructures we have been dealing with. All that said, I do agree that a more formal process is probably required. Now that our docs suite is starting to stabilise, that is something we will be able to introduce. I'll add it to my list, and make sure it gets discussed in planning.

PR #399, opened 2020-07-29
Adding missing ay files to nav

Loquacity:
# Description

The AY files were included in the Client Cfg directory, but missing from the nav. In the absence of any comments to the contrary, I'm assuming they went missing during a recent reorganisation, and have added them back. Additionally: the example file was off in an appendix at the end, and I've moved it to be with its parent.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/225

### DIFF
--- a/modules/client-configuration/nav-client-configuration-guide.adoc
+++ b/modules/client-configuration/nav-client-configuration-guide.adoc
@@ -96,7 +96,6 @@ endif::[]
 // Troubleshooting
 ** xref:tshoot-clients.adoc[Troubleshooting Clients]
 // Appendices
-** xref:autoyast-example.adoc[AutoYaST Example]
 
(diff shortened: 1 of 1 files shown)


Loquacity:
> Description is wrong ;) You finally moved the file from the end to the AY section. Might be a good thing or not ;) In the past, we tried to offer longer examples as an appendix at the end of a book.

Isn't that what I said? ;) Anyway, I think having examples at the end of the book is useful in some situations, especially where we're expecting the book to be read as a narrative or curricula, or if there's no obvious parent section. In this case, though, it makes more sense to me to have it as a child of the AY section.

PR #413, opened 2020-08-03
Update SSH and Azure

Loquacity:
# Description

Adds suggested wording from bug re: using SSH with public clouds, and Azure instructions.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes

* https://bugzilla.suse.com/show_bug.cgi?id=1173596
* https://github.com/SUSE/spacewalk/issues/11857

### DIFF
--- a/modules/quickstart-public-cloud/pages/clients.adoc
+++ b/modules/quickstart-public-cloud/pages/clients.adoc
@@ -22,7 +22,7 @@ Public cloud images usually do not allow SSH login with username and password, b
 If you want to use bootstrap from the {webui}, you will need to enable SSH login with username and SSH key.
 You can do this by navigating to menu:Systems[Bootstrapping] and changing the authentication method.
 
-If your cloud provider is Microsoft Azure, you can login with username and password.
+If your cloud provider is Microsoft Azure, you can log in with username and password.
 To do this, you need to allow the AzureUser to run commands as root without a password.
 To do this, open the [path]``/etc/sudoers.d/waagent`` file, and add or edit this line:
 +


modules/quickstart-public-cloud/pages/clients.adoc
```diff
@@ -17,10 +17,27 @@ You might need to add the FQDN of the server to [path]``/etc/hosts`` local resol
 Check using the [command]``hostname -f`` command with the local IP address of the server.
 ====
 
+
 Public cloud images usually do not allow SSH login with username and password, but only SSH with a certificate.
-If you want to use bootstrap from the {webui}, you will need to enable SSH login with username and password.
+If you want to use bootstrap from the {webui}, you will need to enable SSH login with username and SSH key.
 You can do this by navigating to menu:Systems[Bootstrapping] and changing the authentication method.
 
+If your cloud provider is Microsoft Azure, you can login with username and password.
```
keichwa, line 25:
```suggestion
If your cloud provider is Microsoft Azure, you can log in with username and password.
```

modules/quickstart-public-cloud/pages/clients.adoc
```diff
@@ -17,10 +17,27 @@ You might need to add the FQDN of the server to [path]``/etc/hosts`` local resol
...
+To do this, open the [path]``/etc/sudoers.d/waagent`` file, and add or edit this line:
++
+----
+AzureUser ALL=(ALL) NOPASSWD: ALL
+----
+
+[WARNING]
+====
+Allowing the AzureUser to run commands as root without a password carries a security risk.
+Use this method for testing only.
```
keichwa, line 36: In this case, we should define what testing means. I'd say accessible by the test user only.

Loquacity: I think "test user" is actually a little specific. What I'm trying to say is "use this to try it out, not in prod". I don't want people to think it's safer if they just create a user called "test".

keichwa:

> I think "test user" is actually a little specific. What I'm trying to
> say is "use this to try it out, not in prod". I don't want people to
> think it's safer if they just create a user called "test".

Yes, but if such a system—even if not in prod—is accessible it could be
dangerous.  Probably your wording is good or safe enough, I'm not a
security guy.

-- 
Karl Eichwalder

Loquacity:
> Lana Brindley <notifications@github.com> writes:
> I think "test user" is actually a little specific. What I'm trying to say is "use this to try it out, not in prod". I don't want people to think it's safer if they just create a user called "test".
> Yes, but if such a system—even if not in prod—is accessible it could be dangerous. Probably your wording is good or safe enough, I'm not a security guy.
> […](#)
> -- Karl Eichwalder

I agree, but that's a little more detail than I'm really willing to go into in a quickstart. I was really just after wording that says "this is insecure, probably don't do it".

PR #418, opened 2020-08-04
Language Selection feature

Namanbhatia7:
# Description

To select translation, a hover down menu has been added to Navbar. 

langselfinal

@jcayouette @Loquacity Please have a Look!

### DIFF
--- a/branding/supplemental-ui/uyuni/uyuni-webui/js/vendor/langSelection.js
+++ b/branding/supplemental-ui/uyuni/uyuni-webui/js/vendor/langSelection.js
@@ -1,7 +1,7 @@
 function selectLanguage(langId) {
 
     var str = window.location.pathname;
-    var regexPattern =/\/..\// ;
+    var regexPattern =/\/..\/|\/.._..\//;
 
     var matchCode = "/" + langId + "/" ;
 


paususe:
I like the result but we need a more generic JavaScript and a way to generate it. Say we have 5 languages, the langselection.js code would become very convoluted as it's written. I think a better way would be to:

1. Generate a languages.js during `make`, e g. by having each Makefile.en, Makefile.es, etc append a new line with the language code to languages.js. which would look like:
```
en
es
fr
de
zh_CN
```
Or maybe a JSON file with the language codes (although that would be a bit more difficult to generate in the `make` step).
2. Require/import/read languages.js (or languages.json or whatever)
3. Read all those language codes into a variable and make the current langselection.js a bit more generic so that it can take potentially an infinite number of languages without needing to hardcode them in langselection.js

Also: will the current langselection.js work without JavaScript? That's important for SEO, we definitely want the translated docs to be found by search engines.

jcayouette:
@paususe Agreed, more generic would be better. Hardcoding/updating language variables would not be very much fun. So if I understand you correctly:

1. Generate your generic **languages.js** file during a build. 
2. Each **makefile (en, es, fr, de, zh_CN etc)** will add a new line of code for its corresponding **langcode**.
3. Add the these codes to a variable and make the current langselection.js more generic by allowing selection from said variable.
4. I do not think the lang selection will work without js or?

Namanbhatia7:
@paususe @jcayouette , I got your point of making Javascript more generic. So I have modified langSelection.js to make it more generic in my latest commit (https://github.com/uyuni-project/uyuni-docs/pull/418/commits/76e01bd9f31a02c9d89482d2afb936bb598beab2).

1.)This will solve the problem of hard coding as we do not need to edit any code in langSelection.js as we add more and more translation. 

2.) It will not require any new file to store languages code, as in this code, we can directly take the language code from Frontend (ie Css id ).

3.) We will only have to add a select option in the drop-down menu in header-content.hbs ( ie adding one line of code for each language we want to add). 

So by this way, langSelection.js is made generic as well as optimized .

branding/supplemental-ui/uyuni/uyuni-webui/partials/footer-scripts.hbs
```diff
@@ -67,6 +67,7 @@
 {{#if (eq env.DOCSEARCH_ENGINE 'lunr')}}
     <script src="{{uiRootPath}}/js/vendor/lunr.js"></script>
     <script src="{{uiRootPath}}/js/vendor/search.js"></script>
+     <script src="{{uiRootPath}}/js/vendor/langSelection.js"></script>
```
renner, line 70:
```suggestion
    <script src="{{uiRootPath}}/js/vendor/langSelection.js"></script>
```
I would suggest to use the same indentation as for the other `script` tags?

branding/supplemental-ui/uyuni/uyuni-webui/css/site-extra.css
```diff
@@ -905,3 +905,30 @@ img.errorsvg {
...
+}
+
+
+#en{
+  cursor: pointer;
+}
+
+#es{
+  cursor: pointer;
+}
```
renner, line 934:
```suggestion

/* Language Selection */

img.langIcon.english {
  width: 27px;
  margin-right: 6px;
}

img.langIcon.spanish {
  width: 27px;
  margin-right: 6px;
}

#en {
  cursor: pointer;
}

#es {
  cursor: pointer;
}

```
This is just a general formatting suggestion, i.e. I would recommend to not put so many empty lines for better readability, use a blank before the `{`, etc. Please consider e.g. [these CSS formatting rules](https://google.github.io/styleguide/htmlcssguide.html#CSS_Formatting_Rules) for further general information.

branding/supplemental-ui/uyuni/uyuni-webui/js/vendor/langSelection.js
```diff
@@ -0,0 +1,11 @@
...
+
+    var str = window.location.pathname;
+    var regexPattern =/\/..\// ;
+
+    var matchCode = "/" + langId + "/" ;
+
+    var newURL = str.replace(regexPattern, matchCode);
+    window.location.href = newURL;
+
+}
```
renner, line 11:
```suggestion
}

```
Git likes to have a newline at the end of each source file, so I would recommend adding it.

branding/supplemental-ui/uyuni/uyuni-webui/js/vendor/langSelection.js
```diff
@@ -0,0 +1,11 @@
+function selectLanguage(langId){
```
renner, line 1:
```suggestion
function selectLanguage(langId) {
```
Removing trailing whitespace and adding a blank before the `{`. I would recommend to get familiar with Javascript coding style best practices in general, and even to enforce common guidelines using a linter (e.g. [jslint](https://www.jslint.com/)) as part of the CI.

Namanbhatia7:
@renner Thankyou for pointing out these changes and Sorry for not keeping these formatting rules in my mind. I will do the changes as suggested and keep it in my mind in future!
And regarding the language selection, I tried to keep langSelection.js generic. But I think we will have to add code in header-content.hbs, as that code is a drop down menu and cater to only options we provide to users.. And I thought of a solution  to take Id class similar to language code so that  we do not need any additional file to retrieve language code. by keeping in mind that user selects (say spanish), so with a only a click on an option, we could make selection possible.
This is what I came up with. Please let me know what do you think and suggest me, as it would really help me

jcayouette:
@paususe ^^

paususe (approved):

jcayouette:
@paususe @Namanbhatia7 I have pinged Namen on gitter to fix conflicts and we can merge.

Namanbhatia7:
@jcayouette I have resolved the merge conflicts. I have few question though. 
1.) I have sent @paususe a more generic solution to this feature. So if you could please review it and let me know.

2.) Also, Are we using Zh_CN or Zh in url for chinese language in URL? if Zh_CN we will have to add an exception in langSelection,js to make it work for "Zh_CN" as current code works for ISO 639-1 format of languages Code. 

So let me know and I will make required changes. Thankyou :)

jcayouette:
I am not sure @paususe ^^

jcayouette:
@paususe Are we good to merge this? Or should we wait a bit longer?

jcayouette:
Merging and testing today.

PR #427, opened 2020-08-06
SD - Software Channels

Loquacity:
# Description

modules/reference/pages/systems/system-details/sd-software-channels.adoc

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Partial https://github.com/SUSE/spacewalk/issues/11449

### DIFF
--- a/modules/reference/pages/systems/system-details/sd-software-channels.adoc
+++ b/modules/reference/pages/systems/system-details/sd-software-channels.adoc
@@ -1,38 +1,39 @@
 [[sd-software-channels]]
-= SD Software Channels
+= Software Channels
 
-Software channels provide a well-defined method to determine which packages should be available to a system for installation or upgrade based on its operating systems, installed packages, and functionality.
+The menu:Systems[Software > Software Channels] section allows you to manage the software channels on the current client.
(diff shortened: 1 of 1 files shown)


modules/reference/pages/systems/system-details/sd-software-channels.adoc
```diff
@@ -1,38 +1,39 @@
...
+
+The menu:Systems[Software > Software Channels] subtab is split into sections:
+
+
+
+== Base Channel
+
+This section shows the base channel the current client is subscribed to, as well as any other available base channels.
+
+Click the chain icon next to a base channel to see more information about the channel.
```
keichwa, line 16: Now it sounds as if clicking the chain icon will pop up an innocent info text. But it actually will move you the channel details page that belongs to the Software hierarchy.

Loquacity: Really? I've used that language throughout. We don't use popups much at all, but there was one that recall where I said something about it being a popup specifically.

PR #439, opened 2020-08-12
Add SLE client chapter

Loquacity:
# Description

Adding a specific SLE chapter under the "SUSE Client Registration" section.

# Target branches

- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/12134

### DIFF
--- a/modules/client-configuration/pages/clients-sle.adoc
+++ b/modules/client-configuration/pages/clients-sle.adoc
@@ -12,8 +12,6 @@ Before you register {sle} clients to your {productname} Server, check that you h
 
 The products you need for this procedure are:
 
-ifeval::[{suma-content} == true]
-
 [[opensuse-channels-wizard]]
 [cols="1,1,1,1", options="header"]
 .SLE Products - WebUI
@@ -38,10 +36,11 @@ ifeval::[{suma-content} == true]
 
 include::snippets/add_channels_wizard.adoc[]
 
-endif::[]
 
 
 ifeval::[{uyuni-content} == true]
+
+To add the client tools, add these channels from the command prompt:
 
 [[opensuse-channels-cli]]
 [cols="1,1", options="header"]
@@ -50,7 +49,6 @@ ifeval::[{uyuni-content} == true]
 
 | OS Version
 | Client Channel
-// Can't find a base channel, is it because it's not supported on Uyuni? --LKB 2020-08-12
 
 | {sles} 15 SP1
 | sles15-sp1-uyuni-client
@@ -64,6 +62,7 @@ ifeval::[{uyuni-content} == true]
 include::snippets/add_channels_cli.adoc[]
 
 endif::[]
+
 
 
 == Check Synchronization Status


modules/client-configuration/pages/clients-sle.adoc
```diff
@@ -0,0 +1,112 @@
...
+ifeval::[{uyuni-content} == true]
+
+[[opensuse-channels-cli]]
+[cols="1,1", options="header"]
+.OpenSUSE Channels - CLI
+|===
+
+| OS Version
+| Client Channel
+// Can't find a base channel, is it because it's not supported on Uyuni? --LKB 2020-08-12
```
juliogonzalez, line 53: It *is* supported, but you need the same steps you have above, as you get the channels for SLE from SCC

Loquacity: > Doc for Uyuni needs a fix.

Yes, it will go into Uyuni as well.

Loquacity: Oh, no spacewalk-common-channels option then?

juliogonzalez: Only for the client tools channel. And it's to be executed only after the SLE channels are added from SCC.

However for SUSE manager you don't tell the user to enable the client tools from SCC because they come enabled automatically, so I think you can just have that part common for Uyuni and SUSE Manager. Would be much easier to verify with rendered docs :-D

Loquacity: Oh, sorry, I forgot to upload them. I'll update this and upload some for you.

Loquacity:
image
Uyuni: [uyuni_client-configuration_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/5061723/uyuni_client-configuration_guide.pdf)

Loquacity:
> I general looks good, for Uyuni we tell the users to run `spacewalk-common-channels` but we don't tell them to launch a sync for the channels that get added! (unlike channels added from SCC, channels added by `spacewalk-common-channels` do not get automatically synced).
> 
> But my suggestion is:
> 
> 1. Create an issue for the thing I just explained about `spacewalk-common-channels` and make sure you mention that all other OS should get reviewed as well. I think this got removed during one of the rewrites to use includes.
> 2. Merge this PR as it is, as at least we now have SLE back :-)

It's still there:
image

juliogonzalez:
> It's still there:

You are right, I overlooked it because it under the title "Check syncronization Status".

Problem is that that looks common to both SCC and non SCC channels.

> Or are you saying we should add a CLI sync as a final step in the spacewalk-common-channels procedure?

Not really, but I'd take the instructions to run itself after `spacewalk-common-channels` procedures for all OS.

Anyway: NOT for this PR. Just create an issue for that :-)

Loquacity:
> > It's still there:
> 
> You are right, I overlooked it because it under the title "Check syncronization Status".
> 
> Problem is that that looks common to both SCC and non SCC channels.
> 
> > Or are you saying we should add a CLI sync as a final step in the spacewalk-common-channels procedure?
> 
> Not really, but I'd take the instructions to run itself after `spacewalk-common-channels` procedures for all OS.
> 
> Anyway: NOT for this PR. Just create an issue for that :-)

Sure. I can adjust that for all clients.

juliogonzalez:
Maybe you will want a snippet about how to launch the sync, both with CLI and WebUI, to place it after the `spacewalk-common-channels` procedures.

Loquacity:
:heavy_check_mark: backported to `manager-4.1` https://github.com/uyuni-project/uyuni-docs/pull/439

Issue #284, opened 2020-05-19
Feedback from the field: unify suggested partitions


mbologna:
I got from the field a suggestion: unify suggested partition when installing Uyuni/SUSE Manager.

An example:

We start talking about "Partition permissions" https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/installation/general-requirements.html but the illustration of how to set up those partitions is illustrated later on only for public cloud: https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/installation/pubcloud-requirements.html#_separate_storage_volumes

IMHO I would unify everything into a general requirement:

- we require three different partitions
- we suggest using XFS as fstype for those partitions, backed by LVM
- only for public cloud: consider putting those partitions in a volume - not the root volume

Any suggestion welcome!

Loquacity:
Absolutely agree, Michele. It's in the backlog.

PR #441, opened 2020-08-12
Storage reqs

Loquacity:
# Description

Combines storage device content in Install Guide (from pubcloud reqs and general reqs) and moves them to hardware reqs.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/284

### DIFF
--- a/modules/installation/pages/hardware-requirements.adoc
+++ b/modules/installation/pages/hardware-requirements.adoc
@@ -185,7 +185,7 @@ You must set up the storage devices before you run the YaST {productname} setup 
 * Cache: [path]``/var/cache``
 
 We recommend you use XFS as the filesystem type for all volumes.
-Additionally, consider using logical volume management (LVM) to manage the disks.
+Additionally, for on-premise installations, consider using logical volume management (LVM) to manage the disks.
 The size of the disk for repositories storage is dependent on the number of distributions and channels you intend to manage with {productname}.
 See the tables in this section for guides to estimate the size required.
 
--- a/modules/installation/pages/pubcloud-requirements.adoc
+++ b/modules/installation/pages/pubcloud-requirements.adoc
@@ -63,6 +63,7 @@ To access the {productname} {webui}, allow HTTPS when configuring the network ac
 
 We recommend that the repositories and the database for {productname} are stored on separate storage devices to the root volume.
 This will help to avoid data loss.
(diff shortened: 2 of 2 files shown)


modules/installation/pages/hardware-requirements.adoc
```diff
@@ -154,15 +160,95 @@ For more information about inter-server synchronization (ISS), see xref:administ
...
+You must set up the storage devices before you run the YaST {productname} setup procedure.
+
+{productname} requires three different volumes:
+
+* Database volume: [path]``/var/lig/pgsql``
+* Channel volume: [path]``/var/lib/spacewalk``
+* Cache: [path]``/var/cache``
+
+We recommend you use XFS as the filesystem type for all volumes.
+Additionally, consider using logical volume management (LVM) to manage the disks.
```
mbologna, line 188: I am considering a SUSE Manager running in the public cloud here.
XFS is sure an option _if_ the volumes above are non-root volumes (e.g. AWS EBS).
LVM: I do not think that is feasible or recommended in a public cloud setup.

@jgleissner do you have any input in this regard?

jgleissner: @mbologna XFS is recommended in the public cloud too. We even use it for root volumes. LVM is of limited use in the cloud and not recommended.

Loquacity: OK, so the hardware reqs page should be advice common to all installations, and we can keep cloud-specific advice to the pubcloud reqs section. With that in mind, it seem that I can keep the hardware reqs page as-is, but update the pubcloud reqs page to read "do not use LVM"?

mbologna: > OK, so the hardware reqs page should be advice common to all installations, and we can keep cloud-specific advice to the pubcloud reqs section. With that in mind, it seem that I can keep the hardware reqs page as-is, but update the pubcloud reqs page to read "do not use LVM"?

If the hardware requirements (this page) is common to all installations, then my suggestion is to be as general as possible: remove the LVM sentence from this and just add it to the "on-prem" documentation. So we do not have conflicting information between general and specific case setup.

Loquacity: Updated.

PR #445, opened 2020-08-13
Normalise overviews - All books

Loquacity:
# Description

Ensures each top level file in a book is named correctly (`<Book Name> Overview`), and that we don't have nav entries without an associated file.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/11179

### DIFF
--- a/modules/ROOT/pages/index.adoc
+++ b/modules/ROOT/pages/index.adoc
@@ -80,15 +80,15 @@ Download All PDFs icon:caret-right[] icon:file-archive[link="../uyuni-docs_en-pd
 |===
 | View HTML | View PDF | View HTML | View PDF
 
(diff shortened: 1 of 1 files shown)


Loquacity:
Now to see what xrefs I've broken ... :worried:

jcayouette:
woops, mb for leaving the index-uyuni.adoc file around. This is actually depreciated as we used it prior to the ifdef statements in the index.adoc file. We can remove it.

Loquacity:
> woops, mb for leaving the index-uyuni.adoc file around. This is actually depreciated as we used it prior to the ifdef statements in the index.adoc file. We can remove it.

I *thought* it looked outdated!

PR #447, opened 2020-08-13
SP mass migration examples

keichwa:
https://github.com/SUSE/spacewalk/issues/12082

# Description

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [ ] Uyuni

# Links

about https://github.com/SUSE/spacewalk/issues/12082

### DIFF
--- a/modules/upgrade/pages/client-x.adoc
+++ b/modules/upgrade/pages/client-x.adoc
@@ -170,6 +170,7 @@ for system in system_ids
 endfor
 ----
 
+////
 [WARNING]
 ====
 The following examples are referenced for illustrative purposes only.
@@ -180,3 +181,4 @@ If you want to integrate such API calls into your scripts, see:
 
 * https://github.com/bjin01/spmigration
 * https://github.com/bjin01/spmigration-single
+////


modules/upgrade/pages/client-x.adoc
```diff
@@ -169,3 +169,14 @@ for system in system_ids
...
+[WARNING]
+====
+The following examples are referenced for illustrative purposes only.
+{suse} does not support these resources.
+====
+
+If you want to integrate such API calls into your scripts, see:
+
+* https://github.com/bjin01/spmigration
+* https://github.com/bjin01/spmigration-single
```
mcalmer, line 182: I do not like to have private github repos linked here. We have no control if he forget about it and remove the files.
We need an own github repo for any kind of example files and scripts.

We have already https://github.com/SUSE/manager-build-profiles but the name is not good for a generic SUSE Manager examples repository.

I would like to see the doc team to create such a repository.

renner: I agree that there is certain risks here, e.g. we never tested those scripts, they might stop working at some point because our APIs change or they might even disappear if the author decides to delete them. Having a repo owned by us (e.g. `example-scripts` at the `uyuni-project` organization) with such examples would at least prevent the author from deleting them.

While the doc team could own this repo I understand they would probably not be the ones to test and continually maintain those scripts. For this specific case, could we start with creating such a repository and ask @bjin01 if he would be okay with moving his scripts there? If needed we could make him an external repository admin so he can merge PRs according to the [roles and permissions](https://github.com/uyuni-project/uyuni/wiki/Roles-and-permissions).

Loquacity:
@keichwa probably best to merge this, and create an issue to re-assess how we handle scripts.

keichwa:
Ok, but we must hear from Bo. Maybe, I should merge it and then disable it until we have a solution for third-party scripts.

Loquacity:
> Ok, but we must hear from Bo. Maybe, I should merge it and then disable it until we have a solution for third-party scripts.

Sure, that's a reasonable solution.

PR #453, opened 2020-08-19
add sync to add channels CLI

Loquacity:
# Description

In the Client Cfg, we tell people how to view the status of a channel sync (because we assume WebUI, where the sync starts automatically). This adds the mgr-sync command to the add_channels_cli snippet, so that a sync is triggered after the channels are added, and checking the status of the sync makes sense in context.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/12150


jcayouette (approved):
LGTM.

PR #458, opened 2020-08-20
CEFS clarifications

moio:
# Description

Add some clarifications and a warning box to the CEFS instructions.

# Target branches

- [ ] Manager-4.0
- [x] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Context: https://github.com/SUSE/spacewalk/issues/11981#issue-661505743 (bottom of description)

### DIFF
--- a/modules/client-configuration/pages/clients-centos.adoc
+++ b/modules/client-configuration/pages/clients-centos.adoc
@@ -149,7 +149,7 @@ It is not supported by {suse}.
 
 endif::[]
 
-The authors of CEFS provide patches (errata) on a best-effort basis, in the hope they will be useful but with no guarantees of correctness or currency.
+The authors of CEFS provide patches or errata on a best-effort basis, in the hope they will be useful but with no guarantees of correctness or currency.
 This could mean that the patch dates could be incorrect, and in at least one case, the published data was shown to be more than a month old.
 For more information on these cases, see https://github.com/stevemeier/cefs/issues/28#issuecomment-656579382 and https://github.com/stevemeier/cefs/issues/28#issuecomment-656573607.
 


modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -193,10 +202,16 @@ export SPACEWALK_USER='<adminname>';export SPACEWALK_PASS='<password>'
...
 ----
-. Open the ``rhn.conf`` configuration file, and add or edit this line:
-+
-----
-java.allow_adding_patches_via_api = centos7-x86_64-updates,centos7-x86_64,centos7-x86_64-extras
-----
 
 For more information on this tool, see https://cefs.steve-meier.de/.
+
+[WARNING]
```
Loquacity, line 208: Can we move this above the procedure?

Loquacity, line 208: Can we move this above the procedure? There is an `important` admon on L144, and I think we can probably combine them.

moio: Makes sense, addressed.

modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -139,16 +139,26 @@ For more information about how to resolve this error, see ``Registering Older Cl
...
 
 endif::[]
 
+Please note that CEFS authors provide errata (patch) data on a best-effort basis, in the hope they will be useful but with no guarantees of correctness or currency.
+
+In particular, it is acknowledged that patch (errata) dates might be incorrect (https://github.com/stevemeier/cefs/issues/28#issuecomment-656579382) and it was verified in at least one case that published data was more than one month old (https://github.com/stevemeier/cefs/issues/28#issuecomment-656573607).
+
+Additionally, any defects in the service might result in unreliable patch information to be imported in your server. Consequently any reports, audits or other patch-related information produced in terms of currency of your systems, including CVE patch applicability, might be incorrect as a result.
+
+Please consider alternatives such as not relying on patch data or using different OS distributions depending on your security-related requirements and certifications criteria.
```
Loquacity, line 158:
```suggestion
The authors of CEFS provide patches and updates on a best-effort basis, in the hope they will be useful but with no guarantees of correctness or currency.
This could mean that the patch dates could be incorrect, and in at least one case, the published data was shown to be more than a month old.
For more information on these cases, see https://github.com/stevemeier/cefs/issues/28#issuecomment-656579382 and https://github.com/stevemeier/cefs/issues/28#issuecomment-656573607.

Any problems or delays with the patch data might result in unreliable patch information being imported to your {productname} Server. 
This would cause reports, audits, CVE updates, or other patch-related information to also be incorrect.
Please consider alternatives to using this service, such as independently verifying patch data, or choosing a different operating system, depending on your security-related requirements and certifications criteria.
```

Loquacity (changes requested):

modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -139,16 +139,26 @@ For more information about how to resolve this error, see ``Registering Older Cl
...
-[IMPORTANT]
-====
-The third-party errata service described here is provided and maintained by the community.
+The third-party errata service described here, CEFS, is provided and maintained by the community.
 It is not supported by {suse}.
-====
 
 endif::[]
 
+The authors of CEFS provide patches (errata) on a best-effort basis, in the hope they will be useful but with no guarantees of correctness or currency.
```
moio, line 152: @Loquacity: I took the liberty to change this line after your suggestions from "patches and updates" to "patches (errata)".

Rationale: CEFS does not provide updates per se, only metadata about updates, so "updates" does not seem to me to be the optimal word here. I could live with "patches" only, as it is understood / should be in our glossary that patches are pure metadata.

Reason why I mentioned errata is that "patch" is SUSE/Uyuni/Manager jargon and "errata" is Red Hat/Spacewalk/Satellite jargon for exactly the same concept. When browsing for information around CEFS or in CEFS script commandline help, "errata" is used exclusively, while we use "patch" everywhere in our docs. So I thought about adding a suggestion just in case the reader is not really aware they are synonyms.

Is this OK from your POV? Suggestions welcome!

Loquacity: ah, ok. i really dislike parentheses, though. can we make it "patches or errata" ?

moio: Sure, fixed

Loquacity (approved):

Loquacity:
:heavy_check_mark: backported to `manager-4.1`

PR #460, opened 2020-08-21
add note for salt_enabled_kickstart_install_types property

admd:
As part of  fix for https://bugzilla.suse.com/show_bug.cgi?id=1164836, we introduced a new property in rhn.conf to configure if salt should be used for registration. This PR adds a note for that.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes #https://github.com/SUSE/spacewalk/pull/12226

### DIFF
--- a/modules/reference/pages/systems/autoinst-profiles.adoc
+++ b/modules/reference/pages/systems/autoinst-profiles.adoc
@@ -35,6 +35,13 @@ xref:reference:systems/autoinst-distributions.adoc[]
 
 +
 
++
+.Using SALT for registration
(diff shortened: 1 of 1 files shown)


Loquacity:
This page will get updated very soon, as we finish off the Ref Guide (it's one of the last ones to be looked at). For that reason, I'm approving this even though it doesn't match our current style.

Loquacity (approved):
This page will get updated very soon, as we finish off the Ref Guide (it's one of the last ones to be looked at). For that reason, I'm approving this even though it doesn't match our current style.

jcayouette (approved):

PR #463, opened 2020-08-23
Add initial support for Weblate

paususe:
# Description

To reduce number of translation domains, add scripts for merge of small
asciidoc pot+po files to a larger pot and po files and splitting them
back.

There is currently missing a method for obsoleting of deleted strings in
join-pot-po-files.sh. It probably needs additional msgmerge.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

### DIFF
--- a/l10n-weblate/join-pot-po-files.sh
+++ b/l10n-weblate/join-pot-po-files.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -o errexit
+
+for DIR in ../l10n/pot/* ; do
+	if ! test -d "$DIR" ; then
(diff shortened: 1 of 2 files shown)


paususe (commented):
OK for a first translation. In the near future, We will need to introduce new procedures, or changes to these scripts (and make_pot.sh and use_po.sh) to better manage translations. E. g. we may want to avoid storing the individual .pot and .po files for each .adoc file, which means deep changes in split-po-files.sh.

PR #464, opened 2020-08-24
Update translations

paususe:
# Description

Update translations to latest AsciiDoc, remove duplicates (GFDL in each guide) and old files, and add an empty Simplified Chinese directory.

Since WebLate is not capable of dealing with multiple .po files per translation (it would generate a ton of translation domains), use @stanislav-brabec 's scripts to consolidate each guide in a single file per language.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links


stanislav-brabec:
Please note that there is still a minor issue: If some string will become obsolete, recent scripts are not capable to mark it obsolete in the large files. It needs some more po file manipulation to do that. Anyway, it looks usable.

PR #466, opened 2020-08-24
add section re updates channels on sle

Loquacity:
# Description

Adds short section about updates channels on SLE clients.

SME: @mcalmer 

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/12172

### DIFF
--- a/modules/client-configuration/pages/channels.adoc
+++ b/modules/client-configuration/pages/channels.adoc
@@ -35,13 +35,13 @@ On Salt clients, you must apply the highstate after subscribing to the channel t
 
 
 
-== Packages Provided by Appstream
+== Packages Provided by AppStream
 
-Appstream is a modular package repository that provides additional software for most non-{suse} Linux distributions, including {redhat} and {oracle}.
-In most cases, you will need to install the base channel for your distribution, and the appropriate Appstream channel to ensure that you have all the software you need.
+AppStream is a modular package repository that provides additional software for most non-{suse} Linux distributions, including {redhat} and {oracle}.
+In most cases, you will need to install the base channel for your distribution, and the appropriate AppStream channel to ensure that you have all the software you need.
 
-If you need to transform Appstream repositories into non-modular repositories for use with some upgrade operations, you can use the Appstream filter with the content lifecycle management tool.
-For more information about the Appstream filter, see xref:administration:content-lifecycle-examples.adoc[].
+If you need to transform AppStream repositories into non-modular repositories for use with some upgrade operations, you can use the AppStream filter with the content lifecycle management tool.
+For more information about the AppStream filter, see xref:administration:content-lifecycle-examples.adoc[].
 
 
 
--- a/.changelog
+++ b/.changelog
@@ -6,6 +6,8 @@
 
 # Last packaged: 2020-07-02
 
+
+- Reference example scripts for SP Mass Migration in Upgrade Guide
 - Move PoS Terminal Requirements to the Requirements sections in the Retail Guide
 - Updated SP Mass Migration section in Upgrade Guide for clarity
 - Documented Proxy Y Upgrade (SP Migration) in Upgrade Guide
@@ -23,6 +25,7 @@ terminology for versioning.
 - Added more concepts to Client Cfg
 - Documented maintenance windows feature in Admin Guide
 - Some reorganization of Client Cfg & Admin Guides
+- Updates storage device requirements in Install Guide
 - Adds new section for SUMA formulas in the Salt Guide
 - Updates storage device requirements in Install Guide
 - Added reverse proxy information to Monitoring in Admin Guide
--- a/.github/workflows/main_uyuni.yml
+++ b/.github/workflows/main_uyuni.yml
@@ -24,6 +24,12 @@ jobs:
         node-version: '10.x'
     #- run: npm install
 
+    - name: '[TOOLCHAIN] po4a'
+      run: |
+        sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E77D14FE3C9A28E588A34E7469977F2B42813EC6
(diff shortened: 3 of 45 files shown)


modules/client-configuration/pages/channels.adoc
```diff
@@ -42,3 +42,13 @@ In most cases, you will need to install the base channel for your distribution,
 
 If you need to transform Appstream repositories into non-modular repositories for use with some upgrade operations, you can use the Appstream filter with the content lifecycle management tool.
 For more information about the Appstream filter, see xref:administration:content-lifecycle-examples.adoc[].
+
+
+
+== Updates Channels on {sle} Clients
```
mcalmer, line 48: There is an important difference between the normal update channel and the "installer update channel".

Update channel exist for all products.

New are the Installer Updates Channel which contain only updates for the  OS "installer" (yast, libzypp, curl, etc).
These channels require in SCC no credentials and are used to update the unified installer media before it installes the OS.
All packages which are in Installer Updates Channel are also in the normal Update Channel. But the normal Update Channel can only be used after the product is registered. Installer Update Channels can be used also before registration and therefore can also fix bugs in the registration code itself.

We need to make it clear. This section is for the Installer Updates.

mcalmer (changes requested):

modules/client-configuration/pages/channels.adoc
Loquacity: ah ha, that makes sense. i'll edit.

keichwa:
I think we'd better call it "Unified Installer Update Channel(s)" instead of "... Update/s/ Channel(s)".  Even if a channel offers more then one update we normally use singular.

modules/client-configuration/pages/channels.adoc
```diff
@@ -42,3 +42,15 @@ In most cases, you will need to install the base channel for your distribution,
 
 If you need to transform Appstream repositories into non-modular repositories for use with some upgrade operations, you can use the Appstream filter with the content lifecycle management tool.
 For more information about the Appstream filter, see xref:administration:content-lifecycle-examples.adoc[].
+
+
+
+== Unified Installer Updates Channels on {sle} Clients
+
+All {sle} clients require access to an installer updates channel.
+This channel is used by the unified installer to ensure it is up to date before it installs the operating system.
```
mcalmer, line 51:
```suggestion
This channel is used by the unified installer to ensure it is up to date before it installs the operating system.
All {sle} proucts should have access to the installer updates channel during installation.
```

modules/client-configuration/pages/channels.adoc
```diff
@@ -42,3 +42,15 @@ In most cases, you will need to install the base channel for your distribution,
...
+
+For {sles} clients the installer updates channel is installed by default when you add a product that contains them, and are enabled when you first synchronize the client.
```
mcalmer, line 53:
```suggestion
For {sles} clients the installer updates channel is synced by default when you add a product that contains them, and are enabled when you create an autoinstallation distribution with these product channels.
```

modules/client-configuration/pages/channels.adoc
```diff
@@ -42,3 +42,15 @@ In most cases, you will need to install the base channel for your distribution,
...
+
+For all other {sle} variants, including {sle} for SAP,  you must add the installer updates channel manually.
+To do this, clone the installer updates channel that belongs to the appropriate base channel, and assign it to the client.
```
mcalmer, line 56:
```suggestion
To do this, clone the appropriate {sles} installer updates channel below the base channel of these {sle} variants.
When creating a autoinstallation distribution for these {sle} variants after the channel was cloned , it will be used automatically.
```

mcalmer (changes requested):
Much better, but please check my suggest to make things even more clear.

modules/client-configuration/pages/channels.adoc
Loquacity: Why reverse these sentences?

jcayouette (approved):
LGTM.

modules/client-configuration/pages/channels.adoc
mcalmer: The registered client does not need this channel. So I think it is more important to say that the unified installer need it for installing a client.
But more important is the change that the client does not need access to that channel, but all products should get it. If this is given, the rest should work out of the box.
You can change the order, but the change of the first sentence is important from my point of view. Otherwise we give a wrong impression.

Loquacity: OK, updated.

Loquacity:
All the changes are in, and I've done another editing pass.

mcalmer (approved):
Perfect

modules/client-configuration/pages/channels.adoc
```diff
@@ -42,3 +42,16 @@ In most cases, you will need to install the base channel for your distribution,
...
+== Unified Installer Updates Channels on {sle} Clients
+
+This channel is used by the unified installer to ensure it is up to date before it installs the operating system.
+All {sle} proucts should have access to the installer updates channel during installation.
+
+For {sles} clients the installer updates channel is synchronized by default when you add a product that contains them, and are enabled when you create an autoinstallation distribution with these product channels.
+
+For all other {sle} variants, including {sle} for SAP,  you must add the installer updates channel manually.
+To do this, clone the appropriate {sles} installer updates channel below the base channel of these {sle} variants.
+When creating a autoinstallation distribution for these {sle} variants after the channel was cloned , it will be used automatically.
```
keichwa, line 57:
```suggestion
When creating a autoinstallation distribution for these {sle} variants after the channel was cloned, it will be used automatically.
```

keichwa (approved):

PR #469, opened 2020-08-24
l10n-weblate: Fix invalid header in some zh_CN.po files.

stanislav-brabec:
Author: Stanislav Brabec <sbrabec@suse.cz>

# Description

Weblate fails on po files with a broken header. Fix it in the weblate directory. Next call of split-po-files.sh should spread the fix across all po files in l10n.

The failing header is a bug of some versions of msginit.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [ ] Manager-4.2
- [*] Uyuni

# Links


mcalmer (approved):

jcayouette (approved):

paususe (approved):
@stanislav-brabec can you point me to a safe version of msginit, so that we add that to the README?

Issue #480, opened 2020-08-26
Typo Error "Packages Provided by {suse} Package Hub" ?


Shirocco88:
Hallo Uyuni-Team,
I don't know what "Packages Provided by {suse} Package Hub"
see:
https://www.uyuni-project.org/uyuni-docs/uyuni/client-configuration/channels.html#_packages_provided_by_suse_package_hub

Maybe a Typo ?

Best Regards.
Shirocco88

Loquacity:
Ah, this is interesting. The entity in question is {phub} which is defined as '{suse} Package Hub', so it seems the {suse} entity is not resolving properly. I can fix this.

PR #490, opened 2020-08-28
Update {phub} entity for HTML only

Loquacity:
# Description

In HTML, the {phub} entity was not expanding the {suse} entity. Confirmed works in PDF. 

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/480

### DIFF
--- a/suma-site.yml
+++ b/suma-site.yml
@@ -75,7 +75,7 @@ asciidoc:
     suse: SUSE
     slea: SLE
     sleda: SLED
-    phub: '{suse} Package Hub'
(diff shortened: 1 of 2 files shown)


keichwa (approved):
ok, but do we still need the single quotes?

Loquacity:
> ok, but do we still need the single quotes?

Yeah, they're required for anything that includes a space, afaict.

PR #492, opened 2020-08-28
Firewall Tshooting

Loquacity:
# Description

Adds a new section for troubleshooting firewalls to the Admin Tshooting chapter. Also adds an xref from the Ports chapter in the Install Guide.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/12225

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -31,6 +31,7 @@ terminology for versioning.
 - Added reverse proxy information to Monitoring in Admin Guide
 - Add note about accessibility to index
 - Add note about CentOS upstream repository (bsc#1173603)
-- Add section about SLE updates channels in Client Cfg
 - Add firewall troubleshooting to Admin Guide
+- Fix Azure command in Install Guide (thanks Rahul-CTS)
+- Fix broken links in Auto-Install Proxy in Client Cfg (thanks @shirocco88)
 - Adds Ubuntu 20.04 supported features for Uyuni in Client Cfg
--- a/modules/architecture/pages/salt-ssh-push.adoc
+++ b/modules/architecture/pages/salt-ssh-push.adoc
@@ -12,7 +12,6 @@ Salt SSH Push is intended to be used in environments where your Salt clients can
 [NOTE]
 ====
 This feature is not related to Push via SSH for the traditional clients.
-For Push via SSH, see pass:c[xref:bp.contact.methods.ssh.push[Salt SSH Push]].
 ====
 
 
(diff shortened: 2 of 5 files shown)


modules/administration/pages/tshoot-firewalls.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+Consequence: The sync to SCC fails, and the third party products are not shown in the product list.
+
+Fix: Configure the firewall to reject requests from SUMA instead of drop (preferred), or configure a firewall on the server (if no ability to change firewall settings)
+
+Result: Sync to SCC will either be able to reach the URLs required, or will have request rejected so that the request fails rather than times out.
+////
+
+If you are using a firewall that blocks outgoing traffic, it will either ``REJECT`` or ``DROP`` network requests.
+If it is set to ``DROP`` then you might find that synchronizing with the {scc} times out when you are using non-{suse} clients, including {centos}, {ubuntu}, and {oracle}.
```
mcalmer, line 32: It will always time out. Not only when you use these repos.

modules/administration/pages/tshoot-firewalls.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+This occurs because the synchronization process for non-{suse} clients needs to access the third-party repositories that provide packages for these clients, and not just the {scc}.
```
mcalmer, line 34: The reason is, that we "test" access to these external repos. The product page should only show products where the customer has access to. That requires to do a http request to "all" repositories  we can support.

modules/administration/pages/tshoot-firewalls.adoc
```diff
@@ -0,0 +1,47 @@
...
+When the {productname} Server attempts to reach these repositories to check that they are valid, the firewall drops the requests, and the synchronization continues to wait for the response until it times out.
+
+If this occurs, you will notice that the synchronization will take a long time before it fails, and that your non-{suse} products are not shown in the product list.
+
+You can fix this problem in several different ways.
+
+The simplest method is to configure your firewall to allow access to the URLs required by the non-{suse} repositories you are using.
```
mcalmer, line 41: ... in case you want to use it.

modules/administration/pages/tshoot-firewalls.adoc
```diff
@@ -0,0 +1,47 @@
...
+This allows the synchronization process to reach the URLs and complete successfully.
+
+If allowing external traffic is not possible, configure your firewall to ``REJECT`` requests from {productname} instead of ``DROP``.
+This rejects requests to third-party URLs, so that the synchronization fails rather than times out.
```
mcalmer, line 45:
```suggestion
This rejects requests to third-party URLs, so that the synchronization fails early rather than times out, and the products are not shown in the list.
```

mcalmer (changes requested):
Some suggestions and change requests

modules/administration/pages/tshoot-firewalls.adoc
Shirocco88: Hallo Uyuni-Team,
what about Debian Client System

modules/administration/pages/tshoot-firewalls.adoc 
>> when you are using non-{suse} clients, including {centos}, {ubuntu}, and {oracle}.

Could you please add {debian) also.

Thanks in advance.
Best wishes.
Shirocco88

Loquacity: Hi Shirocco88, two points on this: Firstly, I've removed that line now for other reasons (see earlier comments), and secondly, this is just a list of example third-party clients. The purpose of adding a couple of examples is to make it clearer to the reader what we mean by "third-party" or "non-SUSE" clients. For the purposes of example, I picked the most common third-party clients, it was not intended to be an exhaustive list.

keichwa (approved):

jcayouette (approved):

PR #501, opened 2020-09-01
Systems - Activation Keys

Loquacity:
# Description

modules/reference/pages/systems/activation-keys.adoc

Moves old content from Ref Guide to the Activation Keys chapter in the Client Cfg.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Partial https://github.com/SUSE/spacewalk/issues/10600

### DIFF
--- a/modules/client-configuration/pages/activation-keys.adoc
+++ b/modules/client-configuration/pages/activation-keys.adoc
@@ -52,20 +52,40 @@ Ensure you go back and check the box if you need to enable configuration managem
 
 
 
-== Combining Activation Keys
+== Combining Mutiple Activation Keys
 
 You can combine activation keys when executing the bootstrap script on your traditional clients.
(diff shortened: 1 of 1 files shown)


jcayouette (approved):

keichwa:
I do not know which info we have where, but some sections are important, for example: "Using Multiple Activation Keys at Once". The restrictions with the activation key name are important and belong to the Reference book, IMO.

keichwa (approved):

Loquacity:
> I do not know which info we have where, but some sections are important, for example: "Using Multiple Activation Keys at Once". The restrictions with the activation key name are important and belong to the Reference book, IMO.

OK, I'll move the multiple activation key stuff out. Which restrictions do you mean? the bit about special characters (`< > ( ) { }`)?

PR #532, opened 2020-09-14
OpenSCAP edits

Loquacity:
~***DO NOT MERGE UNTIL DOCS FREEZE IS COMPLETE***~

# Description

Edits to the OpenSCAP chapter proposed by @Bischoff  

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes:

* https://github.com/SUSE/spacewalk/issues/12406
* https://bugzilla.suse.com/show_bug.cgi?id=1176413

### DIFF
--- a/modules/administration/pages/openscap.adoc
+++ b/modules/administration/pages/openscap.adoc
@@ -12,7 +12,7 @@ OpenSCAP auditing is not available on Salt SSH clients.
 
 [IMPORTANT]
 ====
-Scanning clients can consume a lot of memory on both the server and the client being scanned.
+Scanning clients can consume a lot of memory and compute power on the client being scanned.
 For {redhat} clients, ensure you have at least 2{nbsp}GB of RAM available on each client to be scanned.
 ====
 


modules/administration/pages/openscap.adoc
```diff
@@ -37,23 +45,22 @@ For more information about the {suse} Security Team, see  https://www.suse.com/s
 == Prepare Clients for an SCAP Scan
```
keichwa, line 45: Already in the intro, you want the user to install a package: spacewalk-oscap.  I think this should be mentioned here.  And here we'd better also say clearly that the packages are to be installed on the client.

modules/administration/pages/openscap.adoc
```diff
@@ -37,23 +45,22 @@ For more information about the {suse} Security Team, see  https://www.suse.com/s
...
-
 
+For scanning {sles} clients, install the ``openscap-content`` and ``openscap-utils`` packages before you begin.
 Use this command to determine the location of the appropriate SCAP files.
-Take a note of these paths for performing the scan:
-+
+Take a note of the file paths for performing the scan:
+
 ----
 rpm -ql openscap-content
```
keichwa, line 54: Wondering whether you can do those package listings with the Web UI?

modules/administration/pages/openscap.adoc
```diff
@@ -13,6 +13,14 @@ OpenSCAP auditing is not available on Salt SSH clients.
 To use openSCAP, you will need the [systemitem]``spacewalk-oscap`` package installed on the clients you want to audit.
 
 
+[IMPORTANT]
+====
+Scanning clients can consume a lot of memory on both the server and the client being scanned.
+You should ensure you have at least 2{nbsp}GB of RAM spare on each client to be scanned.
```
Bischoff, line 19: No, we're okay with default-advised memory on SUSE clients (1 GB).

The problem is only for RedHat clients (where you need at least 1.25 GB according to my tests, but let's advice to use 2 GB).

modules/administration/pages/openscap.adoc
```diff
@@ -37,23 +45,22 @@ For more information about the {suse} Security Team, see  https://www.suse.com/s
...
+
 ----
 rpm -ql openscap-content
 ----
-+
-The output should include these files:
-+
+
+
+For scanning {redhat} clients, install the ``scap-security-guide`` package before you begin.
```
Bischoff, line 58: and ``openscap-utils``

Bischoff (changes requested):
a few changes please

modules/administration/pages/openscap.adoc
Bischoff: Maybe with "run remote command" :-(

From what i can see, in "minion details -> Audit -> Schedule, the field "Path to XCCDF document" is a free input text field, so it's not suggesting a valid value. You have to run such a command to know what to enter there. Of course a drop-down with the valid values would make user"s life easier.

modules/administration/pages/openscap.adoc
Loquacity: I went the other way, and took the line out of the intro. It doesn't really fit well there, I think.

modules/administration/pages/openscap.adoc
Loquacity: Something for the UI team :wink:

modules/administration/pages/openscap.adoc
Loquacity: Ah, I misunderstood. Updated.

keichwa (approved):

modules/administration/pages/openscap.adoc
```diff
@@ -10,7 +10,11 @@ It allows you to schedule and view compliance scans for any client.
 OpenSCAP auditing is not available on Salt SSH clients.
 ====
 
-To use openSCAP, you will need the [systemitem]``spacewalk-oscap`` package installed on the clients you want to audit.
+[IMPORTANT]
+====
+Scanning clients can consume a lot of memory on both the server and the client being scanned.
```
Bischoff, line 15: on the server too ? i don't think i said that ?

where did you get this info ?

Bischoff (changes requested):
still a small gotcha

Bischoff (approved):
LGTM now. Thanks for your patience.

PR #562, opened 2020-09-24
search bug resolved

Namanbhatia7:
# Description
This bug was due to the url difference in running site on local server and hosted site as Guide name in search result is taken from url.
bug-search

@jcayouette @Loquacity Please have a look!


Loquacity:
Looks good, thanks Namen!

jcayouette:
Looks great thanks!

PR #564, opened 2020-09-25
Registering Public Cloud Clients

Loquacity:
# Description

Adds new section to Client Cfg on registering public cloud clients, with xref from the QS:PC

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/12340

### DIFF
--- a/modules/client-configuration/pages/registration-pubcloud.adoc
+++ b/modules/client-configuration/pages/registration-pubcloud.adoc
@@ -4,8 +4,8 @@
 If you have your clients running in a public cloud instance, there are several methods for registering them to your {productname} Server, in addition to the regular bootstrapping methods.
 For other registration methods, see xref:client-configuration:registration-overview.adoc[].
 
-By default, logging for both the {productname} Server and Salt clients is disabled.
-You can enable logging on Salt clients to provide some additional information during registration.
+By default, logging for both the {productname} Server and Salt clients is set to show warnings only.
+You can increase the logging level on Salt clients to provide some additional information during registration.
 Open or create the [path]``/etc/salt/minion.d/log_level.conf`` configuration file on the client, and add or edit this line:
 
 ----
@@ -57,4 +57,9 @@ For a copy of the script, see xref:client-configuration:registration-pubcloud-sc
 == Register Clients Using the API
 
 You can also use the API to register clients.
+These API calls are used for client registration:
+
+* system.bootstrap
+* system.bootstrapWithPrivateSshKey
+
 For the most recent API documentation, see https://documentation.suse.com/suma.


jcayouette (approved):

modules/client-configuration/pages/registration-pubcloud.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+----
+log_level: info
+----
+
+
+
+== Register Salt Clients in the Instance
+
+This method makes the Salt client part of the public cloud image.
```
mcalmer, line 19: This does not sound right. When you register an instance you do not change the image where it was created from.
Or is this special public cloud terminology that "image" is equal to "instance"?

modules/client-configuration/pages/registration-pubcloud.adoc
```diff
@@ -0,0 +1,60 @@
...
+----
+salt-key -a '<client>'
+----
+
+
+
+== Register Clients with SSH
+
+You can use the ``mgr-ssh-push-init`` script to register both traditional and Salt clients with SSH, or SSH tunnelling.
+For a copy of the script, see xref:client-configuration:registration-pubcloud-script.adoc[].
```
mcalmer, line 53: Better than referencing a written copy of a tool which is installed on every server would be an example how to call it.
Maybe we have such an example in a different chapter?

modules/client-configuration/pages/registration-pubcloud.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+
+
+== Register Clients Using the API
+
+You can also use the API to register clients.
+For the most recent API documentation, see https://documentation.suse.com/suma.
```
mcalmer, line 60: Maybe we can be a bit more specific. 

The API calls are 

- bootstrap
- bootstrapWithPrivateSshKey

mcalmer (changes requested):

modules/client-configuration/pages/registration-pubcloud.adoc
mcalmer: Maybe @mbrookhuis wanted really to modify the image. But this means, you somehow need to "save" the image after you did the changes.

modules/client-configuration/pages/registration-pubcloud-script.adoc
```diff
@@ -0,0 +1,293 @@
...
+= Example Script for Registering Clients with SSH
+
+You can use the ``mgr-ssh-push-init`` script to register both traditional and Salt clients with SSH, or SSH tunnelling.
+Before you start, ensure that the SSH key of the {productname} Server is listed in the authorized-keys on the client.
+
+
+
+.Example: Script for Registering Clients with SSH
+----
+#!/bin/bash
```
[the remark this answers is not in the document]
Loquacity: I got the impression from the original post from Michael that this had been modified from the standard one. He says:

> The regular mgr-ssh-push-init script does stricthostchecking. I have attached a
> version where there is an extra option to disable this. Also here you
> have take care that the SSH key of the SUSE Manager Server is already
> present in the authorized-keys of the client. I have tested this method
> also at customers and it works.

modules/client-configuration/pages/registration-pubcloud.adoc
```diff
@@ -0,0 +1,60 @@
...
+If you have your clients running in a public cloud instance, there are several methods for registering them to your {productname} Server, in addition to the regular bootstrapping methods.
+For other registration methods, see xref:client-configuration:registration-overview.adoc[].
+
+By default, logging for both the {productname} Server and Salt clients is disabled.
+You can enable logging on Salt clients to provide some additional information during registration.
+Open or create the [path]``/etc/salt/minion.d/log_level.conf`` configuration file on the client, and add or edit this line:
+
+----
+log_level: info
+----
```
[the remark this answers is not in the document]
Loquacity: Updated the wording, but I think you're right and should be a general tshooting section for this, too.

modules/client-configuration/pages/registration-pubcloud.adoc
Loquacity: Yeah, I was confused about this too?

modules/client-configuration/pages/registration-pubcloud.adoc
Loquacity: Not that I could find.

modules/client-configuration/pages/registration-pubcloud.adoc
Loquacity: Added

modules/client-configuration/pages/registration-pubcloud-script.adoc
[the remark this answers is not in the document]
mcalmer: I made a "diff" between the script installed on our reference server and this in the PR.
100% identical.

@mbrookhuis we need your input here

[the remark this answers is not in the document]
mbrookhuis: Checked with @mcalmer. The patch is present in head uyuni, 4.0 and 4.1.

Note: this will only work if the SUSE Manager public key is in the ssh configuration of the salt-minion. The key should be present in /root/.ssh/know_hosts.

modules/client-configuration/pages/registration-pubcloud.adoc
mbrookhuis: There had been a discussion if it would be needed to include salt-minion in the image. When the image is deployed, a scirpt should be executed (cloud-init) which should build the correct /etc/salt/minion.d/susemanager.conf.  Problem is that this needs to be document for each provider and also I don't know how this works for each provider. If this is to difficult or not doable, mgr-ssh-push-init could be an option. 

I think we need to discuss this with public cloud team. I will ping Don Price. 

Note: for SALT you cannot use the --tunnel option. Also you cannot install salt-ssh. So only usable for normal salt-minion.

mbrookhuis (commented):
Added some comments

Also we need to have some further discussion with cloud team.

Loquacity:
Closing this PR as it's getting old. I'll move the issue to blocked in the backlog.

PR #581, opened 2020-09-30
Simplify translation workflow by using high-level po4a

paususe:
# Description

In the first iteration of AsciiDoc to gettext to enable translation of docs, 4 scripts did a lot of pre- and post-processing. Pushes to WebLate and merges back from WebLate were a fightening moment, and updating the gettext files generated a metric fuckton of gargabe in commits due to updated timestamps in headers.

This PR gets rid of most of that by using the high-level `po4a` tool instead of the low-level `po4a-gettextize`, `po4a-updatepo`, `po4a-translate` and others. Now one .pot/.po file is directly generated per documentation guide, and po4a takes care of generating translated AsciiDoc files form those big .po files.

No translations have been lost: I used `msgmerge` to bring the translated strings from the old .po files to the new .po files with the new .pot as base catalog.

Review advise: ignore most of it. The only interesting thing here is the high-level `po4a` tool uses configuration files, which are the files in `l10n-weblate/*.cfg`. List of files is currently maintained manually but in the near future it will be autogenerated from the navigators in the guides (same we do for the PDF), or automagically captured from the directories (see https://github.com/mquinson/po4a/issues/272).

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [ ] Manager-4.2
- [x] Uyuni

# Links

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -286,11 +286,11 @@ help: ## Prints a basic help menu about available targets
 
 .PHONY: pot
 pot:
-	$(shell $(current_dir)/make_pot.sh)
+	$(current_dir)/make_pot.sh
 
 .PHONY: translations
 translations:
-	$(shell $(current_dir)/use_po.sh)
+	$(current_dir)/use_po.sh
 
 .PHONY: clean
 clean: clean-en clean-es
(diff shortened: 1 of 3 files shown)


paususe:
Since this is blocking the docs components in WebLate, I am going to merge this right away.

PR #584, opened 2020-10-01
Add info about refreshing software repositories

keichwa:
https://github.com/uyuni-project/uyuni/issues/2645
https://github.com/SUSE/spacewalk/issues/12612

# Description

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [ ] Manager-4.2
- [x] Uyuni

# Links

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -6,6 +6,7 @@
 
 # Last packaged: 2020-09-15
 
+- Documented zypper autorefresh feature in Upgrade Guide
 - Update SP Migration chapter in Client Configuration Guide (https://github.com/uyuni-project/uyuni-docs/issues/483)
 - In Client Configuration and Upgrade Guide, add link to valid autoyast
   upgrade settings


keichwa:
@juliogonzalez or @mcalmer can you please confirm that we actually need a strategic `zypper ref` here?

mcalmer:
I would say "no". At least 'zypper patch' should do it automatically and I think 'zypper list-patches' should do it as well.
But a refresh happens only when the cache is declared as outdated. I do not know about the exact time. Maybe 5 Minutes, but could also be 24 hours.
Please ask you zypp guys about the rules.

juliogonzalez:
All I can tell is that I always manually run zypper ref before running zypper in, zypper up, etc... Just in case.

But could be because I am a long time Debian user, where we need to refresh the metadata :-)

So I agree with @mcalmer, check with zypp guys.

mlandres:
It depends on how  **autorefresh**  is configured for the repos!

- `zypper ref` checks and refreshes **all** repositories.
- `zypper CMD` checks and refreshes **autorefresh=1** repositories.

So if you know all enabled repos have autorefresh on, you don't need an explicit refresh. Otherwise you do (unless you know the repos do not contribute to the command).

mlandres:
> I do not know about the exact time. Maybe 5 Minutes, but could also be 24 hours.

The default _no-repo-refresh-again_ timeout is `/etc/zypp/zypp.conf:repo.refresh.delay = 10` min.
(Future zypp versions may allow a value per repo).

keichwa:
Michael Andres <notifications@github.com> writes:

> It depends on how autorefresh is configured for the repos!

Thanks for clarification.  I think the default is autorefresh enabled.
Thus by default we do not need 'zypper ref' here.  I'll add a trouble
shooting section.

-- 
Karl Eichwalder

keichwa:
I added a note and consider it ready for review.

modules/upgrade/pages/server-minor-upgrade-uyuni.adoc
```diff
@@ -43,6 +43,12 @@ spacewalk-service start
 +
 Reboot the server if a patch update recommends rebooting.
 
+[NOTE]
+====
+With default configuration setttings, all zypper commands run a repository refresh every 10 minutes (see ``repo.refresh.delay`` in [path]``/etc/zypp/zypp.conf``).
+If you disabled ``autorefresh``, only an explicit [command]``zypper ref`` command will refresh all repositories.
+====
+
 // We can remove  the version info when 4.1 / 2020.07 is released
```
Loquacity, line 52: these have been released?

modules/upgrade/pages/server-minor-upgrade-uyuni.adoc
```diff
@@ -43,6 +43,12 @@ spacewalk-service start
 +
 Reboot the server if a patch update recommends rebooting.
 
+[NOTE]
+====
+With default configuration setttings, all zypper commands run a repository refresh every 10 minutes (see ``repo.refresh.delay`` in [path]``/etc/zypp/zypp.conf``).
```
Loquacity, line 48:
```suggestion
By default, zypper refreshes the repository every ten minutes (see ``repo.refresh.delay`` in [path]``/etc/zypp/zypp.conf``).
```

modules/upgrade/pages/server-minor-upgrade-uyuni.adoc
```diff
@@ -43,6 +43,12 @@ spacewalk-service start
...
+If you disabled ``autorefresh``, only an explicit [command]``zypper ref`` command will refresh all repositories.
```
Loquacity, line 49:
```suggestion
If ``autorefresh`` is disabled, run [command]``zypper ref`` to refresh all repositories.
```

Loquacity (commented):
Just some suggestions to reduce the verbiage.

keichwa:

>  // We can remove  the version info when 4.1 / 2020.07 is released
>
> these have been released?

yes, but i want to keep it separate from this PR ;)

-- 
Karl Eichwalder

keichwa:
@Loquacity, @mcalmer, and @juliogonzalez , and now the question is whether we want to add this note everywhere where we have this or a similar procedure?

Loquacity:
> @Loquacity, @mcalmer, and @juliogonzalez , and now the question is whether we want to add this note everywhere where we have this or a similar procedure?

I think that could easily get out of hand. Let's save it for places where we know it's confusing.

keichwa:

> I think that could easily get out of hand. Let's save it for places
> where we know it's confusing.

Ok, agreed by my side.

-- 
Karl Eichwalder

mcalmer:
How many places do we have where we describe to call zypper cli directly?
I think this is true for ther server connected to SCC directly. All other systems are managed by the server and that procedure is not needed.

I think when you just find one more place, just duplicate it. If it is more, we might want to check if these descriptions are correct and needed at all.

keichwa:
> I think when you just find one more place, just duplicate it. If it is more, we might want to check if these descriptions are correct and needed at all.

You are right.  With grep, I find it just a 2nd time. I guess I'll duplicate it.

Loquacity:
It was the "or similar procedure" that made me think it would get out of hand :wink:

Loquacity (approved):

PR #594, opened 2020-10-07
Vale style checks prototype

jcayouette:
# Description

This PR adds **Vale** to our Github testing actions. Vale is a command-line tool that brings code-like linting to prose. It's cross-platform (Windows, macOS, and Linux), written in Go.

We have the option to use several different prose styles. Currently I have set it to Google. However we can also select more than one option. 

- Microsoft
An implementation of the Microsoft Writing Style Guide.
- Google
 An implementation of the Google Developer Documentation Style Guide.
- write-good
 An implementation of the guidelines enforced by the write-good linter.
- proselint
 An implementation of the guidelines enforced by the proselint linter.
- Joblint
 An implementation of the guidelines enforced by the Joblint linter.

vale-styles

I have more work to perfect this, however you can review the vale output by selecting:

Actions -> Your latest job with the action subtitled **Linting** -> Prose (From the left nav) -> Select the drop down titled **Run Vale for docs linting** -> Check your results.

The ultimate goal is to provide this output as an annotation to the PRs themselves:

github-pr-annotation

# Target branches

- [ ] Manager-4.0
- [X] Manager-4.1
- [X] Manager-4.2
- [X] Uyuni

# Links

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -1,5 +1,5 @@
 StylesPath = .github/styles
-MinAlertLevel = suggestion
+MinAlertLevel = error
 
 [*.adoc]
 BasedOnStyles = Google


Loquacity:
Love this, nice work! Let's just make it non-blocking for now, though, it's going to take a while to get the docs up to standard. Also, we need to make sure we have the docs team on board with the style choice, since it means we will be deviating fairly significantly from their standard practice (especially around headings).

.vale.ini
```diff
@@ -0,0 +1,6 @@
+StylesPath = /home/runner/work/uyuni-docs/uyuni-docs/styles/vale-styles
```
renner, line 1: I am not sure why we would need another subdirectory under `styles`? I would suggest to put all styles directly under `styles`, and move the `styles` directory to `.github` as [recommended in the vale action docs](https://github.com/errata-ai/vale-action#repository-structure).

.github/workflows/vale_tests.yml
```diff
@@ -0,0 +1,27 @@
...
+      run: gem install asciidoctor
+
+    - name: Set up Homebrew
+      id: set-up-homebrew
+      uses: Homebrew/actions/setup-homebrew@master
+
+    - name: Run vale for docs linting
+      run: |
+        brew install vale
+        vale --debug modules/
```
renner, line 26: Instead of using the available [vale action](https://github.com/errata-ai/vale-action#usage), you are installing `vale` via `homebrew` and calling it manually. What is the problem with using the vale action?

renner (commented):
In general I think it might be preferable to use the upstream (e.g. Google) styles from the available zip file releases as the vale action does. I would try to not even commit those to this repo, and in case you want to create and maintain your own styles to put those separately. This file further seems to be empty and should probably be removed: `styles/vale-styles/Google/vocab.txt `.

.github/workflows/feature_tests_uyuni.yml
```diff
@@ -49,12 +49,14 @@ jobs:
...
     - name: '[BUILD] SUMA Docs with Lunr/Antora Custom Docker Image'
       run: |
-        docker run -u $UID --privileged -e DOCSEARCH_ENABLED=true -e DOCSEARCH_ENGINE=lunr -v $GITHUB_WORKSPACE:/antora/ --rm -t lunr/antora:custom uyuni-site.yml --generator antora-site-generator-lunr
+        docker run -u $UID --privileged -e DOCSEARCH_ENABLED=true -e DOCSEARCH_ENGINE=lunr -v $GITHUB_WORKSPACE:/antora/ --rm -t lunr/antora:custom suma-site.yml --generator antora-site-generator-lunr
         #cd build/
         ls -a
     - name: '[BUILD] SUMA PDF Documents with Asciidoctor-pdf'
       run: make pdf-all-uyuni
+
+
```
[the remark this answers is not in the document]
jcayouette: This was a fix for a simple error with our makefile. If we select uyuni here our xrefs are not checked. This small change fixes this error.

.github/workflows/vale_tests.yml
jcayouette: I tested the action several ways. The action was not picking up styles and therefor throwing a zero results as nothing was being compared "as we covered yesterday". I checked the Homebrew repo as an example and tried using their method. This way works though not as the action intends. 

Of course I agree with you. We should swap back to the default setup. Just wrapping my head around whats going on.

.vale.ini
jcayouette: This was a test as styles were not being picked up by vale.

keichwa (approved):
I'd run this only on request and on the main branch only.  Just to save resources.

.github/workflows/vale_tests.yml
```diff
@@ -0,0 +1,27 @@
+name: Linting
+on: [push]
```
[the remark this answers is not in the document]
jcayouette: Vale is only testing *adoc files as per the test configs instructions: https://gist.githubusercontent.com/jcayouette/90d9b8a40e8d4e61945694bc9c29db04/raw/ea449091b882394065b622493ece0b3fa12f7752/.vale.ini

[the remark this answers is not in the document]
jcayouette: We actually do want this running on every PR. Once we have fixed "current" issues then we want the vale action running on new content especially prior to review.

.github/workflows/vale_tests.yml
jcayouette: I am now using the default methods.

Loquacity:
Ah, i like the suggestions! Although someone is going to have to do a "will" search and replace at some point :wink:

Loquacity (approved):

Loquacity:
https://github.com/SUSE/spacewalk/issues/12488

PR #616, opened 2020-10-18
Add script and GH action to check for unused files

paususe:
# ***DO NOT MERGE: DOCS REPO IS FROZEN***

# Description

From time to time, we find some AsciiDoc file is not referenced anywhere. Sometimes it's old files, sometimes it's we forgot to link it.

This PR adds a script and GH action to check every file on disk is actually used in either the navigation bar, or some other file inside the guide. If unused files are found, they are reported per guide, and the GH action exists with an error.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [ ] Manager-4.2
- [x] Uyuni

# Links

### DIFF
--- a/.github/workflows/find_unused.yml
+++ b/.github/workflows/find_unused.yml
@@ -1,0 +1,13 @@
+name: Find unused files
+on: push
+
+jobs:
+ run:
+   name: Find unused files
+   runs-on: ubuntu-latest
+   steps:
+   - name: Checkout repo
+     uses: actions/checkout@v2
+
+   - name: Look for unused files
+     run: ./find_unused administration client-configuration installation large-deployments quickstart-public-cloud quickstart-sap reference retail salt upgrade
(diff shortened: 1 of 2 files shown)


.github/workflows/find_unused.yml
```diff
@@ -0,0 +1,13 @@
+name: Find unused files
+on: push
```
Loquacity, line 2: This might not need to be done on every push, since there are unlikely to be a lot of these picked up. Can we make it less frequent?

paususe: No, it should be on every push, to make sure we never have any stale file.

@juliogonzalez Do we have a limit on GH actions minutes?

juliogonzalez: @paususe https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration

Loquacity: I think we need to seriously consider the issue of GH actions minutes, especially since we are also implementing l10n and linting actions. I would prefer to spend those minutes on actions that will have a higher positive rate.

Loquacity:
Please don't merge PRs that are marked as `do not merge`. The docs repo will be unfrozen as soon as we have an accepted SR.

PR #625, opened 2020-10-21
Document CentOS7/8 ppc64le for Uyuni

juliogonzalez:
# Description

Document CentOS7/8 ppc64le for Uyuni

SUSE Manager does not support this for now.

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [ ] Manager-4.2
- [x] Uyuni

# Links

Related PR for the code repository: https://github.com/uyuni-project/uyuni/pull/2738
Fixes https://github.com/SUSE/spacewalk/issues/12507

### DIFF
--- a/modules/client-configuration/pages/clients-centos.adoc
+++ b/modules/client-configuration/pages/clients-centos.adoc
@@ -53,8 +53,12 @@ The channels you need for this procedure are:
 | {centos} 8 | centos8      | centos8-uyuni-client | centos8-appstream
 |===
 
+ifeval::[{uyuni-content} == true]
+include::snippets/add_channels_cli_multiarch.adoc[]
+endif::[]
+ifeval::[{suma-content} == true]
 include::snippets/add_channels_cli.adoc[]
-
+endif::[]
 
 [NOTE]
 ====
--- a/modules/client-configuration/pages/snippets/add_channels_cli_multiarch.adoc
+++ b/modules/client-configuration/pages/snippets/add_channels_cli_multiarch.adoc
@@ -1,0 +1,17 @@
+.Procedure: Adding Software Channels at the Command Prompt
+. At the command prompt on the {productname} Server, as root, use the [command]``spacewalk-common-channels`` command to add the appropriate channels:
++
+----
+spacewalk-common-channels \
+-a <architecture>
+<base_channel_name> \
+<child_channel_name_1> \
+<child_channel_name_2> \
+... <child_channel_name_n>
+----
+. Synchronize the channels:
++
+----
+mgr-sync refresh --refresh-channels
+----
+
--- a/modules/client-configuration/pages/supported-features.adoc
+++ b/modules/client-configuration/pages/supported-features.adoc
@@ -111,12 +111,22 @@ The operating system you run on a client is supported by the organization that s
 | {check}
 
 | {centos} 8
+ifeval::[{uyuni-content} == true]
+| {x86_64}, {ppc}
+endif::[]
+ifeval::[{suma-content} == true]
 | {x86_64}
+endif::[]
 | {cross}
 | {check}
 
 | {centos} 7
+ifeval::[{uyuni-content} == true]
+| {x86_64}, {ppc}
+endif::[]
+ifeval::[{suma-content} == true]
 | {x86_64}
+endif::[]
 | {check}
 | {check}
 


modules/client-configuration/pages/snippets/add_channels_cli_multiarch.adoc
```diff
@@ -0,0 +1,17 @@
+.Procedure: Adding Software Channels at the Command Prompt
```
juliogonzalez, line 1: This snipped should be used for any other OS added with `spacewalk-common-channels` that support several architectures, but let's not do that change now, as I am not sure of the list, and could be that we even need to fix some stuff before that happens :-)

juliogonzalez:
PDFs:
[suse_manager_client-configuration_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/5417091/suse_manager_client-configuration_guide.pdf)
[uyuni_client-configuration_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/5417092/uyuni_client-configuration_guide.pdf)

modules/client-configuration/pages/supported-features.adoc
```diff
@@ -111,12 +111,22 @@ The operating system you run on a client is supported by the organization that s
 | {check}
 
 | {centos} 8
+ifeval::[{uyuni-content} == true]
+| {x86_64}, {ppc}
```
juliogonzalez, line 115: I am *not* convinced at all about how {ppc} gets replaced. It gets replaced with `POWER` but what does that mean?

First of all, I guess we mean PowerPC and not Power, but even so: ppc32? ppc64? ppc64le?

In the end from the whole Power family, we only support ppc64le (PowerPC 64 bits little endian) for SLE (with the exception of SLE that I think it still supports ppc -32bit- and ppc64 -big endian-)

FMPOV this is specially important because at the call to `spacewalk-common-channels` you cannot use `POWER`.

@paususe @mcalmer any thoughts?

juliogonzalez: Just for the record, I see the problem with the mention to `ARM` for SLE:

image

juliogonzalez:
Why did I get a lot of notifications of github actions on my forked repo, but at the same time this PR didn't run any :-?

modules/client-configuration/pages/supported-features.adoc
paususe: We have {ppc} and {ppc64le} and both get a replacement that's only useful for "marketing" materials but not for command-line tools or technical information: "POWER" and "Power PC" respectively. 

We should check where each one of those is used and use either "POWER" (marketing, e. g. the table above), "ppc" (technical details and CLI tools for SLE 10), "ppc64" (technical details and CLI tools for SLE 11) or "ppc64le" (technical details and CLI tools SLE 12+).

paususe:
> Why did I get a lot of notifications of github actions on my forked repo, but at the same time this PR didn't run any :-?

GH are configured to run on merge to master. Maybe it's that?

paususe (approved):
LGTM

modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -53,8 +53,12 @@ The channels you need for this procedure are:
 | {centos} 8 | centos8      | centos8-uyuni-client | centos8-appstream
 |===
 
+ifeval::[{uyuni-content} == true]
+include::snippets/add_channels_cli_multiarch.adoc[]
+endif::[]
+ifeval::[{suma-content} == true]
 include::snippets/add_channels_cli.adoc[]
-
+endif::[]
```
Loquacity, line 61: I'm pretty certain you need blank lines around the ifs

modules/client-configuration/pages/supported-features.adoc
Loquacity: We have these:

```
    ppc: POWER
    ppc64le: 'IBM Power PC'
```

I agree with Pau, we need to expand the list. If you don't want to do it, just create an issue and I'll do it for you.

modules/client-configuration/pages/supported-features.adoc
```diff
@@ -111,12 +111,22 @@ The operating system you run on a client is supported by the organization that s
...
 | {cross}
 | {check}
 
 | {centos} 7
+ifeval::[{uyuni-content} == true]
+| {x86_64}, {ppc}
+endif::[]
+ifeval::[{suma-content} == true]
 | {x86_64}
+endif::[]
```
Loquacity, line 129: Did you check that this builds OK?

modules/client-configuration/pages/supported-features.adoc
mcalmer: I think pure PPC is only available for SLE11. Since SLE12 we only support ppc64le.

mcalmer (approved):
Just check how ppc is described but the rest looks good as far as I could see it from the "sourcecode" :-)

modules/client-configuration/pages/supported-features.adoc
paususe: SLE11 is ppc64
SLE10 is ppc

modules/client-configuration/pages/supported-features.adoc
juliogonzalez: See the attached PDFs. They look OK.

modules/client-configuration/pages/supported-features.adoc
juliogonzalez: Ah, so ppc from https://build.suse.de/project/show/SUSE:SLE-11-SP4:Update is not used? I see both there, ppc and ppc64

juliogonzalez: > I agree with Pau, we need to expand the list. If you don't want to do it, just create an issue and I'll do it for you.

If I understand we now want to have:

>     ppc: 'IBM Power PC'
>     ppc64: ppc64
>     ppc64le: ppc64le

The first one to be used for "Marketing", the second one to be used for technical things such as the matrix.

Is that right?

paususe: For marketing purposes:
* "IBM Power PC" (or maybe even shorter: "Power PC") should only be used for SLE 10 SP3, which we no longer support (but client tools are still available from SCC)
* "IBM POWER" for SLE11+, CentOS/RHEL/OL 6+

For technical details:
* SLE 10 SP3: ppc
* SLE 11 SP4, CentOS/RHEL/OL 6: ppc64
* SLE 12+, CentOS/RHEL/OL 7+: ppc64le

juliogonzalez: > For marketing purposes:
> 
> - "IBM Power PC" (or maybe even shorter: "Power PC") should only be used for SLE 10 SP3, which we no longer support (but client tools are still available from SCC)

Yeah, but SLE10 is not documented anywhere anymore, so no need to worry about it.

> - "IBM POWER" for SLE11+, CentOS/RHEL/OL 6+

OK. Just remember we will NOT add ppc64 for CentOS/RHEL/OL6. EoL is almost here. Only 7 and 8 was requested.

But I see a problem with what you said before: you see the matrix as marketing, I see it as technical. I'd really want to use the technical names on the matrix, so the users an use that anywhere else, whey adding repos, syncing, whatever. If we do not do that, then that means that we need to *duplicate* the information of the architectures. Either that or we have footnotes on the matrix explaining what each "marketing name" means.

Which brings me to the next question: we mention on the matrix Itanium, not ia64, IBM Z, and not s390x... and so on. Shouldn't we do the same?

I think this really calls for a separate card, yes.

> For technical details:
> 
> - SLE 10 SP3: ppc

Not documented, so not really needed.

> - SLE 11 SP4, CentOS/RHEL/OL 6: ppc64

OK. Just as said above: no ppc64 on CentOS/RHEL/OL6

> - SLE 12+, CentOS/RHEL/OL 7+: ppc64le

OK.

juliogonzalez: Let's get this PR merged, as the discussion is affecting other architectures as well.

I will create a card.

modules/client-configuration/pages/supported-features.adoc
juliogonzalez: Since the PDFs were able to build, I will merge.

modules/client-configuration/pages/supported-features.adoc
juliogonzalez: https://github.com/SUSE/spacewalk/issues/12836

PR #633, opened 2020-10-26
ppc entities

Loquacity:
# Description

Update the PPC entites suite-wide

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [x] Manager-4.2
- [x] Uyuni

# Links

Fixes https://github.com/SUSE/spacewalk/issues/12836

### DIFF
--- a/modules/client-configuration/pages/supported-features.adoc
+++ b/modules/client-configuration/pages/supported-features.adoc
@@ -51,11 +51,16 @@ The operating system you run on a client is supported by the organization that s
 | {check}
 
 | {sle} 11
-| {x86}, {x86_64}, {ipf}, {ppc64le}, {ibmz}
+| {x86}, {x86_64}, {ipf}, {ppc64}, {ibmz}
 | {check}
 | {check}
 
-| {sles} for SAP
+| {sles} for SAP 15
+| {x86_64}, {ppc64le}
+| {check}
+| {check}
+
+| {sles} for SAP 12
 | {x86_64}, {ppc64le}
 | {check}
 | {check}


modules/client-configuration/pages/supported-features.adoc
```diff
@@ -41,22 +41,22 @@ The operating system you run on a client is supported by the organization that s
...
 
 | {sle} 12
-| {x86_64}, {ppc}, {ibmz}, ARM
+| {x86_64}, {ppc64le}, {ibmz}, ARM
 | {check}
 | {check}
 
 | {sle} 11
-| {x86}, {x86_64}, {ipf}, IBM {ppc}, {ibmz}
+| {x86}, {x86_64}, {ipf}, IBM {ppc64le}, {ibmz}
```
keichwa, line 54:
```suggestion
| {x86}, {x86_64}, {ipf}, {ppc64le}, {ibmz}
```

keichwa (approved):

branding/pdf/entities.adoc
```diff
@@ -2,8 +2,9 @@
 :ibmz: IBM Z
 :zsystems: {ibmz}
 :zseries: {ibmz}
-:ppc: POWER
-:ppc64le: IBM Power PC
+:ppc: IBM Power
```
paususe, line 5: No, this is wrong. You need to check whether we are talking about ppc/ppc64/ppc64le in a technical way (the same way we say "you need 8 GB of RAM, 200 GB of disk space", i. e. precise details) or in a more "marketing" way (the same way we say "you need an Intel-compatible server, you need a network connection", i. e. more generic way).

If the former, then we must use ppc, ppc64 or ppc64le.

If the latter, then we should use "IBM POWER".

That's why I was saying we need to review accurately where were we using the placeholders and for what, we cannot just replace the entities and the tables.

For this case:

```suggestion
:ppc: ppc
```

modules/client-configuration/pages/supported-features.adoc
paususe, line 54: SLE11 is ppc64 (not ppc64le). You can verify for your self what tools are supported for each client by looking in SCC or the sub-projects of https://build.opensuse.org/project/subprojects/systemsmanagement:Uyuni:Stable

```suggestion
| {x86}, {x86_64}, {ipf}, {ppc64}, {ibmz}
```

modules/client-configuration/pages/supported-features.adoc
```diff
@@ -41,22 +41,22 @@ The operating system you run on a client is supported by the organization that s
...
 | {check}
 | {check}
 
 | {sles} for SAP
-| {x86_64}, {ppc}
+| {x86_64}, {ppc64le}
```
paususe, line 59: This does not make any like without a version. It's not "SLES for SAP" but "SLES for SAP 12" and "SLES for SAP 15". And it's the same as for the normal SLES. It probably makes sense to duplicate this column and add the version number.

modules/installation/pages/client-requirements.adoc
```diff
@@ -26,17 +26,17 @@ For non-{suse} operating systems, including {rhel}, {centos}, and {oracle}, only
...
 
 | {sle} 12
-| {x86_64}, {ppc}, {ibmz}, ARM
+| {x86_64}, {ppc64le}, {ibmz}, ARM
 | {check}
 | {check}
 
 | {sle} 11
-| {x86}, {x86_64}, IBM {ppc}, {ibmz}
+| {x86}, {x86_64}, {ppc}, {ibmz}
```
paususe, line 39: Wrong, it's ppc64

```suggestion
| {x86}, {x86_64}, {ppc64}, {ibmz}
```

modules/installation/pages/client-requirements.adoc
```diff
@@ -56,7 +56,7 @@ For non-{suse} operating systems, including {rhel}, {centos}, and {oracle}, only
 | {check}
 
 | {sles} for SAP
-| {x86_64}, {ppc}
+| {x86_64}, {ppc64le}
```
paususe, line 59: Same comment as above: it does not make any sense without the version of SLES for SAP (12 or 15).

suma-site.yml
```diff
@@ -40,8 +40,9 @@ asciidoc:
     zseries: IBM Z
     zsystems: IBM Z
     ibmz: IBM Z
-    ppc: POWER
-    ppc64le: 'IBM Power PC'
+    ppc: 'IBM Power'
```
paususe, line 43:
```suggestion
    ppc: 'ppc'
```

uyuni-site.yml
```diff
@@ -39,8 +39,9 @@ asciidoc:
     zseries: IBM Z
     zsystems: IBM Z
     ibmz: IBM Z
-    ppc: POWER
-    ppc64le: 'IBM Power PC'
+    ppc: 'IBM Power'
```
paususe, line 42:
```suggestion
    ppc: 'ppc'
```

paususe (changes requested):

branding/pdf/entities.adoc
Loquacity: There's no point using an entity for this, though. If we want to write "ppc" we can just write it.

modules/client-configuration/pages/supported-features.adoc
Loquacity: Yeah, I wondered about that. I'll duplicate it.

branding/pdf/entities.adoc
paususe: OK

paususe (approved):

PR #655, opened 2020-11-05
disable SAP in 4.1.3

Loquacity:
# Description

Comments out the SAP Guide from the various makefiles, etc, and moves the book files to the unused modules directory, so it is not indexed.

@jcayouette When you are happy with this PR, please merge and rebuild the 4.1.3 package for @juliogonzalez (Don't forget to remove the SAP Guide from the changelog when you do that).

# Target branches

- [ ] Manager-4.0
- [x] Manager-4.1
- [ ] Manager-4.2
- [ ] Uyuni

# Links

Partial https://github.com/SUSE/spacewalk/issues/13001


jcayouette (approved):
LGTM, thanks.

PR #667, opened 2020-11-12
use sentence style instead of title style

keichwa:
# Description

# Target branches

- [ ] Manager-4.0
- [ ] Manager-4.1
- [ ] Manager-4.2
- [ ] Uyuni

# Links

### DIFF
--- a/modules/client-configuration/pages/activation-keys.adoc
+++ b/modules/client-configuration/pages/activation-keys.adoc
@@ -12,7 +12,7 @@ When the bootstrap script is executed all configuration settings associated with
 An activation key can specify:
 
 * Channel assignment
-* System types (traditionally called add-on entitlements)
+* System types or add-on entitlements
 * Contact method
 * Configuration files
 * Packages to be installed


modules/client-configuration/pages/activation-keys.adoc
```diff
@@ -11,12 +11,12 @@ When the bootstrap script is executed all configuration settings associated with
...
 An activation key can specify:
 
-* Channel Assignment
-* System Types (Traditionally called Add-on Entitlements)
-* Contact Method
-* Configuration Files
-* Packages to be Installed
-* System Group Assignment
+* Channel assignment
+* System types (traditionally called add-on entitlements)
```
Loquacity, line 15: Do you think that using "traditionally" here might be confusing? Perhaps we could do something like this instead:

```suggestion
* System types (or add-on entitlements)
```
```suggestion
* System types (traditionally called add-on entitlements)
```

Loquacity (approved):

modules/client-configuration/pages/activation-keys.adoc
keichwa: yes, your proposal is better.

keichwa:
And now I'm wondering whether I'm actually allowed to merge this.  I'm asking because of the translations.

Loquacity:
> And now I'm wondering whether I'm actually allowed to merge this. I'm asking because of the translations.

This is minor, go ahead and merge.

PR #693, opened 2020-11-23
Build errors

Loquacity:
# Description

Some things I noticed while attempting to debug the build errors. Seems as though the problem lies in the Spanish version of the Client Cfg Guide, it's not building the nav properly. The PDF nav that is created is an empty file, and the HTML nav is non-existent:

image

This is the actual error when I try to build:

image

I also noticed this when the Spanish build begins:

image

(Although the Install Guide is built first, and seems to be working OK, so might not be relevant).

# Target branches

- [x] Master (Default) - Spanish only, English is building just fine :+1: 
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


paususe (approved):
LGTM

Please, always build before committing and check the result of the GH Action before merging! The build was broken even for English due to "snippets:" instead of "snippets/". The GH Action showed this.

jcayouette (approved):

jcayouette:
Merging so we can get master working!

PR #697, opened 2020-11-23
Update cobbler dhcp to include EFI PXE example

aaannz:
~**_DO NOT MERGE DOCS REPO IS FROZEN_**~

# Description

- update cobbler dhcp example to include EFI PXE configuration
- add example using dhcpd formula

# Target branches

- [X] Master (Default)
- [X] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/12542

### DIFF
--- a/modules/client-configuration/pages/cobbler.adoc
+++ b/modules/client-configuration/pages/cobbler.adoc
@@ -57,12 +57,12 @@ All configurable settings are explained in detail in the [path]``/etc/cobbler/se
 
 The PXE boot process uses DHCP to find the TFTP boot server.
 The {productname} Server can act as such a  TFTP boot server and Cobbler can generate the content for it.
-DHCP server can be configured directly or using DHCPd formula.
+The DHCP server can be configured directly, or you can use the DHCPd formula.
 
 To configure DHCP directly, you must have administrative access to the network's DHCP server.
 Edit the DHCP configuration file so that it points to the {productname} Server as the TFTP boot server:
 
-.Procedure: Example for Configuring the ISC DHCP Server
+.Procedure: Configuring the ISC DHCP Server Directly
 . On the DHCP server, as root, open the [path]``/etc/dhcpd.conf`` file.
 . Append a new class with options for performing PXE boot installation.
     For example:
@@ -93,7 +93,7 @@ This example:
 +
 . Save the file.
 
-To configure DHCP using DHCPd formula, follow this example:
+Alternatively, you can configure DHCP using the DHCPd formula instead:
 
 .Procedure: Example for configuring ISC DHCP Server using DHCPd formula
 . Configure the DHCPd formula.


modules/client-configuration/pages/cobbler.adoc
```diff
@@ -82,10 +88,17 @@ This example:
 * Creates a class called ``PXE``.
 * Identifies systems as ``PXEClient`` if they are configured with PXE as the first boot priority.
 * Directs PXEClients to the Cobbler server at ``192.168.2.1``.
-* Retrieves the [path]``pxelinux.0``bootloader file.
+* Retrieves the [path]``grub/grub.efi`` bootloader file for EFI PXE clients.
+* Retrieves the [path]``pxelinux.0`` bootloader file for legacy BIOS PXE clients.
 +
 . Save the file.
 
+.Procedure: Example for configuring ISC DHCP Server using DHCPd formula
```
Loquacity, line 96:
```suggestion
.Procedure: Configuring an ISC DHCP Server with the DHCPd formula
```

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -82,10 +88,17 @@ This example:
...
+. Configure DHCPd formula as described in xref:salt:formula-dhcpd.adoc[]
```
Loquacity, line 97:
```suggestion
. Configure the DHCPd formula.
  For more information, see xref:salt:formula-dhcpd.adoc[]
```

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -82,10 +88,17 @@ This example:
...
+. Use [path]``grub/grub.efi`` in [guimenu]``Filename EFI`` field for EFI PXE support
+. Use [path]``pxelinux.0`` in [guimenu]``Filename`` field for legacy BIOS support
```
Loquacity, line 99:
```suggestion
. In the [guimenu]``Filename EFI`` field, for EFI PXE support type [path]``grub/grub.efi``.
  Alternatively, for legacy BIOS support type [path]``pxelinux.0``.
```

Loquacity (changes requested):
Edits for style and consistency.

keichwa:
Do we actually want to write "DHCPd"?  I'd rather prefer all caps or just "dhcpd".  IIRC, the form editor or the web ui has the habit to just capitalize the word.  That's why we have "Ftp" instead of FTP.  Might be different now...

modules/client-configuration/pages/cobbler.adoc
aaannz: This is not exactly what I meant. I mean that user should put both values there to support both legacy and efi PXE. So that this example will result in same behaviour as the manual example above.

aaannz:
> Do we actually want to write "DHCPd"? I'd rather prefer all caps or just "dhcpd". IIRC, the form editor or the web ui has the habit to just capitalize the word. That's why we have "Ftp" instead of FTP. Might be different now...

I don't really care, I just wanted to be consistent with what we already have in formulas. I can switch it to lowercase everywhere (and Dhcpd if beginning of sentence).

modules/client-configuration/pages/cobbler.adoc
aaannz: I am not sure if this will not confuse people. There are two examples - manual configuring and this one, using dhcpd formula - and its up to user which one, but only one, they will use.

Loquacity: In that case, we should probably include some explanation to that effect in a para before the procedure.

modules/client-configuration/pages/cobbler.adoc
Loquacity: Ah, I understand, let's go back to two steps then, assuming they are two different fields. How about this:

```suggestion
. In the  [guimenu]``Filename EFI`` field, type [path]``grub/grub.efi`` to enable EFI PXE support.
. In the  [guimenu]``Filename`` field, type [path]``pxelinux.0`` to enable legacy BIOS support.
```

Loquacity:
> Do we actually want to write "DHCPd"? I'd rather prefer all caps or just "dhcpd". IIRC, the form editor or the web ui has the habit to just capitalize the word. That's why we have "Ftp" instead of FTP. Might be different now...

I think we use DHCPd throughout, I was taught that when 'd' was for 'daemon' it should be lowercase, but the times might have moved on. The WebUI has horrible capitalization generally :joy:

Loquacity:
> > Do we actually want to write "DHCPd"? I'd rather prefer all caps or just "dhcpd". IIRC, the form editor or the web ui has the habit to just capitalize the word. That's why we have "Ftp" instead of FTP. Might be different now...
> 
> I don't really care, I just wanted to be consistent with what we already have in formulas. I can switch it to lowercase everywhere (and Dhcpd if beginning of sentence).

I think we use DHCPd throughout, I was taught that when 'd' was for 'daemon' it should be lowercase, but the times might have moved on. The WebUI has horrible capitalization generally :joy:

keichwa:
> I think we use DHCPd throughout, I was taught that when 'd' was for 'daemon' it should be lowercase, but the times might have moved on. 

Yes, I see.  In this case, if we want to talk about the daemon in general, we could spell it out as you are indicating it: DHCP daemon.  If we talk about the binary, it is probably
```
[command]``dhcpd``
```
Package or form name could also be different.

Loquacity:
> > I think we use DHCPd throughout, I was taught that when 'd' was for 'daemon' it should be lowercase, but the times might have moved on.
> 
> Yes, I see. In this case, if we want to talk about the daemon in general, we could spell it out as you are indicating it: DHCP daemon. If we talk about the binary, it is probably
> 
> ```
> [command]``dhcpd``
> ```
> 
> Package or form name could also be different.

OK, perhaps we should address this in a new issue. I think we probably need to apply the same thinking to other formulas, too, if we're going to change it.

keichwa (approved):
Just cosmetic suggestions.

modules/client-configuration/pages/cobbler.adoc
aaannz: I added some explaining text above.

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -57,10 +57,10 @@ All configurable settings are explained in detail in the [path]``/etc/cobbler/se
 
 The PXE boot process uses DHCP to find the TFTP boot server.
 The {productname} Server can act as such a  TFTP boot server and Cobbler can generate the content for it.
-You must have administrative access to the network's DHCP server.
-Edit the DHCP configuration file so that it points to the {productname} Server as the TFTP boot server:
-
+DHCP server can be configured directly or using DHCPd formula.
```
Loquacity, line 60:
```suggestion
The DHCP server can be configured directly, or you can use the DHCPd formula.
```

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -57,10 +57,10 @@ All configurable settings are explained in detail in the [path]``/etc/cobbler/se
...
 
+To configure DHCP directly, you must have administrative access to the network's DHCP server.
+Edit the DHCP configuration file so that it points to the {productname} Server as the TFTP boot server:
 
 .Procedure: Example for Configuring the ISC DHCP Server
```
Loquacity, line 65:
```suggestion
.Procedure: Configuring the ISC DHCP Server Directly
```

Just to match it back to the language we use above.

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -82,10 +88,20 @@ This example:
 * Creates a class called ``PXE``.
 * Identifies systems as ``PXEClient`` if they are configured with PXE as the first boot priority.
 * Directs PXEClients to the Cobbler server at ``192.168.2.1``.
-* Retrieves the [path]``pxelinux.0``bootloader file.
+* Retrieves the [path]``grub/grub.efi`` bootloader file for EFI PXE clients.
+* Retrieves the [path]``pxelinux.0`` bootloader file for legacy BIOS PXE clients.
 +
 . Save the file.
 
+To configure DHCP using DHCPd formula, follow this example:
```
Loquacity, line 96:
```suggestion
Alternatively, you can configure DHCP using the DHCPd formula instead:
```

Loquacity (changes requested):

Loquacity (approved):

PR #701, opened 2020-11-24
Fix build of translations

paususe:
# Description

Fix build of translations: since some files moved around, indices need to be updated. Why were they not updated by the GH actions is a mystery, I will take  look later.

Also, make updating the indices part of the "make pot" target.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


Loquacity:
I checked out this branch to take a look and can confirm that all languages appear to be building OK. Some screenshots:

image
image
image
image

paususe:
> Just one thing I noticed was that the make command is exiting with this error:
> 
> `/bin/sh: warning: setlocale: LC_ALL: cannot change locale (cs)`

That might be because you don't have Czech installed on your system, so that locale is not found and cannot be used. What does the date field on the front cover show in Czech? Could you please paste a screenshot?

jcayouette:
Very nice, thanks @paususe !

jcayouette:
@paususe My only concern was the default index should be **index.adoc**. **index-uyuni.adoc* *should be depreciated and therefore removed. "my bad for missing this one." We use conditionals to specifiy the product in the **index.adoc** now.

Loquacity:
> > Just one thing I noticed was that the make command is exiting with this error:
> > `/bin/sh: warning: setlocale: LC_ALL: cannot change locale (cs)`
> 
> That might be because you don't have Czech installed on your system, so that locale is not found and cannot be used. What does the date field on the front cover show in Czech? Could you please paste a screenshot?

image

paususe:
> > > Just one thing I noticed was that the make command is exiting with this error:
> > > `/bin/sh: warning: setlocale: LC_ALL: cannot change locale (cs)`
> > 
> > 
> > That might be because you don't have Czech installed on your system, so that locale is not found and cannot be used. What does the date field on the front cover show in Czech? Could you please paste a screenshot?
> 
> image

It seems to work fine. No idea about the warning.

paususe:
Removed index-uyuni.adoc and regenerated gettext catalogs and po4a configuration files.

paususe:
Unbreak build

Issue #708, opened 2020-11-26
Document ISS troubleshooting [WAS: Integrate TID 000019792]


moio:
## Documentation Update Required

Integrate information in https://www.suse.com/support/kb/doc/?id=000019792

## What does the documentation say now?

Nothing regarding a typical ISS troubleshooting step.

## What change is required?

A general note that, in case of ISS errors, both the ISS-Master-side and the ISS-Slave-side caches can be cleared and that can sometimes resolve the problem (in case we had a bug that resulted in cache corruption, and we later fixed it, but caches remained corrupd).

## Subject Matter Expert (SME)

@moio

## Version

Which version of the software is this feature/update expected to land in?

**Already out in all versions.**

Which version of the documentation must this feature/update land in?

**Ideally in all versions (in addition to the TID above).**

Does the content need to be backported to previous documentation versions?

**yes**

Loquacity:
ISS is documented in the Admin Guide, so let's add this to the Admin Troubleshooting section.

PR #716, opened 2020-12-02
ISS tshooting

Loquacity:
# Description

Adds a new troubleshooting section for ISS, based on https://www.suse.com/support/kb/doc/?id=000019792

# Target branches

- [x] Master (Default)
- [x] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/708


modules/administration/pages/tshoot-iss.adoc
```diff
@@ -0,0 +1,56 @@
...
+.Procedure: Resolving ISS Caching Errors
+. On the ISS Master, at the command prompt, as root, delete the cache file for the Master:
++
+----
+rm -rf /var/cache/rhn/xml-*
+----
+. Restart the service:
++
+----
+capache2 restart
```
keichwa, line 45:
```suggestion
rcapache2 restart
```
[the typo is also in the kb ;) ]

keichwa (changes requested):

modules/administration/pages/tshoot-iss.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+////
+Cause: Caches on the ISS slave and master can contain old or out-of-date entries.
+Consequence: Synchronizing fails, with ``ERROR: Encountered IntegrityError``.
+Fix: Delete the cache on the slave and the master
+Result: Synchronization completes successfully.
+////
+
+Caches on the ISS Master and Slave can contain old or out-of-date entries.
+This leads to synchronization failing with ``ERROR: Encountered IntegrityError: null value in column "installer_updates" violates not-null constraint``.
```
moio, line 30: Hmm, I would suggest to put this in a slightly different angle.

Story is: in the particular case of ISS, we make extensive use of caches, and those are particularly prone to be poisoned by bugs. That means that if version X has a bug, it might create invalid cache files - no surprise. What might be surprising is that even if user updates to version X+1 which has the bug fixed, the same symptoms might still show up. The reason is that cache files are not necessarily re-created every time, so "poisoned" caches from version X may well be still consumed in version X+1, resulting in the same symptoms even if the root cause had been fixed.

So the way to address this properly is not only to upgrade to X+1, but also to clear all caches so that we're sure to start from clean data.

The error message `null value in column "installer_updates" violates not-null constraint` is just the latest symptom of the latest bug of this kind. I would not focus on this message other than being an example at most. 

The following two lines are way more important, as they point directly to this (meta)problem:

```
consider removing satellite-sync cache at /var/cache/rhn/satsync/* and re-run satellite-sync with same options.
If this error persits after removing cache, please contact SUSE support.
```

moio (changes requested):

keichwa (approved):
with moio's background info for clarity, if possible.

jcayouette (approved):

modules/administration/pages/tshoot-iss.adoc
Loquacity: Updated.

keichwa (approved):

PR #719, opened 2020-12-03
Mcalmer patch 1

mcalmer:
# Description

For gitfs the option `top_file_merging_strategy` is usefull/required. Add it to the docs

# Target branches

- [x] Master (Default)
- [x] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/modules/salt/pages/salt-gitfs.adoc
+++ b/modules/salt/pages/salt-gitfs.adoc
@@ -48,6 +48,17 @@ gitfs_remotes:
     - base: master
 ----
 
+top_file_merging_strategy:
+Define the strategy to merge multiple [literal]``top.sls`` files defined in your different fileserver roots.
+As a ``/srv/salt/top.sls`` file exists [literal]``merge_all`` is required to evaluate also the
+[literal]``top.sls`` file from the git repository.
++
+Example:
++
+----
+top_file_merging_strategy: merge_all
+----
+
 ext_pillar::
 List of external pillar interfaces.
 Salt can also serve pillar data from one or more git repositories.


mcalmer:
CC @meaksh @brejoc

modules/salt/pages/salt-gitfs.adoc
```diff
@@ -48,6 +48,17 @@ gitfs_remotes:
...
+top_file_merging_strategy:
+Define the strategy to merge multiple [literal]``top.sls`` files defined in your different fileserver roots.
+As a ``/srv/salt/top.sls`` file exists [literal]``merge_all`` is required to evaluate also the
+[literal]``top.sls`` file from the git repository.
++
+Example:
++
+----
+top_file_merging_strategy: merge_all
+----
```
Loquacity, line 60:
```suggestion
A [path]``/srv/salt/top.sls`` file already exists, so use [command]``merge_all`` to evaluate the [path]``top.sls`` file from the git repository.
For example:

----
top_file_merging_strategy: merge_all
----
```

Loquacity: Not sure about the 'so' here. Maybe @keichwa has a better suggestion.

Loquacity (changes requested):

modules/salt/pages/salt-gitfs.adoc
jcayouette: Perhaps `therefore` ?

mcalmer:
JFYI: We found out that this solution has downsides. So we will not merge this yet. We need to investigate better what we suggest.

modules/salt/pages/salt-gitfs.adoc
Loquacity: it's wordy, but probably clearer.

Loquacity:
> JFYI: We found out that this solution has downsides. So we will not merge this yet. We need to investigate better what we suggest.

Might be time to delete this PR?

Issue #735, opened 2020-12-11
{smr} definition issues


calancha:
## Documentation Update Required

The current definition at `uyuni-site.yml` (i.e., `for Retail`) doesn't fit well on all pages.

## What does the documentation say now?

`There is "an introduction to for Retail in this book."`

## What change is required?
According to my grammar checker, the previous line should read:
`There is "an introduction to Retail in this book."`

The macro could still do a decent job if we drop here the word `to` so that we get:
`There is "an introduction for Retail in this book."`'

Another option is to define `{smr}` as just `Retail`, i.e. we write `for`
_by hand;_ this adds a bit more flexibility; for instance, we can write things like:
`In the case of {smr} blah blah...`

## Subject Matter Expert (SME)

REQUIRED: Who is the subject matter expert for this topic?
@Loquacity 

## Version

Which version of the software is this feature/update expected to land in?
I saw this at 4.1.4, thus it must come after that version.

Which version of the documentation must this feature/update land in?
I am not sure; the master branch and anything else.

Does the content need to be backported to previous documentation versions?
It should go to any new version.

PR #740, opened 2020-12-15
use the smr macro only together with productname.

keichwa:
https://github.com/uyuni-project/uyuni-docs/issues/735

# Description

Fix smr macro usuage.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/735

### DIFF
--- a/modules/salt/pages/formula-tftpd.adoc
+++ b/modules/salt/pages/formula-tftpd.adoc
@@ -1,7 +1,7 @@
 [[tftpd-formula]]
 = TFTPD Formula
 
-The TFTPD formula is an {smr} formula, used to configure the TFTP service on the branch server.
+The TFTPD formula is an {productname} {smr} formula, used to configure the TFTP service on the branch server.
 
 
 
--- a/modules/client-configuration/pages/cobbler.adoc
+++ b/modules/client-configuration/pages/cobbler.adoc
@@ -12,8 +12,8 @@ Only supported features are documented here.
 
 [NOTE]
 ====
-If you intend to use your installation with {smr} formulas, do not follow this guide to configure Cobbler on the branch server.
-In {smr} installations, the TFTPD formula manages these settings.
+If you intend to use your installation with {productname} {smr} formulas, do not follow this guide to configure Cobbler on the branch server.
+In {productname} {smr} installations, the TFTPD formula manages these settings.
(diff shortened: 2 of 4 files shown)


modules/salt/pages/formula-tftpd.adoc
```diff
@@ -1,7 +1,7 @@
 [[tftpd-formula]]
 = TFTPD Formula
 
-The TFTPD formula is an {smr} formula, used to configure the TFTP service on the branch server.
+The TFTPD formula is an {productname} {smr} formula, used to configure the TFTP service on the branch server.
```
calancha, line 4: Sometimes we write:
`a {productname} {smr}`
or:
`an {productname} {smr}`

The former is suitable for Uyuni, the latter for Suse Manager.
Could we handle this `a` VS `an` systematically?

calancha (changes requested):
How could we handle systematically the `a` VS `an`? I mean,
I see the following things through the source:
`a {productname} {smr}`
`an {productname} {smr}`

this way, we are right 50%  of the time.

keichwa:
Tino Calancha <notifications@github.com> writes:

> this way, we are right 50%  of the time.

Yes, thanks for spotting!  There is no easy way out ;)  Of course,
another indirection layer is always possible: we must run a small
preprocessor step depending on the makefile target.

This idea isn't completly crazy.  We could use a general macro expander
such as 'm4' ;-)

Or ifdef within the adoc files.  I'll let Lana and Joseph decide what we
want to do in this case.

-- 
Karl Eichwalder

calancha (approved):
I let you decide if you want to merge it and address the `a` VS `an` issue separately,
or if you prefer address both things here.

modules/salt/pages/formula-tftpd.adoc
Loquacity: This happens throughout the docs with {productname}, and I think the adoc ifdef is not fine-grained enough to handle it neatly. I think for retail we could possibly create an additional entity, so we would have one to use with {productname} to create "{productname} for Retail" and one that is just "Retail" for where we want something shorter (or use SMR? is that an approved abbreviation? Might need input from Pau).

Loquacity (approved):

PR #753, opened 2021-01-05
Appstream note in Client Cfg Guide

Loquacity:
# Description

add note about conflicting packages in webui

# Target branches

- [x] Master (Default)
- [x] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/13547

Bug https://bugzilla.suse.com/show_bug.cgi?id=1179525

### DIFF
--- a/modules/client-configuration/pages/channels.adoc
+++ b/modules/client-configuration/pages/channels.adoc
@@ -55,7 +55,7 @@ For {redhat} based clients, additional packages are available through AppStream.
 In most cases, the AppStream packages are required to ensure that you have all the software you need.
 
 When you are managing AppStream packages in the {productname} {webui}, you might notice that you see contradicting suggestions for package updates.
-This is due to the {productname} not always interpreting the modular metadata correctly.
+This is due to the {productname} not being able to interpret the modular metadata correctly.
 You can use the content lifecycle management (CLM) AppStream filter to transform AppStream repositories into non-modular repositories for use with some upgrade operations.
 For more information about the CLM AppStream filters, see xref:administration:content-lifecycle-examples.adoc[].
 


Loquacity:
@Bischoff @cbbayburt Do you think there should be a troubleshooting entry to go with this? If so, what's the fix/workaround?

modules/client-configuration/pages/channels.adoc
```diff
@@ -54,6 +54,8 @@ For more information about content lifecycle management, see xref:administration
 For {redhat} based clients, additional packages are available through AppStream.
 In most cases, the AppStream packages are required to ensure that you have all the software you need.
 
+When you are managing AppStream packages in the {productname} {webui}, you might notice that you see contradicting suggestions for package updates.
+This is due to the {productname} not always interpreting the modular metadata correctly.
```
cbbayburt, line 58: Hmm, "not always interpreting the modular metadata correctly" doesn't sound right here. SUMA doesn't interpret the modular metadata at all, and the suggestions are expected to be wrong.

The basic idea is:
If you have modular packages in the list, do not expect SUMA UI to display correct stuff because it doesn't interpret modular metadata.

cbbayburt (changes requested):
> @Bischoff @cbbayburt Do you think there should be a troubleshooting entry to go with this? If so, what's the fix/workaround?

The fix is already mentioned here:

> You can use the content lifecycle management (CLM) AppStream filter to transform AppStream repositories into non-modular repositories for use with some upgrade operations.

I don't think we need a troubleshooting entry for this. The same thing is explained with a warning message in the UI as well.

modules/client-configuration/pages/channels.adoc
Loquacity: edited

cbbayburt (approved):
Looks good to me now. Thanks for the change @Loquacity 👍

jcayouette (approved):

Loquacity:
- [ ] Backport to 4.1:

PR #755, opened 2021-01-05
QS: Uyuni - Initial setup

Loquacity:
# Description

File setup, with content copied directly from source material.

Still to do (in a new PR, probably): 

- [ ] Edit to make narrative smoother (and more appropriate to a QS)
- [ ] Possibly move content around to use fewer files 
- [ ] Enable book for Uyuni product only 
- [ ] Deprecate duplicated content from original sources.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Partial: https://github.com/SUSE/spacewalk/issues/13040

### DIFF
--- a/modules/quickstart-uyuni/pages/proxy-reg-uyuni.adoc
+++ b/modules/quickstart-uyuni/pages/proxy-reg-uyuni.adoc
@@ -54,4 +54,4 @@ For more information about bootstrapping, see xref:client-configuration:bootstra
 . Navigate to menu:System Details[Software > Software Channels], and check that the proxy channel is selected.
 
 // Next action:
-For setting up a registered {productname} Proxy, see xref:uyuni-proxy-setup.adoc[].
+For setting up a registered {productname} Proxy, see xref:quickstart-uyuni:setup-uyuni-proxy.adoc[].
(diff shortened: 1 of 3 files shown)


modules/quickstart-uyuni/nav-quickstart-uyuni-guide.adoc
```diff
@@ -0,0 +1,26 @@
+// Quickstart Guide - Uyuni Antora Navigation List
+
+// PDF Installation Guide
+ifdef::backend-pdf[]
+= Quick Start - Uyuni: {productname} {productnumber}
```
keichwa, line 5: This probably needs checking.

keichwa (approved):

modules/quickstart-uyuni/nav-quickstart-uyuni-guide.adoc
Loquacity: Checking for what, exactly? It'll only get built when we build Uyuni, so it will only use Uyuni version numbers.

keichwa: I think this will come out as:

> Quick Start - Uyuni:  Uyuni 2021-02

Maybe, product number will look a little bit different.  I'd rather go for a meaningful sub-title such as "Server Installation".  But it is a minor issue and we can adjust it when all Uyuni guides are available.

Loquacity: Oh, I see what you're saying. Yes, you're probably right. We can take a closer look at this kind of thing when we do the guide enablement.

PR #761, opened 2021-01-08
enable snippets

jcayouette:
# Description

This PR adds snippet support across our documentation. There is one exception. Snippets will not work in the reference guide as we use sub directories. 

Use the following style to call a snippet: `include::{snippet}testing.adoc[]`

Keep in mind when calling a snippet as with any include the correct heading level should be used for the included content. You also must have line breaks above and below the include.

For example:

[admin-intro.adoc]
```
= Administration Guide

Some intro content

 include::{snippet}testing-snippets.adoc[]

some other content
```

[testing-snippets.adoc]
```
== Testing snippets

See heading level! ^^

some content
```

# Target branches

- [X] Master (Default)
- [X] Manager-4.1
- [ ] Manager-4.0

# Links


modules/snippets/pages/testing-snippets.adoc
```diff
@@ -0,0 +1,3 @@
+== Testing Snippets
+
+This is a snippet test!
```
Loquacity, line 3: You probably don't want to commit this file.

Loquacity (approved):

PR #750, opened 2020-12-28
Add details on reinstallting a new proxy and switching to salt managed

jordimassaguerpla:
With the current test it is not clear what happens if you reinstall a
proxy that was managed traditionaly to a salt one. This commit makes
this more explicity by stating that if you use the same activation key
and not remove the proxy from the list of systems, clients won't need to
be re-registered.

fixes #749
related to https://github.com/SUSE/spacewalk/issues/13257

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

# Description

Short summary of why you created this PR

# Target branches

Which documentation version does this PR apply to?

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes #<insert issue or PR link, if any>

PR #763, opened 2021-01-11
Proxy re-reg section

Loquacity:
~***DO NOT MERGE - DOCS REPO IS FROZEN***~

# Description

Adds new procedure to proxy re-registration section for switching a proxy from traditional to Salt. Adds xref from registration section. Some light editing.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Follows on from https://github.com/uyuni-project/uyuni-docs/pull/750

### DIFF
--- a/modules/installation/pages/proxy-setup.adoc
+++ b/modules/installation/pages/proxy-setup.adoc
@@ -244,7 +244,7 @@ After the installation of the new proxy, you might also need to:
 === Change a Proxy from Traditional to Salt
 
 You can reinstall the proxy to switch from a traditional to a Salt proxy.
-In this method, instead of a reactivation key, use the same activation key you used to register the proxy in the first instance.
+In this method, instead of a reactivation key, reuse the same activation key you used to originally register the proxy.
 This means you do not have to re-register the clients.
 
 


modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,46 +194,72 @@ filename: "pxelinux.0"
...
 == Replace a {productname} Proxy
 
-A proxy does not contain any information about the clients that are connected to it.
-Therefore, a proxy can be replaced by a new one at any time.
+You can replace a proxy at any time, as it does not store any information about the clients that are connected to it.
+This process is handled using a reactivation key, which prevents you from losing the history of the proxy.
+If you do not use a reactivation key, the replacement proxy will become a new one with a new ID.
 The replacement proxy must have the same name and IP address as its predecessor.
 
+You can also re-register a proxy to change it from a traditional proxy to a Salt proxy.
```
keichwa, line 202: According to the later description, this does not seem possible.  We probably must test it.

Maybe, I'm getting the wording wrong.  I think we should avoid the word "re-register" etc., if we do not refer to make use of the a re-activation key.  Maybe, "register again" or "register newly" would be easier to understand.

modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,46 +194,72 @@ filename: "pxelinux.0"
...
-After the first successful operation on a client, such as execution of a remote command or installation of a package or patch, this information will automatically be corrected.
-This may take some hours.
-====
+
+
+=== Change a Proxy from Traditional to Salt
+
+You can also use re-registration to switch a traditional proxy to a Salt proxy.
+In this method, instead of a reactivation key, use the same activation key you used to register the proxy in the first instance.
+This means you do not have to re-register the clients.
```
keichwa, line 248: Is this actually the case?  Maybe it is a special Salt feature.

jcayouette (commented):
I wonder if "Consider copying important or manually created data to a central place that can also be accessed by the new proxy."

Should be:

"Consider copying important or custom data to a central place that can also be accessed by the new proxy."

modules/installation/pages/proxy-setup.adoc
jordimassaguerpla: I would not use "re-register" because this could be misunderstood. I would say instead `you can also **reinstall** a proxy to change it .... . In this case, you won't need a reactivation key. See (link to Change a Proxy from Traditional to Salt).`

modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,46 +194,72 @@ filename: "pxelinux.0"
...
+
+
+[IMPORTANT]
+====
+During the installation of the proxy, clients will not be able to reach the {productname} Server.
+After you have deleted a proxy, the systems list can be temporarily incorrect.
+All clients that were previously connected to the proxy will show as being directly connected to the server instead.
+After the first successful operation on a client, such as execution of a remote command or installation of a package or patch, this information will automatically be corrected.
+This may take some hours.
+====
```
jordimassaguerpla, line 212: ?? Where is this change coming from? Are you sure about this?

modules/installation/pages/proxy-setup.adoc
jordimassaguerpla: Yes it is . See https://github.com/SUSE/spacewalk/issues/13257

modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,46 +194,72 @@ filename: "pxelinux.0"
...
-After you have deleted a proxy, the systems list can be temporarily incorrect.
-All clients that were previously connected to the proxy will show as being directly connected to the server instead.
-After the first successful operation on a client, such as execution of a remote command or installation of a package or patch, this information will automatically be corrected.
-This may take some hours.
-====
+
+
+=== Change a Proxy from Traditional to Salt
+
+You can also use re-registration to switch a traditional proxy to a Salt proxy.
```
jordimassaguerpla, line 246: I would use "reinstall" instead of re-registration because I think it could be easy misunderstood with reactivation. I know they are 2 things, but the meaning is so close that is confusing.

modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,46 +194,72 @@ filename: "pxelinux.0"
...
+This means you do not have to re-register the clients.
+
+
+
+.Procedure: Replacing a Traditional Proxy with a Salt Proxy
+. Before starting the migration, save the data from the old proxy, if needed.
+  Consider copying important or manually created data to a central place that can also be accessed by the new proxy.
+. Shut down the proxy.
+. Install a new {productname} Proxy.
+  For installation instructions, see xref:install-proxy-unified.adoc[Proxy Installation].
```
jordimassaguerpla, line 257: Make sure you have the same ip address.

modules/installation/pages/proxy-setup.adoc
Loquacity: It was already extant in the text, I just moved it. Can we verify it somehow?

jordimassaguerpla: If it was there, I have no issue with it. I just didn't see it and thought it was new text.

jordimassaguerpla (approved):

Loquacity:
> I wonder if "Consider copying important or manually created data to a central place that can also be accessed by the new proxy."
> 
> Should be:
> 
> "Consider copying important or custom data to a central place that can also be accessed by the new proxy."

Sorry, I missed this in the last editing pass. Done now.

modules/installation/pages/proxy-setup.adoc
```diff
@@ -194,46 +194,72 @@ filename: "pxelinux.0"
...
-All clients that were previously connected to the proxy will show as being directly connected to the server instead.
-After the first successful operation on a client, such as execution of a remote command or installation of a package or patch, this information will automatically be corrected.
-This may take some hours.
-====
+
+
+=== Change a Proxy from Traditional to Salt
+
+You can reinstall the proxy to switch from a traditional to a Salt proxy.
+In this method, instead of a reactivation key, use the same activation key you used to register the proxy in the first instance.
```
keichwa, line 247: Maybe, I'm wrong, but I think the most important thing is to assign the same name and IP to the new proxy.  Yes, this info is provided later on, but should be mentioned here.

The expression about using the "same" activation key is a little bit ambiguous because it is not available on the newly installed proxy, and it could either mean to create an activation key with the same name/label, the same info, or both.

keichwa (approved):

modules/installation/pages/proxy-setup.adoc
Loquacity: ^^ @jordimassaguerpla ?

jordimassaguerpla: "same" activation key means reusing the one you used for the "old" proxy, so SUMA does already have it in the db. Maybe we should say "reuse" instead of "same"...

Loquacity: ok, i've updated the wording slightly. LMK.

jordimassaguerpla (approved):

keichwa (approved):

PR #769, opened 2021-01-14
A/an {productname}

Loquacity:
# Description

Changes indefinite article (a/an) before {productname} to be a definite article (the). This is grammatically more correct, but also means we don't introduce an error when {productname} is expanded to Uyuni (which starts with a vowel sound) rather than SUSE Manager (which starts with a consonant sound). Additionally, because other languages often have more complicated article use than English does, it should help make our translations better.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/13551

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -22,3 +22,4 @@
   clones in Client Configuration Guide (bsc#1179277)
 - Adds note about Appstream Packages in Channels section
   of Client Cfg Guide (bsc#1179525)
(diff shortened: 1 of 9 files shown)


keichwa (approved):
this was huge, thanks!

PR #784, opened 2021-01-26
only mention clients that are actually discussed later on (registration)

keichwa:
# Description

Clean up client registration intro (SUSE clients)

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/13419

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -8,8 +8,8 @@
 
 - In SUSE client registration intro of the Client Configuration Guide, only
   mention SLES versions that are discussed later on.
-- Adds openSUSE Leap SP migration to the SP migration section of the Client
-  Configuration Guide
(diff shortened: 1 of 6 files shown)


jcayouette (commented):
Do we need to add SP3 for 4.2?

Loquacity (approved):

keichwa:
> Do we need to add SP3 for 4.2? I updated our site.yml files to reflect the service pack migration change already.

Probably, thanks for the reminder!  OTOH, because of potential backports, I'd like to start with referencing SP3 as late as possible.  Maybe, it now is late enough.  I think we should do it with a separate PR.

Issue #783, opened 2021-01-25
Missing example autoyast profile for SLES 15 SPX & Mention to upgrade profiles from previous major release


raulosuna:
## Documentation Update Required

Autoyast example profile for SLES 15 SPX (GA/SP1/SP2 and SP3 in the future) is needed. Many customers don't have a clue about how to generate one, this generates a lot of support cases. A TID is not the place where this should be documented. We're receiving a lot of support cases about this, many of which end up in bugzilla/L3 generating extra workload on Engineering.
In addition to this, most of the queries are for upgrades, rather than new installations (SLES 11 SP4 to SLES 12 SPX, or SLES 12 SPX to SLES 15 SPX). Should there be any changes in the related autoyast profiles, that should be mentioned here (or at least, a link to the documentation for it, in case it's too long for everything to be included in the same place)

## What does the documentation say now?

There's a basic autoyast profile for SLES 12 SPX installation.

## What change is required?

We'd kindly ask for the same in SLES 15 family,

## Subject Matter Expert (SME)

Both SUSE Manager and Yast teams.

## Version

Which version of the software is this feature/update expected to land in?
All supported versions of SUSE Manager: 4.0 & 4.1. 4.2 is about to be released, so this might need to be considered as well.
Which version of the documentation must this feature/update land in?
4.0, 4.1, 4.2
Does the content need to be backported to previous documentation versions?
Yes, 4.0.

Loquacity:
Hi Raul, thanks for the report. Which document are you looking at?

mcalmer:
We have this https://github.com/SUSE/manager-build-profiles/tree/master/AutoYaST/SLES-15-SP2-installation

raulosuna:
> Hi Raul, thanks for the report. Which document are you looking at?

My bad, I thought that the "contribute" button would save the exact page of the documentation I was reporting. It is located here:

https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/autoyast-example.html

Michael's examples would fit here pretty well. I assume that if a SLES for SAP example file is provided for SLES 15, the same should exist for SLES 12.

[...]

PR #788, opened 2021-01-28
AY examples section - update

Loquacity:
# Description

Because the actual scripts had been moved from the doc into the public repo, a lot of the content on this page didn't make sense anymore, and caused some confusion for readers. Additionally, the content that was there was out of date. This edit removes the out of date content, and makes it clearer that the scripts are in the repo. It also points to the instructions provided with the scripts, rather than trying to replicate instructions in the doc. Also updates the xref from the Ref Guide to the Admin Guide (the content used to be in the Ref Guide, but has now moved). 

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/783


modules/client-configuration/pages/autoyast-example.adoc
```diff
@@ -1,103 +1,20 @@
 [[autoyast-example]]
 = AutoYast Example File
 
-{suse} provides templates of AutoYaST profiles in the https://github.com/SUSE/manager-build-profiles[SUSE/manager-build-profiles] public GitHub repository.
+{suse} provides templates of various AutoYaST profiles in the https://github.com/SUSE/manager-build-profiles/tree/master/AutoYaST[SUSE/manager-build-profiles/AutoYaST] public GitHub repository.
 
+Each script requires you to set some variables before you use it.
```
keichwa, line 6:
```suggestion
Each profile requires you to set some variables before you use it.
```
I think we'd better do not switch (descriptive) terms from one sentence to the next.  Or we must say why we now want to call the profiles scripts.  This is just a proposal and in general we can address this later (we probably must check the complete autoinstallation section).  I always forgot to create an issue about this.

keichwa (approved):

PR #789, opened 2021-01-29
QS:Uyuni Technical Review

Loquacity:
# Description

Set up this PR to track changes from @juliogonzalez technical review. Proxy updates (thanks @keichwa!), plus some general editing and syntax edits. 

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/13798

### DIFF
--- a/modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
+++ b/modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
@@ -44,7 +44,7 @@ Begin by installing openSUSE Leap on a physical or virtual machine.
 To ensure that the proxy is accessible across the network, you must have a resolvable fully-qualified domain name (FQDN) on the openSUSE Leap system before you begin the installation.
 You can configure an FQDN with {yast} by navigating to menu:System[Network Settings > Hostname/DNS].
 
-When you have installed openSUSE Leap on the proxy and configured the FQDN, you can prepare the {productname} Server, and and register the openSUSE Leap system as a client.
+When you have installed openSUSE Leap on the proxy and configured the FQDN, you can prepare the {productname} Server, and register the openSUSE Leap system as a client.
 
 
 
@@ -275,43 +275,6 @@ Use this DHCP configuration for the zone which contains the clients to be provis
 next-server: <IP_Address_of_Proxy>
 filename: "pxelinux.0"
 ----
-
-
-////
-== Register a Proxy
-
-When the proxy is set up, you can register it to the {productname} Server.
-
-
-.Procedure: Registering the Proxy
-. On the {productname} Server, create openSUSE Leap and the {productname} Proxy channels with the [command]``spacewalk-common-channels`` command.
-    [command]``spacewalk-common-channels`` is part of the [package]``spacewalk-utils`` package:
-+
-----
-spacewalk-common-channels \
-opensuse_leap15_2 \
-opensuse_leap15_2-non-oss \
-opensuse_leap15_2-non-oss-updates \
-opensuse_leap15_2-updates \
-opensuse_leap15_2-uyuni-client \
-uyuni-proxy-stable-leap-152
-----
-+
-Instead of [systemitem]``uyuni-proxy-stable-leap-152`` you can also try the latest development version, called [systemitem]``uyuni-proxy-devel-leap``.
-For more information, see xref:client-configuration:clients-opensuse.adoc[].
-+
-. Create an activation key with openSUSE Leap as a base channel and the other channels as child channels.
-  For more information about activation keys, see xref:client-configuration:activation-keys.adoc[].
-. Modify a bootstrap script for the proxy.
-  Add the GPG key to the [systemitem]``ORG_GPG_KEY=`` parameter.
-  For more information, see xref:client-configuration:clients-opensuse.adoc[].
-  For more information about bootstrap scripts, see xref:client-configuration:registration-bootstrap.adoc[].
-. Bootstrap the client using the script.
-  For more information, see xref:client-configuration:registration-bootstrap.adoc[].
-. Navigate to menu:Salt[Keys] and accept the key.
-  When the key is accepted, the new proxy will show in menu:Systems[Overview] in the [guimenu]``Recently Registered Systems`` section.
-. Navigate to menu:System Details[Software > Software Channels], and check that the proxy channel is selected.
-////
 
 
 


modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
```diff
@@ -271,7 +271,7 @@ filename: "pxelinux.0"
 ----
 
 
-
+////
```
juliogonzalez, line 274: Why this comment? Shouldn't we just remove this block?

juliogonzalez (commented):
Some comments.

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
```diff
@@ -1,57 +1,94 @@
 [[install-proxy-uyuni]]
 = Install {productname} Proxy with {opensuse} Leap
 
-{productname} Proxy can be installed on openSUSE Leap.
-The proxy is installed in the same way as a client, but is designated as a proxy server during installation.
+{productname} Proxy can be installed as a server extension on openSUSE Leap.
```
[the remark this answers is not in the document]
Loquacity: I think I liked the original version of this better, but since @keichwa changed it, I'd like for him to weigh in here.

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
```diff
@@ -1,57 +1,94 @@
...
+
+
+
+== Register the openSUSE Leap system
+
+Begin by installing openSUSE Leap on a physical or virtual machine.
+To ensure that the proxy is accessible across the network, you must have a resolvable fully-qualified domain name (FQDN) on the openSUSE Leap system before you begin the installation.
+You can configure an FQDN with {yast} by navigating to menu:System[Network Settings > Hostname/DNS].
+
+Prepare the proxy installation settings on the server and register the openSUSE Leap system as a client.
```
Loquacity, line 46:
```suggestion
When you have installed openSUSE Leap on the proxy and configured the FQDN, you can prepare the {productname} Server, and and register the openSUSE Leap system as a client.
```

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
```diff
@@ -1,57 +1,94 @@
...
+
+
+
+.Procedure: Registering the openSUSE Leap system
+. On the {productname} Server, create an activation key with openSUSE Leap as a base channel and the proxy and the other channels as child channels.
+  For more information about activation keys, see xref:client-configuration:activation-keys.adoc[].
+. Modify a bootstrap script for the proxy.
```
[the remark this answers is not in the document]
Loquacity: Sure, but this is a Quick Start. All the other methods are in the main docs.

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
```diff
@@ -1,57 +1,94 @@
...
+  Add the GPG key to the [systemitem]``ORG_GPG_KEY=`` parameter.
```
Loquacity, line 54:
```suggestion
  If required, add your GPG key to the [systemitem]``ORG_GPG_KEY=`` parameter.
```

juliogonzalez: I think I understand now what GPG key you are talking about. The one for Uyuni, I guess, as we suggest for regular clients? I so, this should be specified here.

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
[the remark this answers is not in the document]
juliogonzalez: OK, given my other question about the GPG key (I think you need to specify one in particular), this bootstrap script method is the best one.

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
keichwa: Can we agree on the exact wording?  For me it worked, if I'd use one of those keys provided on the server in the bootstrap directory.  Thus at least by default, you can use the "official" keys.  I'm not sure whether we should this key "your GPG key" here?

I think, by default, the key is always required.  Maybe, it is not needed if you use some kind of cloned channels.

My original wording was probably to short, but I think we must say something along these lines: "Adding the {productname} provided GPG key is mandatory." I could be wrong, though.  @juliogonzalez and @Loquacity can you please help again?

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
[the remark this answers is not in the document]
keichwa: Maybe, this is better:

> The proxy is installed in the same way as a client, but is designated as a proxy server during installation.
> Designation happens when you add the Uyuni Proxy pattern, and execute the proxy setup script.

I think this "in the same way" and "during installation" phrases without saying that quite some additional time and work is required is misleading. Of course, I do not want to insist on my wording!

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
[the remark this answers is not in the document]
keichwa: Maybe, I should add:

> [...] as outlined in the following procedure.

This at least is what I mean. ;) @Loquacity

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
```diff
@@ -1,57 +1,94 @@
...
+. Reboot the client.
 
-When the installation is complete, you can continue with {productname} setup.
-For more information, see xref:installation:uyuni-proxy-registration.adoc[].
 
 
 == Prepare the Proxy
 
 Before you begin, ensure that the proxy pattern is installed correctly.
+// FIXME: there might be a better way to check it:
```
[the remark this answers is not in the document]
keichwa: iirc, some time ago mcalmer/mantel proposed to do it that way. I'll check it next time, separately, but for the moment it is good enough.

modules/quickstart-uyuni/pages/install-uyuni-proxy.adoc
juliogonzalez: @keichwa AFAIK you need to manually add the Uyuni key (which is not *your* key) to ``ORG_GPG_KEY=`` or am I missing something?

Otherwise when you bootstrap the proxy that will work. But if you need to install any other packages, or update, it won't work.

So shouldn't we remove "your GPG key" from here, and just tell the user:

>  Add the GPG key for Uyuni to the [systemitem]``ORG_GPG_KEY=`` parameter like this: ``ORG_GPG_KEY=whatever``

whatever, because I don't remember what the filename is. But it's easy if you look at the server at `/srv/www/htdocs/pub/` (or otherwise via web https://hostname/pub/

Loquacity: Updated

Loquacity: image

jcayouette (approved):

PR #810, opened 2021-02-12
Add tshooting for centos 6 clients

Loquacity:
# Description

CentOS 6 is now EOL, and the repos are providing packages that cause bootstrapping to fail. This PR adds a troubleshooting section to the Client Cfg Guide, and links to the section from admonitions in the CentOS chapter.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/13930
Bug https://bugzilla.suse.com/show_bug.cgi?id=1181932

### DIFF
--- a/modules/client-configuration/pages/tshoot-clients.adoc
+++ b/modules/client-configuration/pages/tshoot-clients.adoc
@@ -64,21 +64,22 @@ If you need to bootstrap new {centos} 6 clients you can edit the existing reposi
 
 .Procedure: Troubleshooting Bootstrapping a New {centos} 6 Client
 . On the {centos} 6 client, at the command prompt, open the ``CentOS-Base.repo`` file, located in the ``/etc/yum.repos.d/`` directory.
-. Locate the ``baseurl`` line, pointing to ``mirrorlist.centos.org``.
+. Locate the ``mirrorlist`` entry, pointing to ``mirrorlist.centos.org``.
 There may be more than one entry.
 For example:
 +
 ----
-baseurl=http://mirrorlist.centos.org/?release=6&arch=$basearch&repo=os
+mirrorlist=http://mirrorlist.centos.org/?release=6&arch=$basearch&repo=os
 ----
 +
-. Edit each ``baseurl`` line to point to a ``vault.centos.org`` URL.
+. Comment out the ``mirrorlist`` entry, to prevent the package manager looking for the URL.
+. Edit the ``baseurl`` line to point to a ``vault.centos.org`` URL, and specify the {centos} 6 repositories.
 For example:
 +
 ----
 baseurl=https://vault.centos.org/centos/6/os/$basearch/
 ----
-. Comment out each ``mirrorlist`` entry, to prevent the package manager looking for the URL.
+. Repeat for each repository listed in the file.
 . Bootstrap the client.
   For more information about bootstrapping {centos} clients, see xref:client-configuration:clients-centos.adoc[].
 


modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,45 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+{centos} 6 is now at end-of-life, and the ISO images provided in the client repository for this operating system are out of date.
+Bootstrapping new {centos} 6 clients using these packages will fail.
+This failure will not happen on {centos} 6 clients that are already installed and bootstrapped.
+
+If you need to bootstrap new {centos} 6 clients you can remove the old repositories, and create a new repository to provide the correct packages.
+
+
+
+.Procedure: Troubleshooting Bootstrapping a New {centos} 6 Client
+. On the {productname} Server, at the command prompt, remove the repositories provided by the {centos} 6 ISO:
```
mcalmer, line 66: This is not on the SUSE Manager Server. This must happen on the CentOS client.

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,45 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
++
+----
+rm CentOS-Base.repo CentOS-Debuginfo.repo CentOS-fasttrack.repo
+----
+. Add a new repository:
++
+----
+zypper addrepo -f http://vault.centos.org/6.10/os/$basearch/
```
mcalmer, line 74: CentOS does not have zypper. The files need to be written manually.
I would suggest we do not describe to remove them, but just change the URL to vault.

mcalmer (changes requested):

modules/client-configuration/pages/tshoot-clients.adoc
Loquacity: Updated.

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,40 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
 
 
 
+== Bootstrap End-of-Life {centos} 6 Clients
+
+{centos} 6 is now at end-of-life, and the ISO images provided in the client repository for this operating system are out of date.
+Bootstrapping new {centos} 6 clients using these packages will fail.
+This failure will not happen on {centos} 6 clients that are already installed and bootstrapped.
+
+If you need to bootstrap new {centos} 6 clients you can remove the old repositories, and create a new repository to provide the correct packages.
```
juliogonzalez, line 61: "Remove and add"?

I'd say that your procedure describes "changing" the baseurl for the existing repositories :-)

juliogonzalez, line 61: "Remove and add"?

I'd say that your procedure describes "changing" the baseurl for the existing repositories :-)

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,40 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+
+
+
+.Procedure: Troubleshooting Bootstrapping a New {centos} 6 Client
+. On the {centos} 6 client, at the command prompt, locate the ``CentOS-Base.repo`` file provided by the {centos} 6 ISO.
+In most cases, it is located at `/etc/yum/repos.d`.
```
juliogonzalez, line 67: I am not aware of any cases where the `.repo` files for the official CentOS repositories won't be there.

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,40 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+. Open the repository file in your preferred text editor, and locate the ``baseurl`` line, pointing to ``mirror.centos.org``.
+There may be more than one entry.
+For example:
++
+----
+baseurl=http://mirror.centos.org/centos/$releasever/os/$basearch
+----
++
+Edit each ``baseurl``  line to point to a ``vault.centos.org`` URL.
```
juliogonzalez, line 76: Seems you have two whitespaces after `baseurl`

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,40 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
 
 
 
+== Bootstrap End-of-Life {centos} 6 Clients
+
+{centos} 6 is now at end-of-life, and the ISO images provided in the client repository for this operating system are out of date.
```
juliogonzalez, line 57: Not just the ISOs. Most likely the cloud images (qcow2) provided by CentOS as well. No idea about the cloud images at AWS or other Cloud providers, but my guess is that they are affected as well.

modules/client-configuration/pages/tshoot-clients.adoc
mcalmer: The original baseurl does not look like this. It was:

`http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=os`

This needs to get mapped to the new URL. But customers who still use this, should know how to map.
I would not provide a list.

modules/client-configuration/pages/tshoot-clients.adoc
Loquacity: oh, I changed the procedure and didn't update the stem :woman_facepalming:

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,39 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+
+
+.Procedure: Troubleshooting Bootstrapping a New {centos} 6 Client
+. On the {centos} 6 client, at the command prompt, open the ``CentOS-Base.repo`` file, located in the `/etc/yum/repos.d/` directory.
+. Locate the ``baseurl`` line, pointing to ``mirrorlist.centos.org``.
+There may be more than one entry.
+For example:
++
+----
+baseurl=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=os
```
juliogonzalez, line 72: Err... `release=7`? This is about CentOS6, isn't it?

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,39 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+----
++
+. Edit each ``baseurl`` line to point to a ``vault.centos.org`` URL.
+For example:
++
+----
+baseurl=http://vault.centos.org/?release=7&arch=$basearch&repo=os
```
juliogonzalez, line 79: Besides the problem, with `release=7`,  I don't think this is the right replacement to be done.

If I am not wrong the right URL is `https://vault.centos.org/centos/6/os/$basearch/`, but please **test** this on a real CentOS6 image in case I am wrong (a docker container should be enough as well).

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,39 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+{centos} 6 is now at end-of-life, and the images provided in the client repository for this operating system are out of date.
+Bootstrapping new {centos} 6 clients using these packages will fail.
+This failure will not happen on {centos} 6 clients that are already installed and bootstrapped.
+
+If you need to bootstrap new {centos} 6 clients you can edit the existing repositories to reflect the correct URL.
+
+
+
+.Procedure: Troubleshooting Bootstrapping a New {centos} 6 Client
+. On the {centos} 6 client, at the command prompt, open the ``CentOS-Base.repo`` file, located in the `/etc/yum/repos.d/` directory.
```
juliogonzalez, line 66: In fact, shouldn't the user do this change for more files than `CentOS-Base.repo`?

Please test the procedure yourself. No need for a SUSE Manager Server. Create a CentOS6 VM or start a docker container and give it a try :-)

juliogonzalez (changes requested):

modules/client-configuration/pages/tshoot-clients.adoc
mcalmer: yes, sorry. As 6 went already away, I had to copy the centos 7 entry as example.
@Loquacity just change `7` to `6` and we are good here.

modules/client-configuration/pages/tshoot-clients.adoc
mcalmer: Right. @Loquacity vault is a direct URL and not the mirror list thingy.

The "old" url uses mirror list, while the new URL is just a plain URL to vault.
This is what I mean with "mapping" is needed. You need to turn the parameters of the mirror list into a real path.

Loquacity: Right, that makes sense.

modules/client-configuration/pages/tshoot-clients.adoc
Loquacity: I did. The original text says to remove `CentOS-Base.repo`, `CentOS-Debuginfo.repo`, and `CentOS-fasttrack.repo`. When I looked at those files, only the base one has entries for the baseurl. So when we changed the procedure to just editing the file rather than deleting/re-writing, it seemed sensible not to worry about the other two. I guess we could still delete them, though, if you think that's important?

mcalmer (approved):

modules/client-configuration/pages/tshoot-clients.adoc
juliogonzalez: > I guess we could still delete them, though, if you think that's important?

What I mean is that, as @mcalmer suggested somewhere else on this PR, we should give then `CentOS-Base.repo` as an example, and request the user to do the changes for the others files as welll, or disable them.

Otherwise I think bootstrapping from client will work (works even if some repositories fail to refresh, as long as packages are available), but maybe not from the WebUI (not 100% sure)

BTW, for both base and fastrack I don't see `baseurl` as you mention below, but `mirrorlist` (`baseurl` is commented)

```
[root@824fbdab45b8 yum.repos.d]# grep -ir 'mirrorlist' *
CentOS-Base.repo:# If the mirrorlist= does not work for you, as a fall back you can try the 
CentOS-Base.repo:mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=os&infra=$infra
CentOS-Base.repo:mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=updates&infra=$infra
CentOS-Base.repo:mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=extras&infra=$infra
CentOS-Base.repo:mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=centosplus&infra=$infra
CentOS-Base.repo:mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=contrib&infra=$infra
CentOS-fasttrack.repo:mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=fasttrack&infra=$infra
```
Are you sure in your tests you had `baseurl=`? I am 99.99% sure it should be `mirrorlist=` (but the hcange you suggest to `baseurl=` seems to be fine.

I'd suggest you try again to apply the instructions, and at the end run a `yum search vim` to see what if you can get the results without errors.

juliogonzalez (commented):
Not 100% sure, but I think some  changes are still required.

modules/client-configuration/pages/tshoot-clients.adoc
Loquacity: I have run the procedure as written, with success. I only changed the baseurl entry in the base repo.

keichwa, line 66:
```suggestion
. On the {centos} 6 client, at the command prompt, open the ``CentOS-Base.repo`` file, located in the `/etc/yum/repos.d/` directory.
```suggestion
. On the {centos} 6 client, at the command prompt, open the ``CentOS-Base.repo`` file, located in the ``/etc/yum/repos.d/`` directory.
```
for consistency.

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,39 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+. Edit each ``baseurl`` line to point to a ``vault.centos.org`` URL.
+For example:
++
+----
+baseurl=https://vault.centos.org/centos/6/os/$basearch/
+----
+. Bootstrap the client.
+  For more information about bootstrapping {centos} clients, see xref:client-configuration:clients-centos.adoc[].
+
+For more information about the {centos} 6 end-of-life, see  http://mirror.centos.org/centos/6/readme[].
```
keichwa, line 84:
```suggestion
For more information about the {centos} 6 end-of-life, see  http://mirror.centos.org/centos/6/readme.
```
I'm not sure whether this makes a difference.

keichwa (approved):

modules/client-configuration/pages/tshoot-clients.adoc
juliogonzalez: Are you 100% sure the original line was `baseurl=` and not `mirrorlist=` as I see in my docker container?

jcayouette (approved):

modules/client-configuration/pages/tshoot-clients.adoc
Loquacity: Yes, I only changed the baseurl entry.

Loquacity: I don't see a change here?

keichwa (approved):

modules/client-configuration/pages/tshoot-clients.adoc
keichwa: > I don't see a change here?

I created an artifact...  suggestion within a suggestion plus a syntax error.  it looks better now: I propose double tics for consistency.

modules/client-configuration/pages/tshoot-clients.adoc
```diff
@@ -52,6 +52,39 @@ Therefore, we recommended that you do not power on bare metal systems multiple t
...
+
+
+.Procedure: Troubleshooting Bootstrapping a New {centos} 6 Client
+. On the {centos} 6 client, at the command prompt, open the ``CentOS-Base.repo`` file, located in the ``/etc/yum/repos.d/`` directory.
+. Locate the ``baseurl`` line, pointing to ``mirrorlist.centos.org``.
+There may be more than one entry.
+For example:
++
+----
+baseurl=http://mirrorlist.centos.org/?release=6&arch=$basearch&repo=os
```
juliogonzalez, line 72: I have to ask again: are you 100% sure this is `baseurl=`?

Usually at Yum, mirrors use `mirrorlist=`.

Loquacity: I will test on my VM again today for you.

Loquacity: This is what I have on my CentOS 6 machine:

image

When I run `yum update` it is successful, although I've just now realised I'm getting a couple of warnings right before it loads the packages:

image

If I change the mirrorlist (just the first entry):

image

Now `yum update` still works, but now I get this error instead:

image

If I comment out the `baseurl` entry:

image

Now it stops working altogether:

image

If I revert to the old mirrorlist URL, and keep the baseurl commented:

image

It fails completely again:

image

My best guess is that it's looking for the URL in the mirrorlist, failing, and falling back to the baseurl, which is succeeding. You might have a better understanding though.

HTH.

juliogonzalez: As I guessed, the VM has `mirrorlist` enabled, so the PR needs to be changed.

What you need, as a user is:

1. Comment the `mirrorlist=` line.
2. Make sure you have a `baseurl=` line with the vault repository that is "equivalent" to the one you had at `mirrorlist=`

And that, for all repositories.

Try the procedure folllowing that instruction and you will see it works just fine.

Loquacity: Yep, that's got it. I'll update the content.

image

mcalmer (approved):

modules/client-configuration/pages/tshoot-clients.adoc
juliogonzalez: I still see `baseurl=http://mirrorlist.centos.org/?release=6&arch=$basearch&repo=os` as the string to be commented? Wha am I missing? Shouldn't it mention `mirrorlist=` instead?

juliogonzalez (commented):
I am still confused :-|

modules/client-configuration/pages/tshoot-clients.adoc
Loquacity: Where do you see that? L81 says to comment out `mirrorlist` :point_down:

juliogonzalez: Line 71.

juliogonzalez: I still see this:

```
.Procedure: Troubleshooting Bootstrapping a New {centos} 6 Client
. On the {centos} 6 client, at the command prompt, open the ``CentOS-Base.repo`` file, located in the ``/etc/yum.repos.d/`` directory.
. Locate the ``baseurl`` line, pointing to ``mirrorlist.centos.org``.
There may be more than one entry.
For example:
+
----
baseurl=http://mirrorlist.centos.org/?release=6&arch=$basearch&repo=os
----
```
AFAIK such line doesn't exist, because it should start with `mirrorlist=` and it's what you need to comment (as you tell the user to do at line 81).

And also I don't think you need to edit ". Edit each ``baseurl`` line to point to a ``vault.centos.org`` URL." :-?

If the user does that, the consequence is enabling all the 6.X repositories as well.

The instructions should be rather something like this:

> This is an example to replace the `os` repository:
> 1. Find the line `mirrorlist=http://mirrorlist.centos.org/?release=6&arch=$basearch&repo=os`
> 2. Comment the line
> 3. Add a new line pointing to the `os` repository at `vault.centos.org`: `baseurl=https://vault.centos.org/centos/6/os/$basearch/`
> Repeat for all other repositories you are using from `mirrorlist.centos.org`

NOTE: Test the procedure I described, just in case, but I am 99.99% sure it will work.

Loquacity: Both of our procedures end up with the same result, I've tidied mine up to make it a little more logical, since there was obviously confusion in the way I wrote it originally. Note that L71 doesn't say to comment anything out, it just says to *locate* the line to work on :wink: Also, there's no need to add a new baseurl entry, they're all already there on a standard install.

juliogonzalez (approved):
LGTM, assuming it was tested (in case my memory didn't work correctly).

PR #817, opened 2021-02-22
Adds clarification for which GPG keys required

Loquacity:
# Description

It turns out knowing which GPG key to install is not always straight forward. This adds some admonitions for specific clients.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

- Fixes https://github.com/SUSE/spacewalk/issues/13908
- Bug https://bugzilla.suse.com/show_bug.cgi?id=1181695

### DIFF
--- a/modules/client-configuration/pages/clients-sle.adoc
+++ b/modules/client-configuration/pages/clients-sle.adoc
@@ -127,6 +127,12 @@ ifeval::[{uyuni-content} == true]
 
 include::snippets/trust_gpg.adoc[]
 
+[NOTE]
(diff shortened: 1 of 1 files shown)


mcalmer (approved):
Not sure if this is usefull or not. These GPG keys should be trusted on the OSes already.
I do not know what problem QA faced.
But if the people wonder where the SLE15 GPG key is, they get now the answer.

So ok from my point of view

keichwa (approved):

PR #829, opened 2021-03-03
Add external network description + update images

Loquacity:
# Description

Adds description of setting up external network, and upgrades network diagrams to new icon set + colours.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/13494

### DIFF
--- a/modules/retail/pages/retail-admin-network.adoc
+++ b/modules/retail/pages/retail-admin-network.adoc
@@ -28,7 +28,6 @@ You can configure this using the PXE formula.
 For more information, see xref:salt:formula-pxe.adoc[].
 * For ``ftp``, you can use an A record that resolves to an IP address instead of a CNAME.
 Alternatively, you can change the terminal boot process using Salt pillars.
-For more information, see xref:retail:retail-deploy-terminals.adoc[].
-
+For more information, see xref:retail-deploy-terminals.adoc[].
 
 For a description of the different networking architectures, see xref:retail:retail-network-arch.adoc[].


Loquacity:
PDF for review (pg12): 

[suse_manager_retail_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/6080683/suse_manager_retail_guide.pdf)

modules/retail/pages/retail-network-arch.adoc
```diff
@@ -31,23 +31,24 @@ endif::[]
 
 == Branch Server Network Configuration
 
-The branch server can operate in several different network configurations.
-The two most common configurations are a dedicated network, or a shared network.
+You can use branch servers in several different network configurations, depending on your installation requirements.
```
keichwa, line 34:
```suggestion
You can use branch servers in different network configurations, depending on your installation requirements.
```

keichwa (approved):
very minor suggestion.

jcayouette (approved):
Looks good!

Loquacity:
Thanks for the reviews @keichwa and @jcayouette :) I would like to get @aaannz 's tick on the images before I merge though, if possible ...

aaannz:
All looks good, but it is missing important information how users need to configure their dhcp and dns in case external or shared networks are used.

Both external and shared network DHCP servers need to have PXE support enabled, for example like this:
```
next-server: <branch server IP>
if option arch = 00:07 {
     filename "boot/shim.efi";
} else {
     filename "boot/pxelinux.0";
}
```
DNS servers then need to resolve `salt` and `ftp` hostnames as CNAMEs to the correct branch server hostname.

Loquacity:
> All looks good, but it is missing important information how users need to configure their dhcp and dns in case external or shared networks are used.
> 
> Both external and shared network DHCP servers need to have PXE support enabled, for example like this:
> 
> ```
> next-server: <branch server IP>
> if option arch = 00:07 {
>      filename "boot/shim.efi";
> } else {
>      filename "boot/pxelinux.0";
> }
> ```
> 
> DNS servers then need to resolve `salt` and `ftp` hostnames as CNAMEs to the correct branch server hostname.

That content would belong in a different place. I realised this when I was going through, that we don't have any actual instructions for setting up the networking architectures we describe. This bit is under 'Installation', but I think the implementation details would need to go under 'Administration'. Perhaps we can merge this as-is, and create a new issue to document implementation? It's going to take a bit more effort, but I think it's probably worth it.

Loquacity:
> > That content would belong in a different place. I realised this when I was going through, that we don't have any actual instructions for setting up the networking architectures we describe. This bit is under 'Installation', but I think the implementation details would need to go under 'Administration'. Perhaps we can merge this as-is, and create a new issue to document implementation? It's going to take a bit more effort, but I think it's probably worth it.
> 
> The thing is this is entirely up to user and they have vastly different networking setup. I do not think having page how to setup such networking will do any good, we just need to document what is absolutely required in this three different scenarios. Compare it to the Cobbler where we just states that there must be a dhcp server with pxe enabled - https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/cobbler.html#cobbler-configure

I take your point, but there is actually a procedure in the Cobbler example. Leave it with me, I'll come up with something.

Loquacity:
Adds a new section for network admin, designed as a stub so we can add more to it later on.

modules/retail/pages/retail-admin-network.adoc
```diff
@@ -0,0 +1,19 @@
...
+----
+next-server: <branch server IP>
+if option arch = 00:07 {
+     filename "boot/shim.efi";
+} else {
+     filename "boot/pxelinux.0";
+}
+----
+
+Your DNS servers can then resolve ``salt`` and ``ftp`` hostnames as CNAMEs to the correct branch server hostname.
```
aaannz, line 17: Not `can`, but must.
Customer need to configure their DNS servers in a way that at minimum `salt` is CNAME and resolve to the correct branch server FQDN. `ftp` should resolve to the branch as well, but can be regular A record (resolving to IP), no need for CNAME.
This is the hardest part of all of the retail to configure correctly.

If this is impossible or hard for customer to guarantee these DNS entries, there are workarounds:
* for `salt` instead of using CNAME, pass `MASTER=<branch_server_fqdn>` on kernel command line when terminal is booting up. PXE formula can be used to configure this value (see https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/salt/formula-pxe.html )
* for `ftp` see  https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/retail/retail-deploy-terminals.html#_customize_the_terminal_image_download_process about using `saltboot_download_server`

Loquacity: OK, I've added this in now.

Loquacity:
New section in the built PDF:

image

Full PDF:

[suse_manager_retail_guide.pdf](https://github.com/uyuni-project/uyuni-docs/files/6145758/suse_manager_retail_guide.pdf)

keichwa (approved):

aaannz (approved):
That looks good. Thanks

Issue #830, opened 2021-03-04
Fix link to AWS Region/Zones doc


calancha:
## Documentation Update Required

In this file `modules/client-configuration/pages/vhm-aws.adoc`
you can read:
"For more information about setting regions and zones, see xref:client-configuration:vhm.adoc[]."
but in such a link there is no discussion on Region/Zones.

According with the following screenshot, the right target for this link seems to be:
`modules/client-configuration/pages/virtualization.adoc`
(`#_susesupport_and_vm_zones`)

![right-target-Screenshot from 2021-03-04 10-05-17](https://user-images.githubusercontent.com/13642261/109939286-32f72980-7cd1-11eb-9f7b-c307aee99998.png)

## What change is required?
Fix the target of a link.

## Subject Matter Expert (SME)

Who is the subject matter expert for this topic (if known)?
Any member of the doc squad.

## Version

Which version of the software is this feature/update expected to land in (if known)?
Found in a SUMa 4.2.0.Alpha2

jcayouette:
@calancha Thanks adjusting now.

calancha:
This could be fixed another way around: instead of update the target, you can move the relevant section
to the current link (that is already discussing clouds).  I let decide doc experts what is better.
Please, do not forget to fix in the same page  `i1234567890abcdef0` -> `i-1234567890abcdef0` (not mentioned in my report).

PR #831, opened 2021-03-04
fixed aws url link, and added missing hyphen

jcayouette:
# Description

In response to:

https://github.com/uyuni-project/uyuni-docs/issues/830

# Target branches

- [X] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/830

### DIFF
--- a/modules/client-configuration/pages/vhm-aws.adoc
+++ b/modules/client-configuration/pages/vhm-aws.adoc
@@ -38,5 +38,5 @@ To see which objects and resources have been inventoried, navigate to menu:Syste
 Instances running on the Amazon public cloud report a UUID to the {productname} Server in the format of an ``i`` followed by seventeen hexadecimal digits:
 
 ----
-i-1234567890abcdef0
+I1234567890abcdef0
 ----


calancha (approved):
LGTM

calancha:
My bad, we don't need to add the dash.  What is said there is that SUMA reports the ID w/o dash; that is correct, as you
can see here:
-id-seen-from-suma-Screenshot from 2021-03-04 14-28-28

Loquacity:
> My bad, we don't need to add the dash. What is said there is that SUMA reports the ID of the instance as an UUID w/o dash; that is correct, as you can see here:
> -id-seen-from-suma-Screenshot from 2021-03-04 14-28-28

Looks like it needs to be a capital letter "I", not lowercase?

modules/client-configuration/pages/vhm-aws.adoc
```diff
@@ -38,5 +38,5 @@ To see which objects and resources have been inventoried, navigate to menu:Syste
 Instances running on the Amazon public cloud report a UUID to the {productname} Server in the format of an ``i`` followed by seventeen hexadecimal digits:
 
 ----
-i1234567890abcdef0
+i-1234567890abcdef0
```
Loquacity, line 41: Looks like this might need to be an uppercase "I", not lowercase?

calancha: That's right: in SUMA, it is shown in uppercase.

jcayouette: Fixed

PR #841, opened 2021-03-08
Fix the three Python scripts

calancha:
# Description

Fix the 3 Python scripts presented on the page.

- modules/large-deployments/pages/hub-auth.adoc:
* Fix local variable usage.
* Use the same variable name in the text than in the scripts.
* Clarify in the script for Hub manual authentication that
      each machine uses its own credentials.

# Target branches

Found during 4.2 manual test cycle. 

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

https://github.com/SUSE/spacewalk/issues/12722

### DIFF
--- a/modules/large-deployments/pages/hub-auth.adoc
+++ b/modules/large-deployments/pages/hub-auth.adoc
@@ -61,13 +61,17 @@ hubSessionKey = client.hub.login(HUB_USERNAME, HUB_PASSWORD)
 # Get the server IDs
 serverIds = client.hub.listServerIds(hubSessionKey)
 
-# Authenticate each server with its own credentials
-usernames = [HUB_LOGIN for s in serverIds]
+# For simplicity, this example assumes you are using the same username and password here, as on the hub server.
+# However, in most cases, every server has its own individual credentials.
+usernames = [HUB_USERNAME for s in serverIds]
 passwords = [HUB_PASSWORD for s in serverIds]
+
+# Each server uses the credentials set above, client.hub.attachToServers needs
+# them passed as lists with as many elements as there are servers.
 client.hub.attachToServers(hubSessionKey, serverIds, usernames, passwords)
 
 # Perform the operation
-systemsPerServer = client.multicast.system.list_systems(hubKey, serverIds)
+systemsPerServer = client.multicast.system.list_systems(hubSessionKey, serverIds)
 successfulResponses = systemsPerServer["Successful"]["Responses"]
 failedResponses = systemsPerServer["Failed"]["Responses"]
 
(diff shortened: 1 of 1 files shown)


Loquacity (approved):

modules/large-deployments/pages/hub-auth.adoc
```diff
@@ -62,12 +62,12 @@ hubSessionKey = client.hub.login(HUB_USERNAME, HUB_PASSWORD)
 serverIds = client.hub.listServerIds(hubSessionKey)
 
 # Authenticate each server with its own credentials
-usernames = [HUB_LOGIN for s in serverIds]
+usernames = [HUB_USERNAME for s in serverIds]
```
agraul, line 65: I don't really know how the hub operates and what this script needs to take care of, but this line looks a bit weird to me.
```python
usernames = [HUB_USERNAME for s in serverIds]
``` 
It assigns a list of the same string (`HUB_USERNAME`) with as many elements of that string as there are `serverIds` to `usernames`. The comment above suggest there are different passwords for each server, but that would require something dynamic (e.g. a function call) that returns the correct username and passwords for each server. 
```python
usernames = [calculate_username_for_server(s) for s in serverIds]
passwords = [calculate_password_for_server(s) for s in serverIds]
``` 
But maybe that's not what is needed. Another option to make this example a bit clearer (to me at least) would be to change the above comment to read something like the following.
```python
# Each server uses the credentials set above, client.hub.attachToServers needs
# them passed as lists with as many elements as there are servers.
```

calancha: Agreed.
The current script seems tailored to the naive test case scenario; i.e., all machines with the same username and password.
It is mentioned in the point 3. just above this script:
```Credentials for each peripheral server are provided to the attachToServers method. This performs authentication against each server’s XMLRPC API endpoint.
```
Adding your comment to the script would make this clear.

modules/large-deployments/pages/hub-auth.adoc
```diff
@@ -61,13 +61,17 @@ hubSessionKey = client.hub.login(HUB_USERNAME, HUB_PASSWORD)
 # Get the server IDs
 serverIds = client.hub.listServerIds(hubSessionKey)
 
-# Authenticate each server with its own credentials
-usernames = [HUB_LOGIN for s in serverIds]
+# For simplicity sake, assumed here the same username/password as the hub server.
```
Loquacity, line 64:
```suggestion
# For simplicity, this example assumes you are using the same username and password here, as on the hub server.
```

modules/large-deployments/pages/hub-auth.adoc
```diff
@@ -61,13 +61,17 @@ hubSessionKey = client.hub.login(HUB_USERNAME, HUB_PASSWORD)
...
+# In the general case, every server will have its own credentials.
```
Loquacity, line 65:
```suggestion
# However, in most cases, every server has its own individual credentials.
```

Loquacity (changes requested):

Loquacity (approved):

jcayouette (approved):

agraul (approved):

PR #861, opened 2021-03-18
adds pam.d/common-auth file

Loquacity:
# Description

The previous instructions didn't set up the common-auth file in PAM config, so we weren't adding site specific pam modules. Without it, PAM can only auth local connections.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/11936

### DIFF
--- a/modules/administration/pages/auth-methods-pam.adoc
+++ b/modules/administration/pages/auth-methods-pam.adoc
@@ -18,7 +18,7 @@ account  include        common-account
 password include        common-password
 session  include        common-session
 ----
-. Ont he {productname} Server, at the command prompt, as root, add the ``sss`` PAM module:
+. On the {productname} Server, at the command prompt, as root, add the ``sss`` PAM module:
 +
 ----
 pam-config -a --sss


modules/administration/pages/auth-methods-pam.adoc
```diff
@@ -20,18 +18,25 @@ account  include        common-account
 password include        common-password
 session  include        common-session
 ----
+. Edit the ``/etc/pam.d/common-auth`` file to add the required PAM modules:
```
Firstyear, line 21: Hmmm, I'd probably shy away from "how to edit" the common-auth file here. Reading the "example" config you have here, I think it may actually have some issues and could raise security risks. Pam is .... horrendous to get right. 

If you *must* have an example I would recommend:

```
auth        required      pam_env.so
auth        [default=1 ignore=ignore success=ok] pam_localuser.so
auth        sufficient    pam_unix.so nullok try_first_pass
auth        requisite     pam_succeed_if.so uid >= 1000 quiet_success
auth        sufficient    pam_sss.so ignore_unknown_user
auth        required      pam_deny.so
```

pam_sss can handle krb OR ldap so that can solve that use case.

Also worth noting that you need to check /etc/pam.d/common-auth is NOT a symlink to /etc/pam.d/common-auth-pc, because the next zypper update will probably nuke your config from audit if you don't remove common-auth as a symlink and turn it into a real file.

Perhaps a "safe" alternative is:

```
pam-config -a --sss
```

Which will do the reconfiguration to add pam_sss for you. Editing pam by hand is dangerous and can result in your being completely locked out of your machine, so I really really really advise you avoid "how to edit" common-auth here ...

Loquacity: OK, changed to the `pam-config` command, and added a little note about not editing by hand. Improvement?

modules/administration/pages/auth-methods-pam.adoc
```diff
@@ -20,18 +18,25 @@ account  include        common-account
 password include        common-password
 session  include        common-session
 ----
+. Ont he {productname} Server, at the command prompt, as root, add the ``sss`` PAM module:
```
Firstyear, line 21: Typo :)

Firstyear (approved):
Looks good 🎉

modules/administration/pages/auth-methods-pam.adoc
Loquacity: grrrrrr

Loquacity:
> Looks good

Yay! Thanks for the help. I owe you $beverage_of_choice :beers:

Firstyear:
> > Looks good
> 
> Yay! Thanks for the help. I owe you $beverage_of_choice 🍻

Happy to help :)

jcayouette (approved):
LGTM, thanks.

PR #863, opened 2021-03-18
Replace Livecycle with Lifecycle

calancha:
# Description

Replace Livecycle with Lifecycle
# Target branches

Found during manual test cycle for 4.2.0 Beta 1; I haven't checked other branches.

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links
https://github.com/SUSE/spacewalk/issues/12722

### DIFF
--- a/modules/administration/pages/live-patching-channel-setup.adoc
+++ b/modules/administration/pages/live-patching-channel-setup.adoc
@@ -8,7 +8,7 @@ Clients using live patching have updates for the running kernel in the live patc
 There are two ways to manage channels for live patching:
 
 Use content lifecycle management to clone the product tree and remove kernel versions newer than the running one.
-This procedure is explained in the xref:content-lifecycle-examples.adoc#enhance-project-with-livepatching[Content Lifecycle Management Examples].
+This procedure is explained in the xref:content-lifecycle-examples.adoc#enhance-project-with-livepatching[].
 This is the recommended solution.
 
 Alternatively, use the `spacewalk-manage-channel-lifecycle` tool.


Loquacity:
The word is wrong, sure, but we shouldn't have anything at all between the square brackets in an xref.

calancha:
> The word is wrong, sure, but we shouldn't have anything at all between the square brackets in an xref.
Dropped the sentence inside brckets.

Sometimes helps the reader: in cases when the file name doesn't describe well the context
of what we are documenting.  There are several good examples elsewhere.
For instance see at `modules/retail/pages/retail-migration-slepostosuma.adoc`:
```
For more information about using an activation key to assign the required channels, see xref:retail-install.adoc#retail.sect.install.install.config[Configuring Server].
```
To me, showing `Configuring Server` looks nicer than an ugly file path.

Loquacity:
> > The word is wrong, sure, but we shouldn't have anything at all between the square brackets in an xref.
> > Dropped the sentence inside brckets.
> 
> Sometimes helps the reader: in cases when the file name doesn't describe well the context
> of what we are documenting. There are several good examples elsewhere.
> For instance see at `modules/retail/pages/retail-migration-slepostosuma.adoc`:
> 
> ```
> For more information about using an activation key to assign the required channels, see xref:retail-install.adoc#retail.sect.install.install.config[Configuring Server].
> ```
> 
> To me, showing `Configuring Server` looks nicer than an ugly file path.

I quite agree, but it doesn't work when we build PDF. For PDF, we need Antora to show the book and section instead of a nice name so that people can find the correct document without being able to click. The most consistent way to do that is to let it auto-generate.

Loquacity (approved):

Issue #848, opened 2021-03-11
Supported Debian features menu entry


calancha:
## Documentation Update Required

Debian is not listed here:
- modules/ROOT/pages/index.adoc

We only list the Debian supported features menu entry for Uyuni;
```
ifeval::[{uyuni-content} == true]
*** xref:supported-features-debian.adoc[Debian Supported Features]
endif::[]
```

But we unconditionally list the Debian client registration menu entry.

Should we drop the `ifeval` above?

## What does the documentation say now?

- Debian is not mentioned in this list:
![supported-oses-Screenshot from 2021-03-11 14-03-08](https://user-images.githubusercontent.com/13642261/110791670-c2bd4a80-8272-11eb-953c-811565bb1acd.png)

- Missing menu entry for Debian supported features.
![missing-debian-entry-Screenshot from 2021-03-11 14-03-44](https://user-images.githubusercontent.com/13642261/110791629-b6d18880-8272-11eb-913e-63287a111720.png)

## What change is required?
Include Debian in the above list and the Debian supported features menu entry.

## Subject Matter Expert (SME)

Who is the subject matter expert for this topic (if known)?
Any doc squad member will do.

## Version

Which version of the software is this feature/update expected to land in (if known)?
Found in the documentation of SUMa 4.2.0 Beta1.
Which version of the documentation must this feature/update land in (if known)?
Does the content need to be backported to previous documentation versions (if known)?
In case the Debian support starts in 4.2 version, then the answers is no.  Otherwise,
we should backport to any version that supports Debian.

## Internal SUSE Reporters

This repository is public.
If you are reporting an issue that needs to remain company confidential, please use the internal Bugzilla instance instead.

juliogonzalez:
@keichwa seems this part was done, but table 1 still does NOT mention Debian at neither 4.1 or 4.2. Reopening.

keichwa:
it is done.

PR #869, opened 2021-03-29
remove conditionals from client cfg

Loquacity:
# Description

When we added support for Debian, we only removed some of the conditionals, this PR removes the rest, and also adds Debian to the list of supported clients in the index. Additionally, removes the erroneous conditional from config mgmt. 

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/848


jcayouette (approved):

juliogonzalez (approved):
LGTM. Remember to backport whatever is needed.

PR #874, opened 2021-03-30
4.1 only remove 15.1 ref

Loquacity:
~***DO NOT MERGE. DOCS REPO IS FROZEN***~

# Description

Removes 15.1 reference from 4.1 docs.

# Target branches

***This is not a backport, it's an edit to the 4.1 docs only***

- [ ] Master (Default)
- [x] Manager-4.1
- [ ] Manager-4.0

# Links

Partial https://github.com/SUSE/spacewalk/issues/14417

### DIFF
--- a/modules/client-configuration/pages/clients-opensuse.adoc
+++ b/modules/client-configuration/pages/clients-opensuse.adoc
@@ -47,12 +47,6 @@ ifeval::[{uyuni-content} == true]
 | Updates Channel
 | Other Channels
 
-| {opensuse} Leap 15.1
-| opensuse_leap15_1
-| opensuse_leap15_1-uyuni-client
-| opensuse_leap15_1-updates
-| opensuse_leap15_1-non-oss and opensuse_leap15_1-non-oss-updates
-
 | {opensuse} Leap 15.2
 | opensuse_leap15_2
 | opensuse_leap15_2-uyuni-client
--- a/modules/client-configuration/pages/supported-features-sles.adoc
+++ b/modules/client-configuration/pages/supported-features-sles.adoc
@@ -29,7 +29,7 @@ The icons in this table indicate:
 | {sle}{nbsp}11
 | {sle}{nbsp}12
 | {sle}{nbsp}15
-| {opensuse}{nbsp}15.2
+| {opensuse}{nbsp}15
 
 | Client
 | {check}


jcayouette (approved):

juliogonzalez (changes requested):
Not sure you want to do it this way. If you just remove `.1` it looks like we support openSUSE Leap 15.0 and 15.1, and that's not correct. Maybe you should check with @paususe what's the best approach here.

Also there are mentions to `15.1` at other places at the doc :-)

Loquacity:
> Also there are mentions to `15.1` at other places at the doc :-)

Do we need to remove it from https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/clients-opensuse.html ?

The other mentions are all in reference to upgrading, so I think they're fine.

juliogonzalez:
> > Also there are mentions to `15.1` at other places at the doc :-)
> 
> Do we need to remove it from https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/clients-opensuse.html ?
> 
> The other mentions are all in reference to upgrading, so I think they're fine.

Upgrading is one thing, but at the start of that page it mentions that you can add Leap 15.1. You can, but you should not do it anymore, because it's not supported. So I'd remove it.

And remember to review 4.0 and master as well. They have the mention to 15.1 not exactly at the same places where you have it for 4.1

For example:
https://documentation.suse.com/external-tree/en-us/suma/4.0/suse-manager/client-configuration/supported-features.html
vs
https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/client-configuration/supported-features.html

keichwa:
Yes, in the Uyuni context, I'd also not remove it.  Better just add warnings where appropriate that upstream (the opensuse community) no longer maintains 15.1, and upgrading is recommended.

Loquacity:
Can I get reviews on this please?

modules/salt/pages/formula-saltboot.adoc
```diff
@@ -120,7 +120,7 @@ In this example, SATA disks are differentiated from USB disks:
 /dev/disk/by-path/*usb*
 ----
 
-If the entered value does not contain "[systemitem]``/``", the entered value is automatically prepended by "[path]``/dev/disk/by-path/``".
+If the entered value does not contain "[systemitem]``/``, the entered value is automatically prepended by "[path]``/dev/disk/by-path/``."
```
keichwa, line 123: I do not understand why we want to have the double quotes here.

juliogonzalez (commented):
Strictly speaking I didn't request the removal of 15.1, but at least adding 15.2.

15.1 went EoL on Feb 2nd 2021, so not sure if we want to keep it documented for a while longer. Please check with Pau.

What's clear is that we need to add 15.2 everywhere (and soon 15.3).

But if we decide to remove 15.1: `grep` gives me more results, for example at `modules/client-configuration/pages/clients-opensuse.adoc`. So in that case have a look with `grep` and change the rest of the results as needed.

keichwa (commented):
Asciidoc-wise it looks fine.  SP migration paths are to be confirmed by Pau or Julio.  I'd have to try it to see what's configured in SCC.

modules/salt/pages/formula-saltboot.adoc
Loquacity: I edited it to make vale happy, but you're right, they probably should go entirely.

Loquacity, line 123:
```suggestion
If the entered value does not contain [systemitem]``/``, the entered value is automatically prepended by [path]``/dev/disk/by-path/``.
```

Loquacity:
> Strictly speaking I didn't request the removal of 15.1, but at least adding 15.2.
> 
> 15.1 went EoL on Feb 2nd 2021, so not sure if we want to keep it documented for a while longer. Please check with Pau.
> 
> What's clear is that we need to add 15.2 everywhere (and soon 15.3).
> 
> But if we decide to remove 15.1: `grep` gives me more results, for example at `modules/client-configuration/pages/clients-opensuse.adoc`. So in that case have a look with `grep` and change the rest of the results as needed.

Yeah, that's what I was asking earlier. Do we need to remove it from modules/client-configuration/pages/clients-opensuse.adoc or just add a note saying it's not supported?

keichwa (approved):

juliogonzalez:
> Yeah, that's what I was asking earlier. Do we need to remove it from modules/client-configuration/pages/clients-opensuse.adoc or just add a note saying it's not supported?

Talk to Pau, but FMPOV if you removed it from the support matrix it should not be there.

If you keep it at `modules/client-configuration/pages/clients-opensuse.adoc` then probably the support matrix should just say "openSUSE Leap 15.X", and there must be a note that says something like "Support is available only for versions supported by openSUSE, and maybe a link to https://en.opensuse.org/Lifetime

keichwa:

> If you keep it at
> `modules/client-configuration/pages/clients-opensuse.adoc` then
> probably the support matrix should just say "openSUSE Leap 15.X", and
> there must be a note that says something like "Support is available
> only for versions supported by openSUSE, and maybe a link to
> https://en.opensuse.org/Lifetime

Referencing the Lifetime page would be a step into the right direction.
In the context of opensuse, we should avoid the plain term "support" or
"supported".  We could say "SUSE Manager supports registering client
XYZ" and then, "see https:// for versions that are maintained".  In the
Uyuni context, "with Uyuni, you can onboard client XYZ".

I hope you see what I mean ;)

-- 
Karl Eichwalder

juliogonzalez:
> I hope you see what I mean ;)

Clear for Uyuni, but not so clear for SUSE Manager, as openSUSE comes from SCC, so my guess is that it's indeed supported. But not my call :-D

modules/client-configuration/pages/client-upgrades-sp-migration.adoc
```diff
@@ -7,7 +7,6 @@ For example, you can migrate {sle}{nbsp}15{nbsp}SP1 to {sle}{nbsp}15{nbsp}SP2.
 You can also migrate openSUSE Leap to a later minor version or to the corresponding SLE SP level, for example:
 
 * openSUSE Leap 15.1 to 15.2, or
-* openSUSE Leap 15.1 to {sle} 15 SP1, or
 * openSUSE Leap 15.2 to {sle} 15 SP2
```
juliogonzalez, line 10: I am not sure if SP migration from Leap 15.1 to SLE15SP1 is still supported or not, as SLE15SP1 is not EoL (but LTSS). Please check with @paususe

modules/client-configuration/pages/clients-opensuse.adoc
```diff
@@ -25,9 +25,6 @@ ifeval::[{suma-content} == true]
 | OS Version
 | Product Name
 
-| {opensuse} Leap 15.1
-| openSUSE Leap 15.1 x86_64
-
```
juliogonzalez, line 30: What about lines 50-54 below?

modules/client-configuration/pages/client-upgrades-sp-migration.adoc
paususe: Yes, this is supported.

image

paususe (changes requested):
Why don't we just say "openSUSE Leap 15" instead of 15.1, 15.2, etc? That's what we do for SLES, RHEL, CentOS, etc: we say "SLES 15", not "SLES 15 SP1", "SLES 15 SP2", etc.

IIRC we already have a support statement saying "only operating systems under general support by the vendor are supported by SUSE Manager".

juliogonzalez:
> Why don't we just say "openSUSE Leap 15" instead of 15.1, 15.2, etc? That's what we do for SLES, RHEL, CentOS, etc: we say "SLES 15", not "SLES 15 SP1", "SLES 15 SP2", etc.

Makes sense, but doc needs to think about how to tell about the channel names and channel labels, of course. Not sure of what we are doing for SLE15 :-)

Loquacity:
> > Why don't we just say "openSUSE Leap 15" instead of 15.1, 15.2, etc? That's what we do for SLES, RHEL, CentOS, etc: we say "SLES 15", not "SLES 15 SP1", "SLES 15 SP2", etc.
> 
> Makes sense, but doc needs to think about how to tell about the channel names and channel labels, of course. Not sure of what we are doing for SLE15 :-)

That's what I did in the first instance, but then Julio pointed out that we need to be specific because 15.1 isn't supported.

juliogonzalez:
> That's what I did in the first instance, but then Julio pointed out that we need to be specific because 15.1 isn't supported.

I guess you mean https://github.com/uyuni-project/uyuni-docs/pull/874#pullrequestreview-623974724

True. I didn't remember we have the note about "We only support not EoL'ed" OS. Sorry about that :\

Still, you need to think how to explain the stuff for the channels, of course. Don't remember how we explain it for SLE, but I guess you can do the same

juliogonzalez (approved):
Looks good.

modules/client-configuration/pages/clients-opensuse.adoc
```diff
@@ -50,12 +47,6 @@ ifeval::[{uyuni-content} == true]
 | Updates Channel
 | Other Channels
 
-| {opensuse} Leap 15.1
-| opensuse_leap15_1
-| opensuse_leap15_1-uyuni-client
-| opensuse_leap15_1-updates
-| opensuse_leap15_1-non-oss and opensuse_leap15_1-non-oss-updates
-
```
paususe, line 58: If we are going to treat 15.1 and 15.2 as different versions, then it's OK to remove 15.1 but then we need to add 15.3.

I would just say "Leap 15" and that's it. In the end, even the client tools are for Leap 15 GA and they will work on Leap 15, Leap 15.1, Leap 15.2 and Leap 15.3.

juliogonzalez: No, you can't say `leap15` for the channel labels. The channel labels are different. Not just for the client tools (even if the URL for the repo is the same), but most importantly for the channels that come from openSUSE.

Another story is if we say `leap15_X` and tell users to replace "X" with the correct revision. That would work.

Loquacity: I think for the channel labels we need to be specific, that's why we created the tables. For the support matrices, etc, being more vague is OK.

keichwa (commented):
I no longer can follow this.  I hope it is still clear that with the SP Migration feature the following is technically possible or not possible:

* you cannot migrate or upgrade Leap 15.0 to anything else (AFAIK, SP Migration is available starting from 15.1)
* starting from 15.1 you can upgrade to any later 15 minor version, sometimes even with skipping minor versions
* starting from 15.1 you can migrate to a matching SLE SP version

modules/client-configuration/pages/clients-opensuse.adoc
paususe: OK

paususe:
> I no longer can follow this. I hope it is still clear that with the SP Migration feature the following is technically possible or not possible:
> 
>     * you cannot migrate or upgrade Leap 15.0 to anything else (AFAIK, SP Migration is available starting from 15.1)
> 
>     * starting from Leap 15.1 you can upgrade to any later Leap 15 minor version, sometimes even with skipping minor versions
> 
>     * starting from Leap 15.1 you can migrate to a matching SLE SP version

That's 100% right

PR #904, opened 2021-04-16
Cookbook - CLM

Loquacity:
# Description

Adds a CLM recipe to the cookbook

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

- Partial https://github.com/SUSE/spacewalk/issues/14304
- Epic https://github.com/SUSE/spacewalk/issues/12021

### DIFF
--- a/modules/cookbook/nav-cookbook.adoc
+++ b/modules/cookbook/nav-cookbook.adoc
@@ -10,8 +10,8 @@ endif::[]
 
 
 * xref:cookbook-overview.adoc[Cookbook]
-** xref:recipe-onboarding.adoc[Recipe - Onboarding]
-** xref:recipe-clm.adoc[Recipe - CLM]
+** xref:recipe-onboarding.adoc[Onboarding]
+** xref:recipe-clm.adoc[Content Lifecycle Management]
 
 // License
 ifdef::backend-pdf[]
--- a/modules/cookbook/pages/recipe-clm.adoc
+++ b/modules/cookbook/pages/recipe-clm.adoc
@@ -33,7 +33,7 @@ Before you start, you should already have:
 
 == Steps
 
-Create a CLM project::
+Create a new CLM project::
 . In the {productname} {webui}, navigate to menu:Content Lifecycle[Projects], and click btn:[Create Project].
   Type a name, label, and description for your project, and click btn:[Create].
 +


modules/cookbook/nav-cookbook.adoc
```diff
@@ -10,8 +10,8 @@ endif::[]
 
 
 * xref:cookbook-overview.adoc[Cookbook]
-** xref:recipe1.adoc[Recipe 1]
-** xref:receipe2.adoc[Recipe 2]
+** xref:recipe-onboarding.adoc[Recipe - Onboarding]
```
keichwa, line 13: I think short dashes in titles are a little bit Unamerican.  Maybe, we'd better use colons?

modules/cookbook/pages/recipe-clm.adoc
```diff
@@ -0,0 +1,81 @@
+[[cookbook-clm]]
+= Content Lifecycle Management
+
+If you are managing a lot of clients and you need to apply customized packages to them, you can use CLM to manage your packages.
```
keichwa, line 4:
```suggestion
If you are managing a lot of clients and you need to apply customized packages to them, you can use content lifecycle management (CLM) to manage your packages.
```

modules/cookbook/pages/recipe-clm.adoc
```diff
@@ -0,0 +1,81 @@
...
+Content lifecycle management (CLM) allows you to customize and test packages before updating production clients.
```
keichwa, line 5:
```suggestion
CLM allows you to customize and test packages before updating production clients.
```

modules/cookbook/pages/recipe-clm.adoc
```diff
@@ -0,0 +1,81 @@
...
+It is also useful if you need to apply updates during a limited maintenance window.
+
+
+
+== Problem
+
+Content lifecycle management allows you to select software channels as sources, adjust them as required for your environment, and thoroughly test them before installing onto your production clients.
```
keichwa, line 12: And now the question is whether to use the abbrev always or not always ;)

keichwa (approved):
I'd avoid some "your" words (such as "your production clients"), but I know that people often think different.

modules/cookbook/nav-cookbook.adoc
Loquacity: I don't even know if we need to prefix each with 'recipe' actually, considering that's all that's really in this book.

modules/cookbook/pages/recipe-clm.adoc
Loquacity: Spell it out the first time in each section, than abbrev after that?

keichwa: sounds like a plan :)

keichwa (approved):

Loquacity:
> I'd avoid some "your" words (such as "your production clients"), but I know that people often think different.

We have a tendency to use this in our docs generally (although not too much), but I think in this guide especially, it's OK. The intention of this guide was always to be slightly less stiff than the rest of our docs, anyway. That's also something we know from surveys that our readers want.

PR #908, opened 2021-04-18
Generate language selector only for built languages

paususe:
# Description

Generates HTML language selector only for languages enabled in Makefile.

It's not too nice at the moment because we still need to hardcode the list of languages in the Makefile but at least we no longer need to modify the branding manually. I don't think the hardcoding can be solved until the Makefiles are rewritten in some kind of 2-stage build system (e. g. CMake).

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


jcayouette (approved):
Looks good, I would like to test it :)

PR #912, opened 2021-04-18
Unbreak failing PDF generation due to bad locale

paususe:
# Description

PDF generation is failing because we use special characters on the front page of CJK languages. By default, the GH Actions runners do not enable UTF-8, so this PR sets some locale with UTF-8 (de_DE.UTF-8) to that UTF-8 is enabled and CJK does not cause an error.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


jcayouette (approved):
Awesome, nice work! Thanks @paususe

PR #913, opened 2021-04-19
Menu: use icons instead of words

paususe:
# Description

Use icon instead of the word "Language" to select language. This is a lot better because otherwise in e. g. Chinese, it says "Language" in English while the rest of the page is in Chinese.

Same for the word "Resources".

To avoid licensing issues, the icons are coming from EOS Project.

Before:

image

After:

image

# Target branches

- [x] Master (Default)
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


Loquacity (approved):
I really like this!

paususe:
@Loquacity I went ahead and replaced "Resources" with another icon. I think it's a lot better now (and makes translations no longer a problem, which was my actual reason to do this).

keichwa (approved):

Issue #888, opened 2021-04-08
Documentation on `configure-tftpsync.sh`


calancha:
## Documentation Update Required

Seen at `modules/client-configuration/pages/cobbler.adoc`,
where it is documented `Procedure: Configuring TFTP on a Proxy`.
```
. On the {productname} Proxy, as root, run the [command]``configure-tftpsync.sh`` script.
. The script interactively asks you for details on the host names and IP addresses of the {productname} Server and Proxy, as well for the location of the [path]``tftpboot`` directory on the Proxy.

For more information, use the [command]``configure-tftpsync.sh --help`` command.
```

I see two issues:
- `--help` is not a valid option for such a script.
- There is no `more information` to see when calling the command w/o arguments (just the signature).
```
$ configure-tftpsync.sh --help
Invalid hostname for proxy: --help
usage: configure-tftpsync.sh proxy1.domain.top proxy2.domain.top ...
```
- The text mentions that the script is interactive, but it's not.

Maybe that description has been copied and pasted from some other similar command.

## What does the documentation say now?
It says that the script is interactive, and it accepts a `--help` option.

## What change is required?
The script is not interactive and it doesn't accept a `help` option.
Instead of refer to the script help, we might add any useful information in the text.

## Subject Matter Expert (SME)

N/A
## Version

Found in the master branch of the SUMa documentation.

Bischoff:
The ``help`` option works, see screenshot.

Bischoff:
![image](https://user-images.githubusercontent.com/1932575/116240435-1c66cf00-a764-11eb-93c7-01d5d8974fd2.png)

Bischoff:
It might also be the interactive mode works. At least it should, according to this help page.

0rnela:
Closing this issue as the entire chapter was rewritten and also @Bischoff had provided screenshot for the --help option.

PR #917, opened 2021-04-21
Rewritten autoinstallation chapter of client configuration guide

Bischoff:
# Description

Deprecated- see #936

PR #919, opened 2021-04-22
General updates and edits to the Cobbler chapter.

Loquacity:
# Description

Fixes an error in the `configure-tftpsync.sh --help` command, plus general updates around wording, syntax, whitespace, etc.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/uyuni-project/uyuni-docs/issues/888

### DIFF
--- a/modules/client-configuration/pages/cobbler.adoc
+++ b/modules/client-configuration/pages/cobbler.adoc
@@ -124,7 +124,7 @@ virsh net-dumpxml --inactive network > network.xml
 ----
 virsh net-define network.xml
 ----
-
++
 Alternatively, you can use the [command]``net-edit`` subcommand, which also performs some error checking.
 
 .Example: Minimal Network XML Description for KVM
@@ -406,7 +406,7 @@ Bare metal systems management can be enabled or disabled in the {productname} {w
 
 [NOTE]
 ====
-New systems are added to the organization of the Administrator who enabled the bare metal systems management feature.
+New systems are added to the organization of the administrator who enabled the bare metal systems management feature.
 To change the organization, log in as an Administrator of the required organization, and re-enable the feature.
 ====
 


modules/client-configuration/pages/cobbler.adoc
```diff
@@ -184,15 +181,12 @@ zypper install susemanager-tftpsync-recv
 
 
 .Procedure: Configuring TFTP on a Proxy
-
 . On the {productname} Proxy, as root, run the [command]``configure-tftpsync.sh`` script.
-. The script interactively asks you for details on the host names and IP addresses of the {productname} Server and Proxy, as well for the location of the [path]``tftpboot`` directory on the Proxy.
+. The script requires the host names and IP addresses of the {productname} Server and Proxy, as well the location of the [path]``tftpboot`` directory on the Proxy.
```
calancha, line 185: I am not an expert on this topic, but looking at the script it seems only requires Proxies IP's (not Server).
This makes sense because you are calling the script from the server.
The script includes the following help:
```
print_help () {
        cat <<HELP
usage: configure-tftpsync.sh proxy1.domain.top proxy2.domain.top ...

HELP
exit 1
}
```

Loquacity: Updated. How does one call that help text, if not with `configure-tftpsync.sh --help` ?

calancha: Likely the author of the script forgot to support the `--help` option.
It is trivial to do it, so maybe we could just support it as in other scripts.

Loquacity: Ah, that's annoying. What should the docs say in this case?

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -126,7 +124,7 @@ virsh net-dumpxml --inactive network > network.xml
 ----
 virsh net-define network.xml
 ----
-+
+
```
keichwa, line 127:
```suggestion
+
```
adoc needs the plus to make the following sentence part of the last step of the procedure.

modules/client-configuration/pages/cobbler.adoc
```diff
@@ -421,7 +406,7 @@ Bare metal systems management can be enabled or disabled in the {productname} {w
 
 [NOTE]
 ====
-New systems are added to the organization of the administrator who enabled the bare metal systems management feature.
+New systems are added to the organization of the Administrator who enabled the bare metal systems management feature.
```
keichwa, line 409: I think we'd better down-case Administrator everywhere.

keichwa (approved):

modules/client-configuration/pages/cobbler.adoc
Loquacity: Might need to go through some other docs as well, in that case.

Loquacity, line 409:
```suggestion
New systems are added to the organization of the administrator who enabled the bare metal systems management feature.
```

Loquacity:
```suggestion
New systems are added to the organization of the administrator who enabled the bare metal systems management feature.
```

keichwa (approved):

modules/client-configuration/pages/cobbler.adoc
keichwa: In this case, we probably better do it later.

Bischoff:
Beware this will get superseded by #917

Bischoff (changes requested):
configure-tftpsync.sh just works, I tested

modules/client-configuration/pages/cobbler.adoc
Bischoff: This option is supported.

Loquacity:
@Bischoff Perhaps we should just delete this PR without merging, then?

calancha:
Let's closed it without merging it.  I was calling a script with the same name from the _server_.
As Eric has pointed out, a script with the same name exists in the proxy and it behaves as documented.

Bischoff:
Yes, this is very error-prone. I was even considering opening a bug for such bad practice (naming identically two different scripts).

PR #943, opened 2021-05-05
SUSE branding: Switch to using Poppins from doc.suse.com

ghost:
# Description

If there are no overriding reasons why we need Google Fonts for this, we can just use the version of the font we (will) use everywhere else on the site as well. Our fonts will tend to be cached on people's disks already and they are much smaller (below 50 KB), because they do not include the Indian character sets that Poppins normally includes--though the effective file size of even Google's version is not the full 1-MB+, because WOFF fonts can be streamed. In terms of web compatibility, it should essentially be the same as Google's version.

(If you do notice any relevant missing characters, I can 

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0


keichwa (approved):

keichwa (approved):

keichwa:
I updated the files of the MR in gitlab for external tree.

PR #952, opened 2021-05-11
use Product Migration instead of SP Migration

keichwa:
https://github.com/SUSE/spacewalk/issues/14491

# Description

See title.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Partially fixes https://github.com/SUSE/spacewalk/issues/14491


modules/client-configuration/pages/client-upgrades-sp-migration.adoc
```diff
@@ -1,7 +1,7 @@
 [[client-upgrades-spmigration]]
-= Service Pack Migration
```
parlt91, line 2: If it does not mean to much trouble. Can we also rename the file to something like `client-upgrated-product-migration`

modules/reference/pages/systems/system-details/sd-sp-migration.adoc
```diff
@@ -1,7 +1,8 @@
 [[ref-systems-sd-migration]]
-= SP Migration
```
parlt91, line 2: Same remark about the file name as above

parlt91 (commented):
Thank you for taking care of this :smiley: 
I left some remarks below

modules/reference/pages/systems/system-details/sd-sp-migration.adoc
keichwa: yes, I'll do it.   I'll commit it separately.

PR #960, opened 2021-05-18
Fix suma API URL.

keichwa:
# Description

https://github.com/SUSE/spacewalk/issues/14903 - partial fix
orthography normalization (drive-by edit)

# Target branches

- [ ] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -6,3 +6,4 @@
 
 # Last packaged: 2021-05-14 10:00
 
+- Fixed URL of API documentation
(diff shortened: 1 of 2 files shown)


juliogonzalez (approved):
Same as the other PR. LGTM, but check if there are other links to the API at the doc, just in case :-)

.changelog
```diff
@@ -6,3 +6,4 @@
 
 # Last packaged: 2021-05-14 10:00
 
+- Fixed URL of API documentation
```
juliogonzalez, line 9: Add the bsc.

juliogonzalez (changes requested):

.changelog
juliogonzalez: Ignore. I didn't create one. Approving again.

juliogonzalez (approved):
As long as you checked that this the only place where we link the API doc, LGTM.

keichwa:
> Same as the other PR. LGTM, but check if there are other links to the API at the doc, just in case :-)

this seems to be the only place.

PR #961, opened 2021-05-18
Fix suma API URL (master)

keichwa:
# Description
orthography normalization (drive-by edit)
Also see https://github.com/uyuni-project/uyuni-docs/pull/960

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/14903

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -7,3 +7,47 @@
 # Last packaged: 2021-05-17
 
 - Fixed URL of API documentation
+
+# Last packaged: 19-04-2021
+
+- Documented Prometheus TLS support in the Administration Guide and the
+  Salt Guide.
+- Fixed certificate names in Red Hat Client Registration chapter of the
+  Client Configuration Guide
+- Use the term "Product Migration" instead of "SP Migration".
+- Documented retracted patches in Administration Guide.
+- Documented filter templates for Content Lifecycle Management in
+  Administration Guide 
+- Documented Ansible Feature in Administration and Reference Guide 
+- Completely rewritten Autoinstallation sections in the Client Configuration
+  Guide.
+- Update OpenSCAP section in the Administration Guide for usability.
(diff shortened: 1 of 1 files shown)


juliogonzalez (approved):
LGTM. Before merging, verify if we have a link to the API somewhere else, just in case :-)

.changelog
```diff
@@ -4,46 +4,6 @@
...
-- Adds instructions to delete vendor channels in
-  Administration Guide (bsc#1183293)
-- Adds troubleshooting entry to Administration Guide for
-  registering traditional RHEL 7 clients (bsc#1182700)
-- Adds missing Virtualization Network tab to System Details
-  section in Reference Guide
-- Adds SLE Micro and openSUSE MicroOS clients in the Client Configuration Guide
-- Adds troubleshooting multiple Red Hat CDN certificates to Client
-  Configuration Guide (bsc#1184448)
+- Fixed URL of API documentation
```
juliogonzalez, line 9: Sorry, not OK, you are missing the bsc :-)

juliogonzalez (changes requested):

.changelog
keichwa: bsc for the API URL?  there is non, afaics.  Or is it about the other changelog entries?  We have the (bad) habit to reset the .changelog file when everything is transferred to the .changes of the package(s).  But I do not mind keeping the complete .changelog here.

juliogonzalez: Ah, crap. I only created a github card :-D

Ignore my comment.

juliogonzalez (approved):
As long as you checked that this the only place where we link the API doc, LGTM.

.changelog
keichwa: ok ;)

PR #962, opened 2021-05-18
Clientsupport #14905 keichwa

keichwa:
# Description

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [x] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/14905


juliogonzalez (commented):
Remember we also have mentions to 16.04 at the "Old ubuntu" page, and to RHEL6 and clones at the RHEL/CentOS/Oracle/etc guides.

But I guess that removing the mentions there, or just adding admonitions is for @paususe to decide. So he needs to be the reviewer of this PR :-)

modules/installation/pages/client-requirements.adoc
```diff
@@ -120,14 +100,12 @@ For non-{suse} operating systems, including {rhel}, {centos}, and {oracle}, only
...
-| {x86_64}
-| {cross}
-| {check}
-
 |===
 
-
+[NOTE]
+====
+{rhel} 6, {oracle} 6, CentOS 6, {sles}-ES 6, and {ubuntu} 16.04 are end-of-life upstream and no longer supported by {suse} as client operating systems.
```
paususe, line 107:
```suggestion
{rhel} 6, {oracle} 6, CentOS 6, {sles}-ES 6, and {ubuntu} 16.04 are end-of-life upstream and no longer supported by {suse} as client operating systems. SUSE Manager Client Tools for these operating systems are still provided from SCC for convenience.
```

paususe (changes requested):
At least in 4.1 (and maybe 4.2 too, why not), I would mention we still provide the client tools, since those operating systems were supported when we shipped 4.1 GA.

PR #973, opened 2021-05-25
In Supported Clients Table of Client ConfigGuide list Debian 10 and 9

keichwa:
# Description

Cf. https://github.com/uyuni-project/uyuni-docs/issues/848

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [x] Manager-4.1
- [ ] Manager-4.0

# Links


juliogonzalez (approved):
Ok, if the other tables contain Debian as well (I think the they did, but I'd suggest you check, just in case).

Then backport to 4.1, as this was reported there as well (FMPOV we should prepare a resubmission for 4.1.8 with this)

juliogonzalez (changes requested):
I think you will need https://bugzilla.suse.com/show_bug.cgi?id=1186427 at the changelog.

.changelog
```diff
@@ -4,6 +4,9 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- In the Supported Clients Table of the Client Configuration Guide, list
+Debian 10 and 9 unconditionally (also in the context of SUSE Manager)
```
keichwa, line 8:
```suggestion
Debian 10 and 9 unconditionally also in the context of SUSE Manager
(bsc#1186427)
```

juliogonzalez (commented):
Ok, if the other tables contain Debian as well (I think the they did, but I'd suggest you check, just in case).

Then backport to 4.1, as this was reported there as well (FMPOV we should prepare a resubmission for 4.1.8 with this)

keichwa:
> Then backport to 4.1, as this was reported there as well (FMPOV we should prepare a resubmission for 4.1.8 with this)

I think a resubmission because of an outdated support table is not need.  I guess and hope the release notes and marketing product info are up-to-date. At least, I already files a backport PR for `manager-4.1`.

PR #984, opened 2021-06-07
Migrate transfer org #14846 keichwa

keichwa:
# Description

transfer between organizations was previously called "migrate".

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/14846


keichwa:
@renner and @paususe can we savely backport it to 4.2 now or better later for the next 4.2 MU?

@paususe can we savely rename the file in the Reference Guide?

renner (approved):
Thank you @keichwa, this looks really good to me!

I guess the `Unchanged files with check annotations` are unrelated and general suggestions from Vale? This is good, but somehow we should avoid that they are added with each PR, if possible.

keichwa:
> Thank you @keichwa, this looks really good to me!
> 
> I guess the `Unchanged files with check annotations` are unrelated and general suggestions from Vale? This is good, but somehow we should avoid that they are added with each PR, if possible.

Yes, that's a temporary limitation only.  Thanks for your review!

PR #989, opened 2021-06-11
Free products are not checked for accessability anymore

mcalmer:
# Description

In the past we checked if free products like CentOS, OEL, Ubuntu or Debian are accessible and hide them from the product list when they are not.

As more and more customers configure their firewall to only allow access to certain hosts, we run often into timeouts.
So we stopped checking it for products where we definitiv know that the repos "should" be accessible.

For these customers it has the consequence that the products are visible, but if they add them, the mirroring will fail.

We adapted also the notification message to show details of the repo sync error. In the past they were available only in the logfiles. Now you can see something in the UI.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [x] Manager-4.1
- [ ] Manager-4.0

# Links


modules/installation/pages/setup-wizard.adoc
```diff
@@ -72,5 +73,6 @@ Synchronization progress for each product is shown in a progress bar next to the
 Depending on the products you have chosen, synchronization can take up to several hours.
 New products will be available for you to use in {productname} when synchronization is complete.
 
-If your synchronization fails, it could be because of a third party GPG key.
+If your synchronization fails, it could be because of a third party GPG key or your company firewall block access to the download server.
```
paususe, line 76:
```suggestion
If your synchronization fails, it could be because of a third party GPG key or your company firewall blocking access to the download server.
```

paususe (changes requested):

modules/installation/pages/setup-wizard.adoc
```diff
@@ -56,7 +56,8 @@ Navigate to the [guimenu]``Products`` tab to browse the products available to yo
 
 Filters help you search for products by description or architecture.
 
-The list is organized by product name.
+The list is organized by product name showing products on top which have a subscription.
+You will always see products which are freely available at the end of the list.
```
keichwa, line 60:
```suggestion
Freely available products appear at the end of the list.
```

keichwa (approved):
I also approved and merged Pau's suggestion.

mcalmer:
@keichwa will not be part of GM. It will be in the first MU and 4.1.9

PR #1000, opened 2021-06-20
Backport language selector to 4.2

paususe:
# Description

This PR is a compendium of fixes and enhancements required to ship translations in SUSE Manager 4.2:

1. Backport of the language selector to manager-4.2. It seems we forked docs too early, so it was not there.
2. Added missing flags for Korean and Japanese
3. From now on `make antora-suma` will build Japanese and Korean too. They were not building because... they were simply added initially disabled O:-)
4. Fix Lunr search. When translations were enabled, a langcode prefix was prepended to the path but Lunr is not aware of that, so the search always lead to 404 page not found errors. Since I have not found a way to properly tell antora-site-generator-lunr to use a prefix, I am fixing the resulting "database" (search-index.js) by way of `sed`.
5. Finally, I am disabling the "too smart" check that would overwrite an incomplete translation with English. It's not working well and it ended up ovewriting even full translations. Maybe it can be re-visited in the near future but we do not need it for SUSE Manager 4.2 GA.

Parts of these changes must be forward-ported to master. That's PR https://github.com/uyuni-project/uyuni-docs/pull/1001

# Target branches

- [ ] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,18 @@ define antora-uyuni-function
 	DOCSEARCH_ENABLED=true DOCSEARCH_ENGINE=lunr LANG=$(2) LC_ALL=$(2) LC_ALL=$(2) antora $(current_dir)/$(1)/uyuni-site.yml --generator antora-site-generator-lunr
 endef
 
+define fix-lunr-search-in-suma-translation
+	$(call fix-lunr-search-in-translation,suse-manager,$(1))
+endef
+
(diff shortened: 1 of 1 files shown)


jcayouette (approved):
LGTM, nice work thanks @paususe

PR #1009, opened 2021-06-23
Update client tables for manager-4.2

paususe:
# Description

The same client table is in two pllaces (Installation Guide and Client Configuration Guide) and it was outdated in both places. I am updating it here by essentially duplicating it to avoid breaking translations too much but we should get rid of the table in the Installation Guide and just refer to the table in the Client Configuration Guide for the future.

I will cherry-pick this to `master` after merging in manager-4.2

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/modules/client-configuration/pages/supported-features.adoc
+++ b/modules/client-configuration/pages/supported-features.adoc
@@ -80,11 +80,6 @@ The operating system you run on a client is supported by the organization that s
 | {check}
 | {check}
 
-| {sles} ES 6
-| {x86}, {x86_64}
-| {check}
-| {check}
-
 | {rhel} 8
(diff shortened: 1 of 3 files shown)


jcayouette (approved):
Looks good to me, thank you.

PadraigDillon1:
These changes have added strings to be translated for 4.2. Do we have time/need to translate these new strings?

paususe:
> These changes have added strings to be translated for 4.2. Do we have time/need to translate these new strings?

It's OK to do it for 4.2.1

PR #1016, opened 2021-07-06
Manager42 - New db migration 0rnela

0rnela:
# Description
Following the changes to DB migration script, where with values are now inferred from the system, DB has been changed to reflect that.
- Individual pages to describe migration to PostgreSQL 12 or 13 have been dropped, and replaced with just one page describing the migration to the latest version.
- Navigation bar has been cleaned up.
- Chapter introduction has been cleaned up to remove redundant sentences and descriptions.
- Chapter introduction has been reorganised to keep the references to very old migration process to PostgreSQL 10 by using the link to external documentation, just like was done in the existing admonition which I reused for that purpose.
-  
# Target branches

- [ ] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/15217

### DIFF
--- a/modules/upgrade/pages/db-intro.adoc
+++ b/modules/upgrade/pages/db-intro.adoc
@@ -65,18 +65,10 @@ This table shows the PostgreSQL version required for each version of {productnam
 
 |===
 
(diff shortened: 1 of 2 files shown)


0rnela:
Error while creating PR. Investigating.

PR #1028, opened 2021-07-10
Use po4a from OBS Uyuni:Utils instead of my PPA

paususe:
# Description

Use po4a package for Ubuntu from Uyuni:Utils OBS repo instead of using Pau's PPA.

NB: 
- as of today, the po4a package in the devel:perl/po4a OBS (which is what the Uyuni:Utils OBS inherits) does not install with the default settings in the GH Actions container. 
- I have fixed the problem in my OBS and submitted a new version to devel:perl/po4a. 
- Since I'm not sure I will still be around when that is accepted, I this PR contains a commented-out GH Actions step with the Uyuni:Utils OBS, and an enabled GH Actions step with my OBS so that this is in a working state
- Make sure you uncomment the Uyuni:Utils one in the future and get rid of the one using my OBS!

NB2: You want to squash this PR when merging!

NB3: You want to forward-port this to `master` (it should be a matter of one cherry-pick).

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/.github/workflows/feature_tests_uyuni.yml
+++ b/.github/workflows/feature_tests_uyuni.yml
@@ -65,6 +65,6 @@ jobs:
 
     - name: '[BUILD] SUMA PDF Documents with Asciidoctor-pdf'
       run: |
-        sudo locale-gen zh_CN.UTF-8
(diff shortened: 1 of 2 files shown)


keichwa:
Hi Pau, thanks for working on this.

@jcayouette Pau says and we must not forget about this:
> Make sure you uncomment the Uyuni:Utils one in the future and get rid of the one using my OBS!

PR #1060, opened 2021-07-26
Fix the procedure migration from Leap 15.2

juliogonzalez:
# Description

Fix the procedure migration from Leap 15.2

Because of https://bugzilla.opensuse.org/show_bug.cgi?id=1184326, the proxy must have the last version of `openSUSE-build-key`. On paper we should tell the user to update all packages, except those for Uyuni, before the migration.

But IMHO, let's avoid complicating the procedure for now, if this works.

Also it seems that a regular upgrade is not enough, as a `dup` is needed. Not sure this worked either for the migration from 15.1 to 15.2, but @keichwa should now better. So this PR also documents how to do it with a remote command (is there a better way?)

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Related: https://github.com/uyuni-project/uyuni/issues/4026

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -4,6 +4,8 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- Update the proxy migration procedure with the specifics for the migration
+  from openSUSE Leap 15.2 to 15.3
 - Documented version 2 of Inter-Server Synchronization in the Administration
   Guide
 - Document Rocky Linux 8 as a client


modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -7,9 +7,20 @@ For more information about maintenance windows, see xref:administration:maintena
...
+This first procedure is needed because of https://bugzilla.opensuse.org/show_bug.cgi?id=1184326
+and should not be needed when we prepare the migration from Leap 15.3 to 15.4
+////
+.Procedure: Update the package opensuse-build-key at the the proxies
+
+. At the command prompt on the {productname} Server, as root, use the [command]``spacewalk-repo-sync`` command to sync all the openSUSE Leap 15.2 channels:
+----
+spacewalk-repo-sync -p opensuse_leap15_2-x86_64
+----
+. Update the package `openSUSE-build-key` at all  {productname} proxies.
```
keichwa, line 20:
```suggestion
. Update the package `openSUSE-build-key` on all  {productname} proxies.
```
I'm not an English native speaker, though.

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -51,7 +61,12 @@ spacewalk-proxy stop
 . Click menu:Software[Software Channels], and as the base channel select the openSUSE Leap {opensuse-version} channel that is listed in the [systemitem]``Customs Channels`` list.
 . In the [guimenu]``Child Channels`` pane, select the {opensuse-version} child channels.
 . Click btn:[Next], and [guimenu]``Confirm Software Channel Change`` with btn:[Confirm].
+////
```
keichwa, line 64:
```suggestion
+
////
```
We'll probably need the + connector here.  If not, it could be the case that the comment would finish the list and the following item would start a new list or procedure...

keichwa (approved):

juliogonzalez:
All suggestions accepted :-)

@keichwa: Should we wait for a second review from @0rnela or can we merge?

keichwa:
Feature- and language-wise it looks ok to me.  You please decide whether we need an entry in the toplevel `.changelog` file ;)  iirc, you have merge permissions.  Please fell free to merge.

juliogonzalez:
Changelog added. Merging.

PR #1069, opened 2021-07-29
Ubuntu tshoot inst spacecmd #1187149 keichwa

keichwa:
# Description

https://github.com/SUSE/spacewalk/issues/15199
https://bugzilla.suse.com/show_bug.cgi?id=1187337

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


keichwa:
Please, review.  If you agree I'll add something along these lines for the older ubuntu clients, too.

Or should be better add this to `client-configuration/pages/tshoot-clients.adoc`?

jcayouette (approved):

mcalmer (approved):
LGTM. In the case of spacecmd we are more looking into the possibility to remove the dependency to "rpm" instead of adding universe. But as spacecmd is only needed on some clients (maybe none) we can live with the fact that for now you need to enable universe.

PR #1073, opened 2021-08-02
Uyuni proxy upgrade #15554 keichwa

keichwa:
# Description

Add sections about:
-  2021.05 -> 2021.08: Major migration, base OS changes from Leap 15.2 to Leap 15.3
- 2021.06 -> 2021.08: Minor migration, base OS does not change

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/15554


jcayouette (approved):

keichwa:
please review again.  I'll upload the PDF soon.

modules/upgrade/nav-upgrade-guide.adoc
```diff
@@ -25,7 +25,8 @@ endif::[]
 ** xref:proxy-intro.adoc[Upgrade the Proxy]
 ifeval::[{uyuni-content} == true]
 // *** xref:proxy-autoyast-uyuni.adoc[Proxy - Upgrade with AutoYaST]
-*** xref:proxy-uyuni.adoc[Proxy - Upgrade Procedure]
+*** xref:proxy-uyuni.adoc[Proxy - Upgrade Procedure from 2021.05]
+*** xref:proxy-minor-uyuni.adoc[Proxy - Upgrade Procedure from 2021.06]
```
juliogonzalez, line 29: Why do we want to mention versions here? Why can't we do this as for SUSE Manager with generic titles :-?

In fact that's what you use later at `modules/upgrade/pages/proxy-intro.adoc`

modules/upgrade/pages/proxy-minor-uyuni.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+
+.Procedure: Updating the {productname} Proxy
+. On the {productname} Proxy, stop the proxy service:
++
+----
+spacewalk-proxy stop
+----
+. In the {productname} Server {webui}, navigate to menu:Systems[Proxy] and click the name of the proxy.
+. Select packages to be updated on the proxy, and then apply the selection.
```
juliogonzalez, line 29: As I user I'd ask: what packages?

And my reply would be `all of them`. Unless you have something else in mind.

modules/upgrade/pages/proxy-minor-uyuni.adoc
```diff
@@ -0,0 +1,37 @@
...
+= Proxy - Minor Upgrade
+
+Before you perform any proxy update, schedule a maintenance window.
+The clients registered to {productname} through the proxy will not be able to connect to {productname} while the update is in progress.
+For more information about maintenance windows, see xref:administration:maintenance-windows.adoc[].
+
+[NOTE]
+====
+Minor proxy upgrades do not include a version upgrade of the operating system.
+For example, while upgrading from 2021.06 to 2021.08 the operating system will stay at {opensuse} Leap 15.3.
```
juliogonzalez, line 11: Do we really want to mention this? I'd just tell the user to check the release notes, to see if a major upgrade is needed.

Otherwise I doubt will remember to update this next time.

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -1,17 +1,25 @@
...
 
 Before you perform any proxy update, schedule a maintenance window.
 The clients registered to {productname} through the proxy will not be able to connect to {productname} while the update is in progress.
 For more information about maintenance windows, see xref:administration:maintenance-windows.adoc[].
 
-== Preparation
+[NOTE]
+====
+Major proxy upgrades include a version upgrade of the operating system.
+For example, the upgrade from 2021.05 to 2021.08 includes the {opensuse} Leap version upgrade from 15.2 to 15.3.
```
juliogonzalez, line 11: Same as before.

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -1,17 +1,25 @@
...
+====
+
+
+
+== Preparation for the Upgrade from 2021.05 to 2021.08
```
juliogonzalez, line 16: As discussed before: I'd not mention the version.

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -35,7 +43,9 @@ uyuni-proxy-stable-leap-153
 ----
 . Fully synchronize all channels with [command]``spacewalk-repo-sync``.
 
-== Upgrade the Proxy
+
+
+== Upgrade the Proxy from 2021.05 to 2021.08
```
juliogonzalez, line 48: Same

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -80,3 +90,19 @@ spacewalk-proxy start
 
 If you need to update many proxies, you can create an action chain of this command sequence on the {productname} Server.
 You can use the action chain to perform updates on multiple proxies at the same time.
+
+
+
+////
+== Update the Proxy (Y)
```
juliogonzalez, line 97: What does `(Y)` means?

modules/upgrade/pages/proxy-uyuni.adoc
```diff
@@ -80,3 +90,19 @@ spacewalk-proxy start
...
+
+To update a proxy use the [guimenu]``Product Migration``:
+
+// FIXME: Update images
```
juliogonzalez, line 101: I guess this can be removed?

juliogonzalez (changes requested):
Some comments. In particular: I'd not mention versions anywhere. Let's keep to procedures (minor/major), and let's point the users to the release notes.

Otherwise, I can accept this, but only if we have:
- An entity called `uyuni-oldversion` which as of today it will be the last supporting Leap 15.2 as base OS (2021.05)
- All current references to 2021.08 using the current entity for the uyuni version, so it's automatically updated.
- A explicit version for the user to *always* check the release notes, in case an extra procedure is needed.

juliogonzalez:
One last thing about my mention to the old-version.

That's the latest you can migrate from. But you can also migrate from anything betweek 2020.07 and 2021.05  (both included!).

Not sure how you want to address that. Two entities? Just tell the user to check the releases notes?

keichwa:
> Some comments. In particular: I'd not mention versions anywhere. Let's keep to procedures (minor/major), and let's point the users to the release notes.

I'll happily rely on version numbers provided by the RN.

>     * An entity called `uyuni-oldversion` which as of today it will be the last supporting Leap 15.2 as base OS (2021.05)

Ok, we now have this in our .yml.  But ATM, I do not use it.

>     * All current references to 2021.08 using the current entity for the uyuni version, so it's automatically updated.

Ok, I hope I catch all occurrences.

>     * A explicit version for the user to _always_ check the release notes, in case an extra procedure is needed.

Such a waning is in now.  I'll try to rebuild the PDF.

modules/upgrade/pages/proxy-minor-uyuni.adoc
keichwa: ok

modules/upgrade/pages/proxy-minor-uyuni.adoc
keichwa: ok, but we'd better check this always.

modules/upgrade/pages/proxy-uyuni.adoc
keichwa: ATM, this is commented (block comment with 3 tics).  This should be the upgrade with the OS upgrate from X.Y to X.Y+1 (aka SP or Product Migration).  Something for the future.

modules/upgrade/pages/proxy-uyuni.adoc
keichwa: If we really want to enable Product migration we probably must check the screen shots...)

keichwa:
> One last thing about my mention to the old-version.
> 
> That's the latest you can migrate from. But you can also migrate from anything betweek 2020.07 and 2021.05 (both included!).
> 
> Not sure how you want to address that. Two entities? Just tell the user to check the releases notes?

Just release notes should be good enough.

keichwa:
General remark I do not like using productnumber (aka version number) entities too much.There are often tiny or not that tiny differences.  But this is a team document, thus I happily adjust this according to your proposals (that makes sense and can be followed without much trouble).

juliogonzalez:
For me it's OK if you get rid of all the versions, as I told. We can provide that information at the release notes, as we did in the past.

All I said is: if you provide "one" version, then provide all of them. Otherwise let's remove all of them and refer the users to the release notes :-)

keichwa:
I think I removed all explicit version info in these upgrade sections.

jordimassaguerpla (approved):

PR #1089, opened 2021-08-09
Removed additional duplications of SLE SLE in SLE Micro document.

0rnela:
# Description

SLE Micro clients had duplicate SLE in the copy, i.e. SLE SLE Micro.

Redundant SLE was removed.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2 https://github.com/uyuni-project/uyuni-docs/pull/1100
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/modules/client-configuration/pages/clients-slemicro.adoc
+++ b/modules/client-configuration/pages/clients-slemicro.adoc
@@ -1,14 +1,14 @@
 [[clients-slemicro]]
-= Registering {slea}{nbsp}{micro} Clients
+= Registering {nbsp}{micro} Clients
 
-This section contains information about registering clients running these {slea}{nbsp}{micro} operating systems:
+This section contains information about registering clients running these {nbsp}{micro} operating systems:
 
-* {slea}{nbsp}{micro}{nbsp}5.0 {x86_64}
-* {slea}{nbsp}{micro}{nbsp}5.0 {arm64}
+* {nbsp}{micro}{nbsp}5.0 {x86_64}
+* {nbsp}{micro}{nbsp}5.0 {arm64}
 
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -1,14 +1,14 @@
 [[clients-slemicro]]
-= Registering {slea}{nbsp}{micro} Clients
+= Registering {nbsp}{micro} Clients
```
juliogonzalez, line 2: Why do we need `{nbsp}`? in front of `{micro}` now? Same for all other changes.

juliogonzalez (approved):
Assuming this is the way you want it done, instead of changing `micro` entity to be `Micro`, then LGTM (but first fix the unneeded `{nbsp}` and check the rendering, just in case).

jcayouette (approved):

modules/client-configuration/pages/clients-slemicro.adoc
jcayouette: @juliogonzalez the nbsp ensures we have the space for styles reasons.

keichwa: Yes, but here without a leading word non-breaking space does not make that much sense.  I think it actually could securely be removed with the leading "{slea}".

juliogonzalez: @jcayouette are you **sure** we won't have now two whitespaces?

juliogonzalez:
BTW: Remember about the backports.

PR #1099, opened 2021-08-11
remove red hat 6 and ubuntu 16.04

keichwa:
# Description

https://bugzilla.suse.com/show_bug.cgi?id=1188656
https://github.com/SUSE/spacewalk/issues/15618

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [x] Manager-4.1
- [ ] Manager-4.0

# Links

https://github.com/SUSE/spacewalk/issues/15618

### DIFF
--- a/modules/client-configuration/pages/clients-rh-cdn.adoc
+++ b/modules/client-configuration/pages/clients-rh-cdn.adoc
@@ -177,6 +177,12 @@ The channels you need for this procedure are:
 | {redhat} 8 | RHEL or SLES ES or CentOS 8 Base | rhel8-pool-x86_64
 |===
 
+[NOTE]
+====
+{redhat} 6 is now at end-of-life, and the ISO images provided in the repository are out of date.
+Bootstrapping new {redhat} 6 clients using these packages will fail.
+If you need to bootstrap new {redhat} 6 clients, follow the troubleshooting procedure in xref:client-configuration:tshoot-clients.adoc[].
+====
 
 include::snippets/manual_channels.adoc[]
 


keichwa:
systems were already removed from the list of supported clients.  I now also removed info about channel names, etc.

juliogonzalez (commented):
If we are going to do this, what about Oracle 6?

keichwa:
> If we are going to do this, what about Oracle 6?

In either case, I've no strong opinion.  I can also remove that one or keep the others.  @mcalmer or Julio, what do you think?

mcalmer:
From the Release Notes:
```
Deprecated products
* Red Hat Enterprise Linux 6
* SUSE Linux Enterprise Server Expanded Support 6
* Oracle Linux 6
* CentOS 6
* Ubuntu 16.04 LTS
```

Oracle Linux 6 is also deprecated and should be removed.

mcalmer:
I checked also 4.1 checkbox. The removal need to happen also in the 4.1 docs.

juliogonzalez:
> Oracle Linux 6 is also deprecated and should be removed.

And CentOS6 and RES6, of course

keichwa:
I added warnings as needed, and kept the details (channel names).

juliogonzalez (commented):
Please generate the html doc, tar.gz it and attach the resulting tarball for review.

Reviewing the PR I'd say we are adding way too many nontes.

keichwa:
> Please generate the html doc, tar.gz it and attach the resulting tarball for review.
> 
> Reviewing the PR I'd say we are adding way too many nontes.

Ok, do you mean the output of `make obs-packages-suma-en`?  Or `make antora-suma-en` (and a manually created tar.gz)?  I'll attach the antora output of the client-config.  Is this enough?

[suse-manager.tar.gz](https://github.com/uyuni-project/uyuni-docs/files/6990252/suse-manager.tar.gz)

keichwa:
Can I have some tics?  @juliogonzalez is the .tar.gz good enough?

juliogonzalez:
I meant `make antora-suma-en, and a manually created tarball, yes. But the whole thing with all guides (basically the same you'd submit for a MR for d.s.c IIUC).

This tarball only contains the client guide and I need to look for the files and open them one by one.

Please regenerate the tarball.

keichwa:
ok, thanks for advise.  attaching here permanently fails.  I copied it to my suse export.

juliogonzalez (approved):
Looks good, but for RH (CDN and RHUI) the not is in the middle of the document. I am not sure if we should move leave it there, move it to the top of the doc (and for other clients), or repeat it :-)

But approving, since that's a decision for you to take.

keichwa:
> Looks good, but for RH (CDN and RHUI) the not is in the middle of the document. I am not sure if we should move leave it there, move it to the top of the doc (and for other clients), or repeat it :-)

RHUI looks ok to me.  in CDNit seems to make sense to have an addtitional earlier warning.

PR #1100, opened 2021-08-11
Backport to 42 - Removed double SLE SLE 

0rnela:
# Description

SLE Micro clients had duplicate SLE in the copy, i.e. SLE SLE Micro.

Redundant SLE was removed.

# Target branches

- [x] Master (Default) https://github.com/uyuni-project/uyuni-docs/pull/1089
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/modules/client-configuration/pages/clients-slemicro.adoc
+++ b/modules/client-configuration/pages/clients-slemicro.adoc
@@ -1,23 +1,31 @@
-[[clients-slemicro]]
-= Registering {nbsp}{micro} Clients
+[clients-slemicro]]
+= Registering {micro} Clients
 
 This section contains information about registering clients running these {nbsp}{micro} operating systems:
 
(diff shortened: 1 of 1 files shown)


juliogonzalez (commented):
Same question as for the master PR. I think `{nbsp}` is not needed now that you removed the leading `{slea}`. Did you check the render? Maybe it doesn't matter... or maybe we are adding two whitespaces.

0rnela:
> Same question as for the master PR. I think `{nbsp}` is not needed now that you removed the leading `{slea}`. Did you check the render? Maybe it doesn't matter... or maybe we are adding two whitespaces.

You are correct, we have two double spaces!
I'll correct it here and on master too.

keichwa (approved):
the leading {nbsp} is still strange, but for now it is ok.

PR #1107, opened 2021-08-13
add information about pam service name

rjmateus:
Signed-off-by: Ricardo Mateus <rmateus@suse.com>

# Description

Adds some clarification about rules when specifying pam service name (filename).

# Target branches

- [X] Master (Default)
- [X] Manager-4.2
- [X] Manager-4.1
- [ ] Manager-4.0

# Links


rjmateus:
ping @mcalmer for awareness

modules/administration/pages/auth-methods-pam.adoc
```diff
@@ -9,6 +9,7 @@ PAM is a suite of libraries that allows you to integrate {productname} with a ce
 
 .Procedure: Enabling PAM
 . Create a PAM service file at [path]``/etc/pam.d/susemanager``.
+  Filename must be in lower case and readable by ``tomcat`` user.
```
mcalmer, line 12: I think I know what was going wrong for this customer. We use `/sbin/unix2_chkpwd` to do the pam authentication.
That bin should have the suid bit set:
```
$> ls -la /sbin/unix2_chkpwd
-rwsr-xr-x 1 root shadow 14792 19. Mai 14:13 /sbin/unix2_chkpwd*
```
I assume with this bit set, you do not need ownership to tomcat for this file.
The customer might have applied paranoid permissions settings.

rjmateus: Hey @mcalmer 
I had asked the customer to change read permissions on the file and that worked: https://bugzilla.suse.com/show_bug.cgi?id=1188558#c7

I also tested it locally and If the user that is running the pam check doesn't have read permissions on the file will get an error saying the service doesn't exist.
I think the `tomcat` user does not need to have ownership of the file but needs to be able to read it.

keichwa (approved):

modules/administration/pages/auth-methods-pam.adoc
keichwa: @mcalmer can you please take a look again now?

mcalmer: If it was tested like this it should be ok. On our test server this file has the following permissions:
```
-rw-r--r-- 1 root root 165  1. Jul 2020  /etc/pam.d/susemanager
```
which is in line with what we write now.

mcalmer (approved):

keichwa:
Thanks for all your help!  I think now I can backport it.  Or do you want to do it yourself, Ricardo?

rjmateus:
> Thanks for all your help! I think now I can backport it. Or do you want to do it yourself, Ricardo?

Hey  @keichwa 
I just opened the PR's for manager 4.1 and 4.2.

keichwa:
> I just opened the PR's for manager 4.1 and 4.2.

Thank you, Ricardo!  I reviewed and merged them now.

PR #1111, opened 2021-08-17
SUSE Manager: Universe is not required for installing spacecmd on Ubuntu

juliogonzalez:
# Description

We already provided the require binaries for Ubuntu 18.04, but the channels definitions were wrong. That will be fixed for 4.2.2.

For Ubuntu 20.04 will start providing the packages as well.

So no more Universe requirement for SUSE Manager.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [x] Manager-4.1
- [ ] Manager-4.0

# Links

https://github.com/SUSE/spacewalk/issues/15199


.changelog
```diff
@@ -4,6 +4,11 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- SUSE Manager: Universe is not required for installing spacecmd
+  on Ubuntu (bsc1187337)
+
+# Last packaged: ?
```
juliogonzalez, line 10: I am not sure what do do here. I think Joseph should have cleaned this up when he packaged last time. Anyway, please fix it on a separate PR :-)

keichwa: :+1: if nobody knows better, we can write `Last packaged: before 2021-08-18`.

keichwa (approved):
thanks for the cleanup.  Looks save!

.changelog
```diff
@@ -4,8 +4,13 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- SUSE Manager: Universe is not required for installing spacecmd
+  on Ubuntu (bsc1187337)
 - In the Prometheus chapter of the Administration Guide advise to store
   data locally (bsc#1188855)..
```
juliogonzalez, line 10: @keichwa I guess that part of the changelog is not right, as it contains two dots ^, make sure you fix it so Joseph doesn't need to take care of it on packaging day :-)

juliogonzalez:
@keichwa FYI: Seems what I am removing here was never backported to 4.2 or 4.1, so need to backport this PR either.

PR #1134, opened 2021-08-27
SLE Micro client support matrix page

moio:
# Description

* replaces remaining occurrences of "Service Pack Migration" to "Product Migration" (as we renamed the feature in 4.2)
* rewords the Advanced virtual guest management description for clarity
* **adds a comparison page for SLE Micro supported features**

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0


modules/client-configuration/pages/supported-features-sle-micro.adoc
```diff
@@ -0,0 +1,165 @@
...
+| {salt}
+
+| Organization custom states
+| {salt}
+
+| System set manager (SSM)
+| {salt}
+
+| Service pack migration
+| N/A
```
mcalmer, line 69: Should we better say {question} here? Or is somebody trying this out already and we just wait for the result?

modules/client-configuration/pages/supported-features-sle-micro.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+| Basic Virtual Guest Management {star}
+| {cross}
+
+| Advanced Virtual Guest Management {star}
+| {question}
```
mcalmer, line 75: Basic is not working, but advanced should be supported in future? This does not make much sense, or do I miss something?

modules/client-configuration/pages/supported-features-sle-micro.adoc
```diff
@@ -0,0 +1,165 @@
...
+| {cross}
+
+| Configuration file management
+| {salt}
+
+| Snapshots and profiles
+| {salt}: Profiles supported, Sync not supported
+
+| Power management
+| {question}
```
mcalmer, line 132: I think Power Management depends more on the Hardware than on the OS.
Only a wrong ACL in the UI could prevent this working. Do we have a problem with the UI ACL?

mcalmer (commented):
Some remarks added

modules/client-configuration/pages/supported-features-sle-micro.adoc
moio: Well, SLE Micro does not have Service Packs AFAIU, only major versions. So this should be called differently - do we want to diverge from the feature list of all other OSs?

modules/client-configuration/pages/supported-features-sle-micro.adoc
moio: AFAIU from @cbosdo basic actually means "the set of functionality that the traditional stack provides". Given only Salt is supported here, we could maybe change the title for all OSs. I was confused myself.

modules/client-configuration/pages/supported-features-sle-micro.adoc
moio: Not that I am aware of. We will just need to test this before we remove the "tech preview" label.

modules/client-configuration/pages/supported-features-sle-micro.adoc
mcalmer: Well, the feature in SUMA to be used is the same. And we renamed the Tab to "Product Migration" which would even better match.
So we can think about renaming the description to Product Migration.

modules/client-configuration/pages/supported-features-sle-micro.adoc
mcalmer: But this "set of functionality" is also provided by salt. AFAIK there is a list at the end of this page which describe what Basic and Advanced mean.

modules/client-configuration/pages/supported-features-sle-micro.adoc
mcalmer: sure, we should let QA went over all of these items and let them explicitly test it.

modules/client-configuration/pages/supported-features-sle-micro.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+
+{star} Virtual Guest Management:
+
+In this table, virtual guest management is split into basic and advanced.
+
+Basic virtual guest management includes listing VMs, slow refresh, VM lifecycle actions (start, stop, resume, pause), and modifying VM vCPU and Memory.
+
+Advanced virtual guest management includes fast refresh, VM lifecycle actions (delete, reset, power off), modifying VM disk, network, graphical display, and graphical display configuration.
```
moio, line 165: @cbosdo is it right to say Advanced virtual guest management includes everything in the Basic list as well?

Or are those separate?

cbosdo: Advanced means Basic + VM console, delete, edit + Virtual nets + Virtual storage

moio:
@keichwa this should be good now from the tech POV, can you take a look from the docs part and merge if it's OK?

mcalmer (approved):

keichwa (approved):
What a heroic effort--thanks so much, Silvio!

PR #1136, opened 2021-08-31
Document Debian 11

juliogonzalez:
# Description

Document Debian 11

# Target branches

- [x] Master (Default). **Do NOT merge before the `Uyuni-2021.12` branch is created!!!!!**
- [ ] Manager-4.2: https://github.com/uyuni-project/uyuni-docs/pull/1533
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

https://github.com/SUSE/spacewalk/issues/15683


.changelog
```diff
@@ -4,6 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- Document Debian 11 as a supported OS as a client
```
```suggestion
- Documented Debian 11 as a supported OS as a client in Client
  Configuration Guide
```
[According to the recipe above.  We probably should review our recipes...]

keichwa (dismissed):

keichwa:
@juliogonzalez do you have commit rights?  Or is this still WIP?  If it is WIP, we'd better convert it to "draft mode"--there is this option in the "Reviewers" section.

juliogonzalez:
@keichwa this is still WIP, I am still testing Debian11 so this is not ready to be merged.

juliogonzalez:
Ready to be reviewed. Sorry for not adding the render of the doc, but (at least today), GitHub is rejecting .zip and .tar.gz files, and other compression formats are not accepted :(

jcayouette (approved):

jcayouette (changes requested):
Could we split this content into individual Uyuni/SUMA tables?

We should not use ifdefs inside of our tables as they are known to break occasionally for localized documents. "I still have not figured out why".

It is fine to use conditional statements (ifeval, ifdef) as wrappers but using them inside of a table can cause bugs.

So two tables:

One for suma-content = true

One for uyuni-content= true

Sorry I did not mention this before. This bug was discovered in a localization review from Patrick.

juliogonzalez:
@jcayouette not on this PR, sorry. I just did what we did in the past.

Besides the tables are huge, so then you need to make sure they are coordinated because the content is almost the same.,

juliogonzalez:
Can we merge the PR? The code is already merged.

Later you can discuss with your squad how to handle those huge tables. I am not sure not using if/else is the right approach in this case, given the maintenance effort for two tables that will be almost the same. But have a meeting inside the team, discuss it, do the split, and make sure you document it, as otherwise we'll end with unsynced tables when someone updates one, but not the other (even if updates for both are needed).

jcayouette:
Sure but just know it may break in the localized versions.

jcayouette:
Technically using a snippet for these supported tables would be best. I need to ping upstream to see why these bits are breaking. It is not always the case.

jcayouette (approved):
Approved but we need to revisit conditionals in tables at some point.

PR #1156, opened 2021-09-07
fix indentation in antora yaml

mbussolotto:
# Description

fix indentation in antora yaml 

# Target branches

- [ ] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/uyuni-site.yml
+++ b/uyuni-site.yml
@@ -123,11 +123,11 @@ asciidoc:
 content:
   sources:
   # Uncomment to build from your local git clone
-  #- url: .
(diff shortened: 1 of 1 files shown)


jcayouette (changes requested):
The remote should not be the default currently as we end up with conflicts with our weblate automation. Just keep the default local setting and comment out the remote.

mbussolotto:
Thanks @jcayouette ! It should be better now

jcayouette (approved):

PR #1192, opened 2021-09-21
Clean before building the OBS packages

juliogonzalez:
# Description

Clean before building the OBS packages. Otherwise if you build suma, and then you build uyuni (without using `make clean`), the suma PDFs are not removed and get included (as it happened at https://build.opensuse.org/request/show/920670).

I tested locally and it seems to be working fine, but since I am not an expert in the new structure, I guess the reviewer will catch any potential issues.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [x] Manager-4.1
- [ ] Manager-4.0

Not sure about the backporting, as I am not sure if 4.1 and 4.2 have the new make system in place.

# Links

Fixes https://github.com/SUSE/spacewalk/issues/15682

### DIFF
--- a/parameters.yml
+++ b/parameters.yml
@@ -131,7 +131,7 @@ targets: #Other makefile target not strictly related to book. These targets will
   - name: antora-suma
     pre_lang_prerequisite: configure-suma copy-branding 
   - name: obs-packages-suma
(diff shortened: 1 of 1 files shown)


mbussolotto (approved):
good for me!

PR #1204, opened 2021-09-27
Adding aarch64 for a selection of clients.

0rnela:
# Description

Client Config Guide - aarch64 for selected clients added.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2 https://github.com/uyuni-project/uyuni-docs/pull/1204
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/14440


juliogonzalez (changes requested):
Is this is a WIP, it's a WIP and it's not ready to be reviewed.

Request a review again, when the PR is ready :-)

In any case you need still need to modify the pages for each OS, as right now we only mention channels for x86_64.

jcayouette (approved):
LGTM. Nice work. Ornela. As Julio said, please do not label something a WIP and request a review. :)

juliogonzalez:
@jcayouette the description clearly says:

# PLEASE DO NOT APPROVE YET AND DO NOT MERGE

0rnela:
[PR1204.tar.gz](https://github.com/uyuni-project/uyuni-docs/files/7242829/PR1204.tar.gz)

0rnela:
Thanks @juliogonzalez @jcayouette for clarification - don't add reviewers as placeholders while still at WIP stage. 
Going forward:
1. FIRST make sure everything is fully ready
2. THEN add reviewers

juliogonzalez:
> Thanks @juliogonzalez @jcayouette for clarification - don't add reviewers as placeholders while still at WIP stage. Going forward:
> 
> 1. FIRST make sure everything is fully ready
> 2. THEN add reviewers

Yes. In some situations you may want people to have a look, for example in this case, because you did a change that you will later need to replicate. But still pinging them is the right thing to do.

If you add then as reviewers at that point, and they they approve, they will not review the final state of the PR (and they should)

modules/installation/pages/client-requirements.adoc
```diff
@@ -36,65 +36,132 @@ For non-{suse} operating systems, including {rhel}, {centos}, and {oracle}, only
...
+ifeval::[{uyuni-content} == true]
+| {alibabaclo} 2
+| {x86_64}, {aarch64}
 | {cross}
 | {check}
+endif::[]
 
-| CentOS 7
+| {almalinux} 8
 | {x86_64}
```
juliogonzalez, line 81: You are missing aarch64 here. The original issue says "Document aarch64 for CentOS7/8, Oracle7/8, Rocky Linux 8, AlmaLinux 8, Amazon Linux2, openSUSE Leap 15.3". Review the others OS, in case I am missing some more.

modules/client-configuration/pages/clients-almalinux.adoc
```diff
@@ -33,9 +33,13 @@ You do not need to disable SELinux to register {almalinux} clients to {productna
 
 Before you can register {almalinux} clients to your {productname} Server, you need to add the required software channels, and synchronize them.
 
+To check whether the product and architecture you need is supported, see xref:client-configuration:supported-features.adoc[].
+
+
 ifeval::[{suma-content} == true]
 
-The products you need for this procedure are:
+For example, when working with x86_64 architecture, you need this product:
```
juliogonzalez, line 41: I see a problem with this approach: for the channels, the architecture is `x86_64`, but at the matrix you render it as `x86-64`. For Z series, the architecture for the channel is ' s390x` but at the matrix you have "IBM Z". For ARM64, the architecture is `aarch64` for the channels, but at the matrix you have `ARM`.

image


So this won't work unless the users can copy the architecture from the matrix.

modules/installation/pages/client-requirements.adoc
0rnela: Added after review.

modules/client-configuration/pages/clients-almalinux.adoc
0rnela: This is correct., but I prefer to separate this kind of tidy-up in separate task as it is not regression.

jcayouette: Agreed if this falls in line with former process than merge as is. We can clean up in a new PR.

jcayouette (approved):

modules/client-configuration/pages/clients-almalinux.adoc
juliogonzalez: > This is correct.
However, the changes added are in line with how the affected docuents were updated in the past, and the existing content is not regression.
> I would therefore recommend that we to separate the kind of tidy-up that would definitely be good in separate PR.

I agree on other situations, but not for this one. What will happen when a customer tries to add `almalinux8-ARM` via `mgr-sync`?

If you want cleanup, then please change this message to:

```
Two architectures are supported: `x86_64` and `aarch64`

For example, when working with `x86_64` architecture, you need this product:
```
Even without using entitires, if needed.

FMPOV this is easy to do, and make things understandable while we get the problem fixed with a tidy up. Specially because I don't think the tidy up is going to happen in the next few days, in the for 4.2.3 (otherwise feel free to ignore this comment).

Remember to check if all OS form the card are documented at the table, I saw AlmaLinux, but could be others.

0rnela: @juliogonzalez changed for the clients from this round of changes.

.changelog
```diff
@@ -4,6 +4,8 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- Updated Client Config and Install Guides with aarch64 support for
```
juliogonzalez, line 7: Use the same format as the previous entry :-)

Seems you decided to use that, instead of doing as we do for the code and then with tito.

juliogonzalez (changes requested):
Seems you added aarch64 for Leap 15, but didn't change the client registration page for it. Remember to add it there as well.

Since you didn't provide the rebuild, I will assume the tables are already looking good, and that all OS that require aarch64 to be added are OK

.changelog
```diff
@@ -4,8 +4,10 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
-  * Documented AWS Permissions for Virtual Host Manager in VHM and
-    Amazon Web Services chapter in the Client Configuration Guide
+- Added aarch64 support for selection of clients in the Install
```
```suggestion
- Added aarch64 support for selection of clients in the Installation
```

keichwa (approved):

jcayouette (approved):

juliogonzalez (approved):
Looks OK. Remember to backport to `Manager-4.2`

PR #1208, opened 2021-09-29
Backport42 - Adding aarch64 info

0rnela:
# Description

Client Config Guide - aarch64 for selected clients added.

# Target branches

- [x] Master (Default) https://github.com/uyuni-project/uyuni-docs/pull/1204
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/14440

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -4,7 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
--  Added aarch64 support for selection of clients in the Installation
+  * Added aarch64 support for selection of clients in the Installation
    Guide and Client Configuration Guide
   * Documented AWS Permissions for Virtual Host Manager in VHM and
     Amazon Web Services chapter in the Client Configuration Guide


.changelog
```diff
@@ -4,6 +4,8 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+-  Added aarch64 support for selection of clients in the Installation
```
juliogonzalez, line 7: Wrong changelog format, it seems. Please agree with the squad what format will be used for the changelogs, according to the changelog regulations I sent to the mailing lists.

If you won't bump versions for packages (as you do now for 4.1 and 4.2), then your format for this PR is OK, as the result will be:

```
- Change A
- Change B
- Change C
```
However this is not OK for Uyuni, where we bump versions, as the changelog should look like as:
```
- Version 2021.10
  * Change A
  * Change B
  * Change C
```
Please agree with the squad what to do. In the end if you don't manage this at the `.changelog`, whoever prepare the SRs will need to adjust the format.

juliogonzalez (approved):
Assuming tables are OK. Please talk to your squad about the changelogs.

.changelog
0rnela: Changed. Thanks.

PR #1236, opened 2021-10-19
Add serving big files from proxy

jordimassaguerpla:
# Description

fixes bsc#1185465 by adding a section on how to serve big files from the proxy.

# Target branches

- [X] Master (Default)
- [X] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/14730

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -4,7 +4,8 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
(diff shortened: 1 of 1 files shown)


keichwa:
It is not clear to me what you want to accomplish.  Do you want to serve the files from the proxy or through the proxy?  I think you want to use the proxy as a file server.  In this case ou probably should clearly say:
```
On the PROXY_HOSTNAME system, copy the big files to the /srv/www/htdocs/pub directory.
```
I'd probably also provide minimal background info, and thus would write:
```
... big files such as ISO images to ...
```
A `modules/.changelog` entry is required with bug reference ;)

jordimassaguerpla:
> Thank you, Jordi! Good collaboration :)

you are welcome :+1:

PR #1296, opened 2021-11-18
Debian repositories #16335 keichwa

keichwa:
# Description

https://github.com/SUSE/spacewalk/issues/16335

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/modules/administration/pages/custom-channels.adoc
+++ b/modules/administration/pages/custom-channels.adoc
@@ -81,7 +81,8 @@ Add Debian non-flat repositories using ``uyuni_suite``, ``uyuni_component``, and
 is mandatory.
 In Debian documentation, this is also known as [literal]``distribution``.
 With this parameter you specify the apt source.
-Without this parameter or if it ends with a ``/``, the original approach is used (= flat repository).
+Without this parameter the original approach is used.
+If the parameter ends with ``/``, the repository is identified as flat.
 ``uyuni_component``::
 is optional.
 This parameter can specify only one component.


keichwa:
@vzhestkov I changed the wording a little bit.  Can you please check whether it still matches how the features is implemented?

vzhestkov (changes requested):
I think it's better to add some examples. For original URL formatting it's really unclear for the customers what URL should be specified. What it should points to.
In case of new approach it points to the root of the repo, but the original approach expecting pointing somewhere deep inside the repo, I'm not sure if I have proper words to describe where it should points to originally.

jcayouette (approved):
Looks good to me.

modules/administration/pages/custom-channels.adoc
```diff
@@ -71,6 +72,41 @@ For more information, see xref:administration:tshoot-registertradrhel.adoc[].
...
 
+[NOTE]
+====
+Add Debian non-flat repositories using ``uyuni_suite``, ``uyuni_component``, and ``uyuni_arch`` query parameters.
+====
+
+``uyuni_suite``::
+is mandatory.
+With this parameter you specify the apt source.
+Without this parameter the original approach is used.
```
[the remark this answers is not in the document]
keichwa: ok, I'm now mention it in my local tree.

modules/administration/pages/custom-channels.adoc
```diff
@@ -79,8 +79,9 @@ Add Debian non-flat repositories using ``uyuni_suite``, ``uyuni_component``, and
 
 ``uyuni_suite``::
 is mandatory.
+In Debian documentation, this is also known as [literal]``distribution``.
 With this parameter you specify the apt source.
-Without this parameter the original approach is used.
+Without this parameter or if it ends with a ``/``, the original approach is used (= flat repository).
```
vzhestkov, line 84: No, it's not true. The removed line was correct. Additional note about `/` at the end of `uyuni_suite` is a separate case when the repo is defined as a flat. The previous approach is trying to detect flat and non-flat repos by presence of Release file in certain place of hierarchy. (but not sure if we need to explain it. So:
```suggestion
Without this parameter the original approach is used.
If the parameter ends with ``/``, the repository is identified as flat.
```

vzhestkov (approved):
Just one correction. Everything else is good. Approved, but please fix the pointed wrong sentance.

modules/administration/pages/custom-channels.adoc
keichwa: yes, now I see.

PR #1332, opened 2021-12-13
Replaced obsolete patch instruction with update

digdilem:
# Minor update instructions were misleading and inaccurate. 

As mentioned on the uyuni-users mailing list, the old "zypper patch" instructions no longer appear to work. 

This edit attempts to improve upon that whilst still keeping things succinct.

### DIFF
--- a/modules/upgrade/pages/server-minor-upgrade-uyuni.adoc
+++ b/modules/upgrade/pages/server-minor-upgrade-uyuni.adoc
@@ -46,7 +46,7 @@ zypper up
 ----
 +
 
-. If zypper reports that the Uyuni package won't be upgraded, run that manually:
+. If Zypper reports that the Uyuni package will not be upgraded, run the following command manually:
 +
 
 ----


keichwa:
We probably did not describe clear enough what this section is meant to cover.  Here we really only want tell how the user can apply just (openSUSE) patches.  These are OS security fixes that are independently released from uyuni releases.

Maybe I got it wrong.  Before taking acation, I must check the ML discussion.

digdilem:
> Here we really only want tell how the user can apply just (openSUSE) patches

But this is the page to **perform a minor upgrade of Uyuni** - it's linked and described as such, so my feeling is that it should really describe the process to upgrade Uyuni - which merely applying patches does not.  

As they were, the process described does not perform a minor upgrade.

I welcome other thoughts from the review.

keichwa:
Yes, you are not wrong `;)`  Maybe, we should rename the file from "*minor" to "*normal" and give an example such as upgrade from 2021.11 to 2021.12. Maybe, we can also keep the contents of the old file and change the title somehow ("Security fixes inbetween releases", or sth similar).  Yes, we need more opinions.  Or I should really read the MU discussion now...

modules/upgrade/pages/server-minor-upgrade-uyuni.adoc
```diff
@@ -35,26 +37,23 @@ zypper ref
...
 +
 
 ----
-zypper list-patches
+zypper up
 ----
 +
 
-. Apply all available patches:
+. If zypper reports that the Uyuni package won't be upgraded, run that manually:
```
jcayouette, line 49: "If **Zypper** reports that the Uyuni package will not be upgraded, run the following command manually:"

jcayouette (changes requested):
See comment.

digdilem:
Thanks for the improved wording @jcayouette  - updated file with your suggestion.

keichwa (approved):
Still not convinced, but approving nevertheless (it is probably more useful than before).

0rnela:
@digdilem, thank you for this PR. Due to recent changes in books organisation, new PR has replaced this one, with your contributions added there.

PR #1348, opened 2022-01-04
Retail versions #1193292 keichwa

keichwa:
# Description

Fix Retail related version numbers (4.2 instead of 4.1).

https://bugzilla.suse.com/show_bug.cgi?id=1193292

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

https://github.com/SUSE/spacewalk/issues/16474


keichwa:
Eric (the reporter of the original bug) proposed to add information about 4.1 and 15SP2 as currently also supported.  I think this is not needed.  If you think different, I can do it nevertheless.  Now I'll provide a .changelog entry...

aaannz (approved):

PR #1356, opened 2022-01-06
Gpg custom channel #15369 keichwa

keichwa:
# Description

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

https://github.com/SUSE/spacewalk/issues/15369


keichwa:
I hope this helps with clarifying custom channel and GPG importing issue.

modules/administration/pages/custom-channels.adoc
```diff
@@ -25,7 +25,18 @@ Many third party vendors secure packages with GPG.
...
+
+If remote channels and repositories are signed with GPG keys, you can import and trust these GPG keys.
+For example, execute the [command]``spacewalk-repo-sync`` from the command line on the {productname} Server:
+
+----
+/usr/bin/spacewalk-repo-sync -c <channellabelname> -t yum
+----
+
+The underlying [command]``zypper`` call will import the key, if it is available.
+The {webui} does not offer this feature.
```
mcalmer, line 39: This works only, when the repository you want to mirror is setup in a special way and provide the "key" in the repo next to the signature. This is the case for all repos generated by OBS, but not typical for other repos from other OSes.

For them you need to manually import the GPG key into the keyring.

It should be possible to import it into `/var/lib/spacewalk/gpgdir` keyring using gpg commandline tool.
This would be kind of permanent and the key would also persist if the chroot env would be cleaned.
Not sure how it is possible to import the key into the RPM DB which is used for syncing.
@meaksh is this correct and what would you do?

The gpg command to import a key into the pub keyring is:
```
$> gpg --keyring /var/lib/spacewalk/gpgdir  --import /path/to/gpg.key
$> gpg --edit-key <keyid>
```
Using gpg is horrible. You need to find out the keyid, you need to trust "ultimately" and save the change.

mcalmer (changes requested):

modules/administration/pages/custom-channels.adoc
keichwa: I added this info.  I'd say it is ready for review.  Maybe, we can add more info/details later.

keichwa:
Hi reviewers, is sth still missing?  If yes, I need your input please.

0rnela (approved):

PR #1381, opened 2022-01-24
remove old installation and upgrade guide

mcalmer:
# Description

After merging installation and upgrade guide into 1 new book, we can remove the old translation components.

**Merge only after we clarified the future of the upgrade weblate component**

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0


jcayouette (approved):
Good but we need to make sure the weblate changes do not conflict before merging.

PR #1400, opened 2022-02-08
Salt minion instsource #16689 keichwa

keichwa:
# Description

Add a warning about the origin of the salt-minion package in the
Register on the Command Line (Salt) section of the Client Configuration Guide
https://github.com/SUSE/spacewalk/issues/16689

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/modules/client-configuration/pages/registration-cli.adoc
+++ b/modules/client-configuration/pages/registration-cli.adoc
@@ -22,6 +22,12 @@ This procedure requires that you have installed the [package]``venv-salt-minion`
 Both use configuration files in different locations and filenames remain the same.
 The systemd service filename is different.
 
+[WARNING]
+====
+Bootstrapping this way will only work if you use the [package]``salt-minion`` being part of the client tools channels or of an official {suse} distributions.
+====
+
+
 Salt Bundle ([package]``venv-salt-minion``)::
 +
 * [path]``/etc/venv-salt-minion/``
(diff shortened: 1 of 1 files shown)


jcayouette (approved):

modules/client-configuration/pages/registration-cli.adoc
```diff
@@ -80,6 +80,11 @@ Salt ([package]``salt-minion``)::
 * [path]``/etc/salt/minion.d/NAME.conf``
 * systemd service file: [path]``salt-minion.service``
 
+[WARNING]
+====
+Bootstrapping this way will only work if you use the [package]``salt-minion`` being part of the client tools channels or of an official {suse} distributions.
+====
+
```
meaksh, line 87: I see this is included inside `.Procedure: Registering Clients with Salt Bundle Configuration File` but this also applies for the other procedure in this page (`.Procedure: Registering Clients with Salt Minion Configuration File`).

I would rather put this warning message before starting with the different procedures, so around line 34 of this page.

meaksh (changes requested):

modules/client-configuration/pages/registration-cli.adoc
keichwa: I think in this case it would make sense to to have the warning even a little bit earlier.  Here is a new try now

meaksh (approved):

PR #1485, opened 2022-03-17
Salt Bundle: wording improvements

keichwa:
# Description

Cosmetic wording improvements.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

### DIFF
--- a/modules/client-configuration/pages/contact-methods-saltbundle.adoc
+++ b/modules/client-configuration/pages/contact-methods-saltbundle.adoc
@@ -1,14 +1,18 @@
 [[contact-methods-salt]]
 = Salt Bundle
 
+
+
 == What is Salt Bundle?
 
-The Salt Bundle is a single binary package containing Salt Minion, Python 3, additional Python modules, and libraries.
+The Salt Bundle is a single binary package containing Salt Minion, Python 3, required Python modules, and libraries.
 
-The Salt Bundle is shipped with Python 3 and all the requirements for Salt to run. Thus Salt Bundle does not use the Python version install on the client as system software.
+The Salt Bundle is shipped with Python 3 and all the requirements for Salt to run. Thus Salt Bundle does not use the Python version install on the client as system software.
 The Salt Bundle can be installed on clients that do not meet the requirements for a given Salt version.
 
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -3,11 +3,12 @@
 
 == What is Salt Bundle?
 
-The Salt Bundle is a single binary package containing Salt Minion, Python 3, required Python modules and libraries.
+The Salt Bundle is a single binary package containing Salt Minion, Python 3, additional Python modules, and libraries.
```
vzhestkov, line 6: Not sure if it's a proper rewording as there is only the modules required for salt, what are the modules additional to? We just imlementing the ways to add additional modules inside with `pip`.

vzhestkov (changes requested):
@keichwa thanks for improving it.
Just one point I don't like in the comments. I would suggest not to touch this part yet as there are really no any additionals but just only required modules for Salt.

jcayouette (changes requested):
Please fix the vale style check item. :) Thanks!

modules/client-configuration/pages/contact-methods-saltbundle.adoc
keichwa: ok, I reverted this

vzhestkov (approved):
Thanks :+1:

keichwa:
> Please fix the vale style check item. :) Thanks!

this requires an own PR: https://github.com/uyuni-project/uyuni-docs/pull/1487

keichwa:
> > Please fix the vale style check item. :) Thanks!
> 
> this requires an own PR: #1487

I merged that one.

keichwa:
@vzhestkov I forgot to ask... Do we want this in 4.2, too?

vzhestkov:
If I remember right the plan is to add the support for of Salt Bundle to 4.2 after releasing 4.3 to make it possible to use the same client tools.
@admd am I right?

admd:
Yes, that is correct Victor @vzhestkov

PR #1489, opened 2022-03-21
docs: Correct cobbler's buildiso command

nodeg:
# Description

This will include the missing `--distro` parameter cobbler `buildiso` needs to work.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Fixes https://github.com/SUSE/spacewalk/issues/17308

CC @SchoolGuy

### DIFF
--- a/modules/client-configuration/pages/autoinst-cdrom.adoc
+++ b/modules/client-configuration/pages/autoinst-cdrom.adoc
@@ -30,9 +30,8 @@ In order to prepare an ISO image with Cobbler, you need to prepare a distributio
 For information about creating a distribution, see xref:client-configuration:autoinst-distributions.adoc[Autoinstallable Distributions].
 For information about creating a profile, see xref:client-configuration:autoinst-profiles.adoc[Autoinstallation Profiles].
 
-The Cobbler [command]``buildiso`` command takes parameters to define the name and output location of the boot ISO. A distribution
-has to be provided in all cases with [option]``--distro``. 
-
+The Cobbler [command]``buildiso`` command takes parameters to define the name and output location of the boot ISO.
+Specifying the distribution with [option]``--distro`` is mandatory.
 For example:
 
 ----


nodeg:
Maybe we should

- add a reference to the upstream documentation regarding the `buildiso` command.
- mention that when running Cobbler via systemd, you cannot mount the ISO or define the buildiso dir inside `/tmp` because of the used systemd security hardening option that is in place
> When running Cobbler via systemd, you cannot mount the ISO to /tmp or a sub-folder of it because we are using the option Private Temporary Directory, to enhance the security of our application.
We only added the following note:
image

I think we should explicitly note or warn that `/tmp` cannot be used with Cobbler.

modules/client-configuration/pages/autoinst-cdrom.adoc
```diff
@@ -30,11 +30,13 @@ In order to prepare an ISO image with Cobbler, you need to prepare a distributio
 For information about creating a distribution, see xref:client-configuration:autoinst-distributions.adoc[Autoinstallable Distributions].
 For information about creating a profile, see xref:client-configuration:autoinst-profiles.adoc[Autoinstallation Profiles].
 
-The Cobbler [command]``buildiso`` command takes parameters to define the name and output location of the boot ISO.
+The Cobbler [command]``buildiso`` command takes parameters to define the name and output location of the boot ISO. A distribution
+has to be provided in all cases with [option]``--distro``. 
+
```
keichwa, line 35:
```suggestion
The Cobbler [command]``buildiso`` command takes parameters to define the name and output location of the boot ISO.
Specifying the distribution with [option]``--distro`` is mandatory.
```

keichwa: For your background info: We (unfortunately) have 1-sentence-per-line policy.

keichwa (changes requested):

modules/client-configuration/pages/autoinst-cdrom.adoc
nodeg: I somehow did not find that on the [wiki](https://github.com/uyuni-project/uyuni-docs/wiki/wug) but good to know. Thank you @keichwa.

jcayouette (approved):

jcayouette:
Looks good, @nodeg Thanks.

keichwa (approved):

PR #1505, opened 2022-04-06
registration of SLE 12 clients

keichwa:
# Description

https://github.com/SUSE/spacewalk/issues/14610

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


keichwa:
@juliogonzalez I only added info about registration of sle 12 sp5 clients. If this is not enough, I could add more. We also do not list sle 15 [sp0].

Another thing is that we only list the `x86_64` architecture.  ATM, I do not know whether we have a statements about archs somewhere in our guides...

juliogonzalez:
> @juliogonzalez I only added info about registration of sle 12 sp5 clients. If this is not enough, I could add more. We also do not list sle 15 [sp0].

Should be OK, but I'd mention that other SPs can be added by replacing the number (and we can leave SP0 out of the story), as long as they are supported.

> Another thing is that we only list the `x86_64` architecture. ATM, I do not know whether we have a statements about archs somewhere in our guides...

Yes, we have. Check the client guide itself. There's "Supported Clients and Features"

keichwa:

> Should be OK, but I'd mention that other SPs can be added by replacing
> the number (and we can leave SP0 out of the story), as long as they
> are supported.

Good.  I just noticed that we already have this statement here:

> You can also use these instructions for older {sle} versions and
> service packs.

I think it is good enough.

-- 
Karl Eichwalder

SUSE Software Solutions Germany GmbH, Geschäftsführer: Ivo Totev
HRB 36809, AG Nürnberg  Maxfeldstr. 5,  D-90409 Nürnberg

juliogonzalez:
> Julio González Gil ***@***.***> writes:
> Should be OK, but I'd mention that other SPs can be added by replacing the number (and we can leave SP0 out of the story), as long as they are supported.
> Good. I just noticed that we already have this statement here:
> You can also use these instructions for older {sle} versions and service packs.
> I think it is good enough.

Not related to this PR, but I find the current text confusing, as it says:

> The products you need for this procedure are:

And...

> Alternatively, you can add channels at the command prompt.
> The channels you need for this procedure are:

And then gives you `x86_64`.

Maybe it should say something like "are the following for x86_64, adjust the list with the right architecture as needed" :-)

jcayouette (approved):

0rnela (approved):

keichwa:
> > Julio González Gil _**@**_.***> writes:

> Maybe it should say something like "are the following for x86_64, adjust the list with the right architecture as needed" :-)

Yes!  We now track it here: https://github.com/SUSE/spacewalk/issues/17459

PR #1513, opened 2022-04-07
add section about tuning parameters for report db synchronization

mcalmer:
# Description

Add some information about tuning report db sync

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -4,6 +4,8 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- In the Large Deployments Guide, enhance section about hub reporting to add tuning options
+  for the taskomatic jobs
 - In the Large Deployments Guide, enhance section about hub reporting with taskomatic jobs
 - In Client Configuration Guide, added information about registration of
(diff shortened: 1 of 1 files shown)


modules/specialized-guides/pages/large-deployments/hub-reporting.adoc
```diff
@@ -78,3 +78,37 @@ All times are in the local timezone of the server.
...
+[cols="1,1"]
+|===
+| Description          | The maximum number of workers requesting data from peripheral servers on a hub at the same point in time.
+| Tune when            | The number of peripheral servers increases significantly (more than 100), or a faster synchronization is required.
+| Value default        | 2
+| Value recommendation | 2 - 10
+| Location             | [path]``/etc/rhn/rhn.conf``
+| Example              | `report_db_hub_workers = 5`
+| After changing       | Check xref:memory-usage[memory usage].
+                         Monitor memory usage closely before and after the change.
```
mackdk, line 110: Maybe should we mention that the cpu usage during the synchronization should also be checked? If the cpu is already maxed out during the process they will not gain anything from increasing the number of workers. In fact in that case they might achieve better performance by lowering it.

mackdk (approved):

0rnela (approved):

PR #1577, opened 2022-05-24
note about importing CA certifcate in Installation and Upgrade Guide

keichwa:
# Description

https://bugzilla.suse.com/show_bug.cgi?id=1198358
https://github.com/SUSE/spacewalk/issues/17513

Followup: https://github.com/uyuni-project/uyuni-docs/pull/1601

# Target branches

- [x] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


keichwa:
Maybe, I should better add this statement to a general trouble shooting section and reference it from the server setup sections?

0rnela (approved):

jcayouette (approved):

modules/installation-and-upgrade/pages/server-setup.adoc
```diff
@@ -75,6 +75,18 @@ The main administration account is the _highest authority account_ within {produ
 For security it is recommended that the main administrator creates _low level admin accounts_ designated for administration of organizations and individual groups.
 ====
 
+[NOTE]
+====
+Newer browser versions can block web access to the {productname} Server FQDN because of HSTS.
```
renner, line 80: Remember that we decided to disable HSTS per default, so you would run into this problem only in case HSTS was manually enabled from the config file. Not sure if you maybe want to change the wording here, e.g.:

```
Newer browser versions can block web access to the {productname} Server FQDN in case HSTS was enabled.
```

renner (commented):
Sorry for the late review, just a small note / suggestion, but it's already good as it is.

modules/installation-and-upgrade/pages/server-setup.adoc
keichwa: ok, thanks for your follow-up!  I created a new PR with additional small modifications for clarity (hopefully):

https://github.com/uyuni-project/uyuni-docs/pull/1601

PR #1593, opened 2022-05-31
An issue template for the recipe for common workflows book

admd:
# Description

This is to make it easier for contributors. Cloning repo, editing, and then submitting the PRs is asking for a lot and creates a lot of overhead which doesn't help us get the contributions at the end.

It will create a bit of overhead for because we will need to convert the contribution to asciidoc format but the effort is worth it if we can get more contributions.

# Target branches

- [x] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


0rnela:
Makes sense. Thanks.

PR #1631, opened 2022-06-27
Reactivation key step during Uyuni proxy upgrade not needed anymore

phibid:
# Description

Removed reactivation key steps in the Uyuni proxy major upgrade procedure

# Target branches

- [X] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links


mbussolotto (changes requested):
we should remove also the re-register part (few lines below)
```
. Re-register with the reactivation key using the [command]``rhnreg_ks`` command-line utility.	
  The system is re-registered with the same ID, history, and groups.
```

mbussolotto (approved):
LGTM, thanks!

keichwa (approved):

PR #1665, opened 2022-07-19
4.2: Add missing part of enabling optional Salt Bundle with Salt SSH

vzhestkov:
# Description

Add missing part of the documentation to enable optional support of the Salt Bundle with Salt SSH.

It makes sense only for optional enabling it in `4.2`, so no need to port it to any other branches.

# Target branches

- [ ] Master (Default)
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Found during the discussion of the issue: https://github.com/SUSE/spacewalk/issues/18367

### DIFF
--- a/modules/client-configuration/pages/contact-methods-saltbundle.adoc
+++ b/modules/client-configuration/pages/contact-methods-saltbundle.adoc
@@ -96,7 +96,7 @@ web.ssh_salt_pre_flight_script = /usr/share/susemanager/salt-ssh/preflight.sh
 
 web.ssh_use_salt_thin = false
 ----
-Create the extra drop-in configuration file for Salt Master [path]``/etc/salt/master.d/ssh-preflight.conf``:
+Create the extra drop-in configuration file for Salt Master [path]``/etc/salt/master.d/ssh-preflight.conf`` with [literal]``ssh_run_pre_flight`` set to [literal]``true``:
 ----
 ssh_run_pre_flight: true
 ----


keichwa (approved):
Maybe, for clarity you could add "with [literal]``ssh_run_pre_flight`` set to [literal]``true``.".  I did not check the context.  My enhancement suggestion could be superfluous.

vzhestkov:
@keichwa do you mean not to put the file content example, but just mention which parameter and value should be specified?
For the salt master config it could be a bit confusing as there is `rhn.conf` with the different format mentioned above, so the user could try to put `ssh_run_pre_flight = true` instead of `ssh_run_pre_flight: true`

modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -96,6 +96,11 @@ web.ssh_salt_pre_flight_script = /usr/share/susemanager/salt-ssh/preflight.sh
 
 web.ssh_use_salt_thin = false
 ----
+Create the extra drop-in configuration file for Salt Master [path]``/etc/salt/master.d/ssh-preflight.conf``:
```
keichwa, line 99: I meant in addition.  But feel free to ignore it. I'm not sure whether will cause more trouble than good:
```suggestion
Create the extra drop-in configuration file for Salt Master [path]``/etc/salt/master.d/ssh-preflight.conf`` with [literal]``ssh_run_pre_flight`` set to [literal]``true``:
```

vzhestkov: Oh, this way. Looks good to me :+1:

PR #1674, opened 2022-07-25
Add Salt Bundle functionality extension with PIP modules

vzhestkov:
# Description

Adds the note about extending the Salt Bundle functionality with `PIP` modules

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [x] Manager-4.2
- [ ] Manager-4.1
- [ ] Manager-4.0

# Links

Tracks https://github.com/SUSE/spacewalk/issues/17259

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -4,7 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
-- Added Extending Salt Bundle functionality with Python packages using pip
+- Added Extend Salt Bundle functionality with Python packages using pip
 - Salt Configuration Modules are no longer Technology Preview in Salt Guide.
 - Described disabling local repositories in Client Configuration Guide
 - Documented Ubuntu 22.04 LTS as a supported client OS in Client Configuration Guide


SchoolGuy (approved):
LGTM. To understand the implications of the Note and the Important section one has to read it a few times but I am not aware of an easier way to describe the mechanism.

modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -94,3 +94,24 @@ Using `salt-thin` can be enabled as a fallback method, but it requires Python 3
...
+
+== Salt Bundle functionality extension with PIP modules
+
+The Salt Bundle includes [literal]``PIP`` Python module to make it possible to extend the functionality of the bundled Salt Minion with extra Python modules.
+
+The default location to deploy modules on calling [command]``salt <minion_id> pip.install <module_name>`` is [path]``/var/lib/venv-salt-minion/local``.
+
+[NOTE]
+====
+In case if [path]``/var/lib/venv-salt-minion/local`` is not applicable for some reasons and the other location for the modules is required the destination path can be overridden by setting [literal]``VENV_PIP_TARGET`` environment variable for the [literal]``venv-salt-minion.service``.
```
agraul, line 106:
```suggestion
If needed, the path [path]``/var/lib/venv-salt-minion/local`` can be overridden by setting the [literal]``VENV_PIP_TARGET`` environment variable for the [literal]``venv-salt-minion.service``.
```

modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -94,3 +94,24 @@ Using `salt-thin` can be enabled as a fallback method, but it requires Python 3
...
+It is recommended to use systemd drop-in config file for the service. It could be done with the config file [path]``/etc/systemd/system/venv-salt-minion.service.d/10-pip-destination.conf``:
+----
+[Service]
+Environment=VENV_PIP_TARGET=/new/path/local/venv-salt-minion/pip
+----
+====
+
+[IMPORTANT]
+====
+The Salt Bundle does not maintain the modules installed with [literal]``PIP``. In case of updating the Salt Bundle it is possible that modules should be reinstalled, for example on updating the Salt Bundle with updated Python inside.
```
agraul, line 116:
```suggestion
The Salt Bundle does not keep Python packages installed through [literal]``pip`` on updates. To ensure that such packages are available after an update, it is recommended to install them with a Salt state that is applied after Salt Bundle updates.
```

agraul (changes requested):
Thanks for putting this together, this looks quite good already. I would like us to change the wording from "PIP modules" to "Python packages installed with/through/via pip" because that's closer to the wording that pip and PyPI (Python *Package* Index) use.

I left some suggestions that take care of this change. As the name suggests, those are only suggestions. Feel free to dismiss them if you prefer another way.

.changelog
```diff
@@ -4,6 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Changelog_section_.28.25changelog.29
 
+- Added Salt Bundle functionality extension with PIP modules
```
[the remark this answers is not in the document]
vzhestkov: Yup it's clear with pip, but I'm a bit confused with word ordering `functionality extension` => `extension functionality`. Your suggestion sounds more like we are talking about some Salt Bundle extension, but not functionality extension.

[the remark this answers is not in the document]
agraul: The way I see it is that we're adding the functionality to extend the Bundle. "functionality extension" focuses on the extension of the Bundle's functionality, which is slightly different. Enabling pip is also a "functionality extension", but IMO that's not the focus of this documentation.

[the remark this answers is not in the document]
vzhestkov: `Salt Bundle extension functionality with Python packages through pip` for me sounds more confusing than `Salt Bundle functionality extension with Python packages through pip`. Maybe `Extending Salt Bundle functionality with Python packages using pip`?

keichwa:
Thanks for all your work of all of you.  I think most of Alexander's suggestions make sense, but please decide on your own. When all suggestions are evaluated I'll read and review the whole file.

.changelog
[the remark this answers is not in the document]
agraul: Sounds good to me :+1:

modules/client-configuration/pages/contact-methods-saltbundle.adoc
keichwa: It is probably not needed to explain the implementation too detailed.  I'd basically say:

> The Salt Bundle is self-contained and works independent from installed system components such as the Python interpreter and the Python modules.
> The Salt Bundle does neither update nor uninstall a system component such as Python.

I'm not sure whether we must explicitly say that updating Python will happen in the run of regular system updates.

agraul: The point here is that a user can install additional Python packages into the bundle, but those aren't updated when everything else in the bundle (Python, Salt, the Python packages we include) is updated. 

I previously thought that a Salt Bundle update would remove these additionally installed Python packages (which might still happen once we tackle https://confluence.suse.com/display/SUSEMANAGER/Problems+around+updating+Salt+Minion+%28bundled%29+with+Salt), but apparently that's not the case now. Either way, while these additional packages are placed into the bundle, it needs to be clear that the bundle is not actively doing anything with them.

vzhestkov: Is it sounds good if reword it this way?
```suggestion
The Python packages installed through [literal]``pip`` are not changing on updating the Salt Bundle. To ensure that such packages are available and functional after an update, it is recommended to install them with a Salt state that is applied after Salt Bundle updates.
```

agraul: Sounds good!

vzhestkov:
@agraul thanks for the suggestions :+1: could you please approve if it's fine now
@keichwa I think all the suggestions were addressed

agraul (approved):
Looks good, thanks a lot for your work @vzhestkov

modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -94,3 +94,24 @@ Using `salt-thin` can be enabled as a fallback method, but it requires Python 3
 This method is not recommended nor supported and exists for development purposes only.
 Set [option]``web.ssh_use_salt_thin`` to `true` in the [path]``/etc/rhn/rhn.conf`` configuration file.
 ====
+
+== Extending Salt Bundle functionality with Python packages using pip
```
keichwa, line 98: I'm sorry that I must start with this again.
```suggestion
== Extend Salt Bundle with Python packages using pip
```
In titles we prefer verbal form.  I think this is short for "How to extend the Salt Bundle ...".  The -ing form is reserved for procedure titles.  That's how Lana argued.  I'd like to stay with this style.  If you agree here, we also should change the .changelog entry.

modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -94,3 +94,24 @@ Using `salt-thin` can be enabled as a fallback method, but it requires Python 3
...
+
+The Salt Bundle includes [literal]``pip`` to make it possible to extend the functionality of the bundled Salt Minion with extra Python packages.
```
keichwa, line 100:
```suggestion
The Salt Bundle includes [literal]``pip`` to make it possible to add features to the Salt Bundle connection method with extra Python packages.
```
Your version can be read as if this is a possibility to add Python packages to the Client in general.

By all means, we must avoid the term "minion".  We prefer the term "client" here.

Last I'm wondering whether we should say "Python packages" or "Python modules" (pip has probably something to do with packages).

keichwa (approved):

modules/client-configuration/pages/contact-methods-saltbundle.adoc
vzhestkov: No, now it's not correct. Salt Bundle is not a connection method by itself, it's more like a different flavour of Salt Minion.

Python package here is not the RPM or DEB package of the linux, but the package of Python (a set of modules). And we are talking here only about extending the bundled python with extra packages which will be available for the bundled python and salt bundle (bundled salt minion to be exact).

```suggestion
The Salt Bundle includes [literal]``pip`` to make it possible to extend the functionality of the Salt Bundle with extra Python packages.
```
But here I would like to avoid to mention Salt Bundle twice in the same sentance, so in original I've replaced the second occurance with the bundled Salt Minion as it's actually the same in this context.

modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -94,3 +94,24 @@ Using `salt-thin` can be enabled as a fallback method, but it requires Python 3
...
+It is recommended to use systemd drop-in config file for the service. It could be done with the config file [path]``/etc/systemd/system/venv-salt-minion.service.d/10-pip-destination.conf``:
+----
+[Service]
+Environment=VENV_PIP_TARGET=/new/path/local/venv-salt-minion/pip
+----
+====
+
+[IMPORTANT]
+====
+The Python packages installed through [literal]``pip`` are not changing on updating the Salt Bundle. To ensure that such packages are available and functional after an update, it is recommended to install them with a Salt state that is applied after Salt Bundle updates.
```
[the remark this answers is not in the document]
vzhestkov: It's not strange, but I have no idea about it. Thanks for pointing.

vzhestkov:
@keichwa thanks for the comments. Only one point left unresolved. I would like to leave it as it is now, unfortunately your suggestion is making the note confusing.

modules/client-configuration/pages/contact-methods-saltbundle.adoc
keichwa: thanks, I think I now understand it.  We can close this as resolved.

keichwa:
I think it is now ready to be merged and then backported.

PR #1718, opened 2022-08-12
Converted the admonition about type of managed clients to plain text.

0rnela:
# Description

Admonition cleanup - we have an extremely high number of admonitions per client registration page, thus defeating their own purpose. 

In this PR, the admonition about type of managed clients has been converted to plain text in order to visually, and effectively, enhance the page and help the user distinguish the basic and necessary information from what is truly deserving of being an admonition, i.e. calling for special attention.
 
# Target branches

- [x] Master (Default)
- [ ] Manager-4.3
- [ ] Manager-4.2

# Links

Addresses: https://github.com/SUSE/spacewalk/issues/18609


jcayouette (approved):
Looks ok. I'm not sure how I feel about removing admonitions but I agree we should stick to a standard convention.

0rnela:
> Looks ok. I'm not sure how I feel about removing admonitions but I agree we should stick to a standard convention.

When we have too many of them, they stop serving the purpose. A review of the existing ones in the Cl Conf Guide was quite ovredue.

PR #1734, opened 2022-08-23
Container proxy FQDN important note

cbosdo:
# Description

Make it very clear that the container proxy has to have a unique FQDN

# Target branches

- [X] Master (Default)
- [X] Manager-4.3
- [ ] Manager-4.2

# Links

Fixes

### DIFF
--- a/modules/installation-and-upgrade/pages/proxy-container-installation.adoc
+++ b/modules/installation-and-upgrade/pages/proxy-container-installation.adoc
@@ -45,8 +45,13 @@ endif::[]
 
 [IMPORTANT]
 ====
-Ensure the container proxy has a unique FQDN. This should NOT be the same as the FQDN of the container host host, and should be resolvable for any assigned clients of the proxy.
+The container proxy must have a unique Fully Qualified Domain Name (FQDN).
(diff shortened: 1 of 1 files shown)


modules/installation-and-upgrade/pages/proxy-container-installation.adoc
```diff
@@ -43,6 +43,10 @@ endif::[]
 
 |===
 
+[IMPORTANT]
+====
+Ensure the container proxy has a unique FQDN. This should NOT be the same as the FQDN of the container host host, and should be resolvable for any assigned clients of the proxy.
+====
```
keichwa, line 49: I propose reusing phrases from the `network-requirements.adoc` file:
```suggestion
The container proxy must have a unique Fully Qualified Domain Name (FQDN).
This should not be the same as the FQDN of the container host.
To ensure that the {productname} domain name can be resolved by its clients, both container proxy and client machines must be connected to a working DNS server.
You also need to ensure that reverse lookups are correctly configured.
====


```

keichwa (approved):
I guess we should backport this admonition.

PR #1695, opened 2022-07-30
Add Rocky Linux 9 client support documentation (#1661)

lumarel:
# Description

**~Blocked until https://github.com/uyuni-project/uyuni/pull/5682 is merged.~** [merged]

This PR documents the upcoming support for Rocky Linux 9, it should be in with version 2022.08 if I understood that correctly.

A few notes:

- it mentions that ppc64le and s390x is supported, this will be needed to be removed if the venv doesn't get built for these platforms
- it also removes ppc64le support from Rocky Linux 8, as this arch never took off in the project for 8
- I did not adjust the support tables for SUSE Manager, because I don't have insight in the SUSE internas

Beside that, Rocky Linux 9 is pretty much the same story as 8 in Uyuni, the only difference is, that there will not be a normal build for the salt-minion only a venv-salt-minion, but I don't think this needs to be promoted on the individual registration page as this is a overall topic, where as far as I understood it all clients will be moved to venv only.

# Target branches

Which documentation version does this PR apply to?

- [x] Master (Default)
- [ ] Manager-4.3
- [ ] Manager-4.2

# Links

Fixes #1661

PR #1735, opened 2022-08-24
Add Rocky Linux 9 client support documentation (#1695)

keichwa:
# Description

- [x] port missing bits to `master`. (done, imo)
- [x] More SLE adjustments are needed. (done)

https://github.com/SUSE/spacewalk/issues/18472

# Target branches

- [ ] Master (Default)
- [x] Manager-4.3
- [ ] Manager-4.2

# Links


modules/client-configuration/pages/clients-rocky.adoc
```diff
@@ -83,7 +81,7 @@ endif::[]
 ifeval::[{uyuni-content} == true]
 [IMPORTANT]
 ====
-For {rocky} 8 clients, add both the Base and AppStream channels.
+For {rocky} 8 and 9 clients, add both the Base and AppStream channels.
```
0rnela, line 84:
```suggestion
For {rocky} 8 and {rocky}9 clients, add both the Base and AppStream channels.
```

modules/client-configuration/pages/clients-rocky.adoc
```diff
@@ -97,7 +95,7 @@ This is due to the way that {rocky} manages their repositories.
 {rocky} removes older version of packages when a new version is released, while {productname} keeps all of them, regardless of age.
 ====
 
-If you are using modular channels, you must enable the Python 3.6 module stream on the client.
+If you are using modular channels with {rocky} 8, you must enable the Python 3.6 module stream on the client.
```
0rnela, line 98: Why is only Rocky 8 specifically mentioned here? How or are things different for Rocky 9?

keichwa: iirc, this has something to do with the Salt bundle. But I admit, I'm not 100% sure...

juliogonzalez: I don't think this is needed. There's no such module on Rocky 9. @vzhestkov can confirm, just in case.

modules/client-configuration/pages/clients-rocky.adoc
```diff
@@ -21,7 +18,7 @@ You do not need to disable SELinux to register {rocky} clients to {productname}.
 
 Before you can register {rocky} clients to your {productname} Server, you need to add the required software channels, and synchronize them.
 
-The architectures currently supported are: `x86_64` and `aarch64`.
+The architectures currently supported are: `x86_64` and `aarch64`, partially {ppc64le} and {s390x}.
```
juliogonzalez, line 21: Why partially?

modules/client-configuration/pages/supported-features.adoc
```diff
@@ -263,8 +270,13 @@ ifeval::[{uyuni-content} == true]
 | {check}
 | {check}
 
+| {rocky} 9
+| {x86_64}, {aarch64}, {ppc64le}, {s390x}
+| {cross}
+| {check}
+
 | {rocky} 8
-| {x86_64}, {ppc64le}, {aarch64}
```
juliogonzalez, line 267: Nice that you saw this. ppc64le is not supported for 8, indeed.

modules/client-configuration/pages/clients-rocky.adoc
vzhestkov: I don't know the reason of this note. Salt bundle will not use any of Python 3.6 modules from the system. Maybe ask Can, if I'm not wrong he was developing modular support.

modules/client-configuration/pages/clients-rocky.adoc
keichwa: Maybe, for clarity I should better write "On version 9, also {ppc64le} and {s390x} are supported."

modules/client-configuration/pages/clients-rocky.adoc
keichwa: somewhere else Can said, from devel POV both version are handled the same.  Thus I assume that we can drop the statement about Python 3.6 here completely.  If nobody vetoes, I'll remove it soon.

PR #1763, opened 2022-09-15
Explicit info about EOL clients.

0rnela:
# Description

Proposal for adding explicit info about EOL clients. 

Pro: Rather than using admonition, giving these clients separate section makes them more obvious to the client.

Con: contains explicit dates in the new table.

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [x] Manager-4.2

# Links

Fixes https://github.com/SUSE/spacewalk/issues/18168


keichwa (approved):
ok, but we must find a way to deal with these support statements differently.

keichwa (approved):

PR #1792, opened 2022-09-29
Warn about ownership in Admin

keichwa:
# Description

https://bugzilla.suse.com/show_bug.cgi?id=1202612
https://github.com/SUSE/spacewalk/issues/19061

# Target branches

- [x] Master (Default)
- [ ] Manager-4.3
- [ ] Manager-4.2

# Links


jcayouette (approved):

0rnela (approved):

modules/administration/pages/backup-restore.adoc
```diff
@@ -411,6 +411,13 @@ This section assumes you have setup your server using separate partitions for th
 
 
 
+[IMPORTANT]
+====
+When you backup files and directories, the backup tools store the user and group names of the filenames.
+The tools will automatically use the right IDs when you extract the files on the target system again.
+But if you mount existing partitions, you must align the ownership to the new system.
```
mcalmer, line 418: I would write this different as I do not know if all backup systems really store "names" and not "ids".
I know "tar" is doing it. But I never touched a real backup system.

```suggestion
After a new installation of a system most users and groups get different ids.
Most backup systems store the names instead of the IDs and will restore the files with the correct ownership and permissions.
But if you mount existing partitions, you must align the ownership to the new system.
```

mcalmer (changes requested):

modules/administration/pages/backup-restore.adoc
keichwa: thanks, makes sense to me.

PR #1858, opened 2022-10-31
Extend note about using Salt SSH with Salt Bundle in 4.2

ycedres:
# Description

For Manager-4.2, the Important Note about using Salt SSH with Salt Bundle was not complete as one mandatory step was missing. For Manager 4-3 this is not necessary because  it has a different default behaviour.

In this PR I extended that note and included the missing step.

# Target branches

- [ ] Master (Default)
- [ ] Manager-4.3
- [X] Manager-4.2

# Links

Tracks https://github.com/SUSE/spacewalk/issues/18916


keichwa (approved):

vzhestkov (approved):
Looks good.
@ycedres thanks for taking care of it

PR #1882, opened 2022-11-09
Nav bar tidying - installation section

0rnela:
# Description

Nav bar was out of sorts. This change tidied things up. (Part 1 - installation segment only)

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [ ] Manager-4.2

# Links

Fixes https://github.com/SUSE/spacewalk/issues/19442
4don


keichwa (approved):
I think it would be good if we could avoid deep sectioning.  Approving nevertheless.

PR #1883, opened 2022-11-10
Nav bar tidying - setting up section

0rnela:
# Description

Nav bar was out of sorts. This change tidied things up. (Part 2 - setup segment only)

# Target branches

- [x] Master (Default)
- [x] Manager-4.3 https://github.com/uyuni-project/uyuni-docs/pull/1891
- [ ] Manager-4.2

# Links

Fixes https://github.com/SUSE/spacewalk/issues/19442
4don2

### DIFF
--- a/modules/installation-and-upgrade/nav-installation-and-upgrade-guide.adoc
+++ b/modules/installation-and-upgrade/nav-installation-and-upgrade-guide.adoc
@@ -50,30 +50,34 @@ ifeval::[{suma-content} == true]
 endif::[]
 
 ** xref:setup-intro.adoc[Setup]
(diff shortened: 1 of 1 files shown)


0rnela:
@keichwa This request came from Don and upon closer look, I agree with him that our nav bar was very disorganized. I'd prefer to go ahead with the solution I'm proposing now as it is an improvement (except for the deep sectioning, but this is the quickest and simplest win right now). Any book reorganization to separate Server and Proxy sounds better, but obviously not possible right now due to its scope.

jcayouette (approved):

PR #1939, opened 2022-12-05
Capping single-server support to 10 000 minions

0rnela:
# Description

From the original issue: "Some customers are still using suse manager deployment with just one server to manage a huge amount of minions (18k)."

This PR explicitly states that the maximum number of clients supported per single server is 10 000.

Content added to:
1. HW Requirements page in the form of warning and link to Large Deployments.
2. Existing Large deployments file changed to give more explicit guidelines (i.e. "you might find it easier" was replaced by "you need to use multiple {productname} Servers in a hub").

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [x] Manager-4.2

# Links

Fixes https://github.com/SUSE/spacewalk/issues/19491

### DIFF
--- a/modules/installation-and-upgrade/pages/hardware-requirements.adoc
+++ b/modules/installation-and-upgrade/pages/hardware-requirements.adoc
@@ -127,11 +127,12 @@ Double the space if the server is an ISS Master.
 endif::[]
 
 ifeval::[{suma-content} == true]
-[WARNING]
+[IMPORTANT]
 ====
-Maximum number of clients which can be managed by a single {productname} server is 10 000.
-Registering higher than maximum number is not supported by {suse}.
-Customers can ask support for advice or exception depending on their own workload.
+{productname} performance depends on a number of underlying resources, for example network bandwidth and latency between clients and servers.
+
+Based on the experience and different deployments that are in use, the advice for optimal performance of {productname} Server  with an adequate number of proxies is to not exceed 10 000 clients per single server.
+It is strongly recommended to move to Hub setup and involve consultancy once you have more than 10 000 clients, as this can lead to performance issues.
 
 For more about managing large number of clients, see xref:specialized-guides:large-deployments/multi-server.adoc[].
 ====
--- a/modules/specialized-guides/pages/large-deployments/multi-server.adoc
+++ b/modules/specialized-guides/pages/large-deployments/multi-server.adoc
@@ -1,16 +1,11 @@
 [[lsd-multi-server]]
 = Using Multiple Servers to Manage Large Scale Deployments
 
-A single {productname} Server can manage up to 10 000 clients when tuned appropriately.
+If you need to manage a large number of clients, in most cases you can do so with a single {productname} Server, tuned appropriately.
+However, if you need to manage tens of thousands of clients, you might find it easier to use multiple {productname} Servers, in a hub, to manage them.
 
-[IMPORTANT]
-====
-Registering higher than maximum number on a single {productname} Server is not supported by {suse}.
-====
+{productname} Hub helps you manage very large deployments.
 
-To manage large deployments with more than 10 000 clients, you need to use multiple {productname} Servers in a hub.
-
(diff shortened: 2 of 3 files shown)


rjmateus (approved):
Looks good to me

modules/installation-and-upgrade/pages/hardware-requirements.adoc
```diff
@@ -124,10 +124,16 @@ Double the space if the server is an ISS Master.
...
 
-
-
 endif::[]
 
+ifeval::[{suma-content} == true]
+[WARNING]
+====
+Maximum number of clients which can be managed by a single {productname} server is 10 000.
+Reistering higher than maximum number is not supported by {suse}.
```
aaannz, line 133:
```suggestion
Registering higher than maximum number is not supported by {suse}.
```

modules/specialized-guides/pages/large-deployments/multi-server.adoc
```diff
@@ -1,10 +1,16 @@
...
 = Using Multiple Servers to Manage Large Scale Deployments
 
-If you need to manage a large number of clients, in most cases you can do so with a single {productname} Server, tuned appropriately.
-However, if you need to manage tens of thousands of clients, you might find it easier to use multiple {productname} Servers, in a hub, to manage them.
+A single {productname} Server can manage up to 10 000 clients when tuned appropriately.
 
-{productname} Hub helps you manage very large deployments.
+[IMPORTANT]
+====
+Reistering higher than maximum number on a single {productname} Server is not supported by {suse}.
```
aaannz, line 8:
```suggestion
Registering higher than maximum number on a single {productname} Server is not supported by {suse}.
```

aaannz (commented):
Two typos and hope @admd is aware of this as it is in quite strong language. Appears like hard limit, which is ok I just imagine some customers may start panicking after reading it.

rjmateus:
That is a good point @aaannz 
Maybe something like "{suse} may not support systems with a registration number a higher than maximum number on a single {productname} Server."

What do you guys think?

0rnela:
Sounds good @rjmateus and @aaannz, I'll amend.

My understanding was that we indeed need to prevent the customers from overstretching the servers, and that we should express it more directly.

I'll make the new changes.

0rnela:
@rjmateus @aaannz -- after discussing with @keichwa we take the approach that we should avoid "may, may be" phrasing as it does not give customers clear guidelines to follow and leaves things open to interpretation.

The intention of this PR / original issue was to cap the number of clients per single server, and these changes address that.

However, to avoid the risk of panic of any kind, the following sentence was added:
"Customers can ask support for advice or exception depending on their own workload."

I hope this resolves the issues.

Please let me know otherwise.

0rnela:
> 👍🏾

Thank you @aaannz .

modules/installation-and-upgrade/pages/hardware-requirements.adoc
```diff
@@ -124,10 +124,18 @@ Double the space if the server is an ISS Master.
...
 |===
 
-
-
 endif::[]
 
+ifeval::[{suma-content} == true]
+[WARNING]
+====
+Maximum number of clients which can be managed by a single {productname} server is 10 000.
```
admd, line 132: I am not sure about the wording here as was pointed by other colleagues. We should be going for like "It is advisable or strongly recommended to move to Hub setup ad involve consultancy once you have more than 10k number of minions" , not exactly what I mentioned but along those lines.

admd (changes requested):
I agree with the reservations mentioned above. There are quite a few customers out there who are running more than 10k with a single SUMA and number of proxies. Adding such paragraph would definitely create panic as it suggests they are not supported anymore. 

We need better wording and going more along the advise rather black & white supported/not supported.

0rnela:
A lot of additional files got pulled in during conflict resolutions. Creating new, clean PR.

Issue #1968, opened 2022-12-20
Wrong Scheduler Recommended in Large Deployment 


pmullin99:
## Documentation Update Required
The wrong scheduler recommended in large deployment guide for SUSE Manager 4.3. 

## What does the documentation say now?
If the SUSE Manager Server is virtualized, enable the elevator=noop kernel command line option, for the best input/output performance. You can check the current status with cat /sys/block/<DEVICE>/queue/scheduler. This command will display a list of available schedulers with the currently active one in brackets. To change the scheduler before a reboot, use echo noop > /sys/block/<DEVICE>/queue/scheduler.

## What change is required?
In 15sp4, noop is no longer an option.  The only valid options are:
cat /sys/block/sda/queue/scheduler
mq-deadline kyber [bfq] none

What content needs to be added?
change "noop" to "none"

## Version
Since SLES 15SP2 there was a transition from non-multiqueue schedulers (noop, cfq, deadline) to multiqueue schedulers (none, mq-deadline, bfq, kyber).
This should be backported to 4.1 to 4.2.

0rnela:
@pmullin99 - is this the change you had in mind (**_noop_** to **_none_** in thise two lines):

If the SUSE Manager Server is virtualized, enable the elevator=**none** kernel command line option, for the best input/output performance. You can check the current status with cat /sys/block/<DEVICE>/queue/scheduler. This command will display a list of available schedulers with the currently active one in brackets. To change the scheduler before a reboot, use echo **none** > /sys/block/<DEVICE>/queue/scheduler.

pmullin99:
You may want to indicate device in path, ex.  cat /sys/block/<block device>/queue/scheduler.  Yes, every place noop appears.

0rnela:
Fixed. Closing the issue.

PR #2000, opened 2023-01-12
Changing noop to none

0rnela:
# Description

According to the issue raised on Uyuni board https://github.com/uyuni-project/uyuni-docs/issues/1968: In 15sp4, noop is no longer an option.

WIP as I await response about the extent of the change.

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [ ] Manager-4.2

# Links

Fixes https://github.com/SUSE/spacewalk/issues/19975


keichwa (approved):

0rnela:
Adding @mcalmer for approval following the conversation with @cbbayburt.

Michale, could you please provide feedback or point me to the right person? Thank you in advance.

mcalmer (approved):
It seems that all these values are targeting the SUSE Manager Server / Uyuni Server itself.
So we do not need to take SLES12 into consideration.
I assume this is the case, then we can do the change

PR #2008, opened 2023-01-18
Fix SLE Micro for SUSE Manager and Uyuni

juliogonzalez:
# Description

SLE Micro is a supported client for Uyuni as well, and the OS channels come from SCC, so the if is removed.

SLE Micro 5.1 is supported, for some reason, whoever added it added only the client tools for Uyuni, but didn't mention the SCC channels.

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [ ] Manager-4.2

# Links

Partially handles https://github.com/SUSE/spacewalk/issues/20221

### DIFF
--- a/modules/client-configuration/pages/clients-slemicro.adoc
+++ b/modules/client-configuration/pages/clients-slemicro.adoc
@@ -35,6 +35,7 @@ On the client, before registering it, run as [literal]``root``:
 ----
 transactional-update pkg install salt-transactional-update
 ----
+And then reboot the client.
 ====
 
 
@@ -45,7 +46,6 @@ Before you register {micro} clients to your {productname} Server, you need to ad
 
 include::snippets/arch-other-note.adoc[]
 
-ifeval::[{suma-content} == true]
 The products you need for this procedure are:
 
 [[slemicro-channels-wizard]]
@@ -55,6 +55,15 @@ The products you need for this procedure are:
 
 | OS Version
 | Product Name
+
+| {micro} 5.1 {x86_64}
+| SUSE Linux Enterprise Micro 5.1 x86_64
+
+| {micro} 5.1 {arm64}
+| SUSE Linux Enterprise Micro 5.1 aarch64
+
+| {micro} 5.1 {s390x}
+| SUSE Linux Enterprise Micro 5.1 s390x
 
 | {micro} 5.2 {x86_64}
 | SUSE Linux Enterprise Micro 5.2 x86_64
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -45,7 +46,6 @@ Before you register {micro} clients to your {productname} Server, you need to ad
 
 include::snippets/arch-other-note.adoc[]
 
-ifeval::[{suma-content} == true]
```
juliogonzalez, line 48: As mentioned, SLE Micro comes from SCC for Uyuni as well (but not the client tools). A subscription is required.

modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -136,14 +153,14 @@ To add the client tools, add these channels from the command prompt:
 | OS Version
 | Client Channel
 
-| {micro} 5.3
-| sle-micro-5.3-uyuni-client
+| {micro} 5.1
```
juliogonzalez, line 156: I changed the order to match what we show above from SCC. 5.1 first, then 5,2 and finally 5.3

modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -136,14 +153,14 @@ To add the client tools, add these channels from the command prompt:
...
+| suse-microos-5.1-uyuni-client
```
juliogonzalez, line 157: Yes, `suse-microos` is right, that's the name for the SCC channels as well, for both 5.1 and 5.2

modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -183,4 +200,12 @@ This window may be several hours after the client is registered.
 To speed up {micro} registration, manually reboot the client after the registration script finishes.
 ====
 
+To register your {micro} clients, you need a bootstrap repository.
```
juliogonzalez, line 203: I copied this from the other OS.

To be honest, I think you should transform the next things into a snippet.

keichwa: I created an issue.

modules/client-configuration/pages/clients-slemicro.adoc
keichwa: ok

keichwa (approved):
In general, we wanted to settle on latest (= most recent) version first. But ATM, docs are mixed.  Maybe, in the end it is better to have lowest version number always first.  I've no strong opinion in this case.

juliogonzalez:
> In general, we wanted to settle on latest (= most recent) version first. But ATM, docs are mixed. Maybe, in the end it is better to have lowest version number always first. I've no strong opinion in this case.

Me neither, but we should have a rule for that :-D

Or at the very least, we should have the same order in the same page (that's what I did with this PR).

PR #2034, opened 2023-02-02
Reviewing and reducing admonitions, live patching SLES 12 file(s)

0rnela:
# Description

Reviewing and reducing admonitions wherever possible, as we have cases when they are either overused (that way losing their effect and meaning) or unnecessary, i.e. can be replaced with regular text.

Here, two sequential admonitions "Important" were condensed to just one.

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [ ] Manager-4.2

# Links


keichwa (approved):
see #2036.

jcayouette (approved):

PR #2121, opened 2023-03-13
Update space-management.adoc

ptinley-vt:
# Description

Adding a blank line on line 59. It seems that the fact that there is no blank line between 'For example:' and '----', causes the formatting of the zh_CN Administration Guide PDF to become corrupted from section 9.3 onwards. Other Asian languages (JA and KO), and of course EN, were not effected by this missing blank line.

# Target branches

- [x] Manager-4.3

# Links

Administration Guide - space-management - zh_CN_issue1b

### DIFF
--- a/modules/administration/pages/space-management.adoc
+++ b/modules/administration/pages/space-management.adoc
@@ -56,6 +56,7 @@ A value of ``true`` enables the shut down feature.
 Any other value disables it.
 
 For example:
+
 ----
 spacecheck_shutdown = true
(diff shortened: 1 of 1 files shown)


0rnela (approved):

jcayouette:
Nice work Patrick! We should take care to backport these changes to master, manager-4.3 and manager-4.2.
Otherwise this work will not show up in the next release as we create our releases from the release branches.

PR #2140, opened 2023-03-20
Broken table. Update supported-features-es.adoc

ptinley-vt:
# Description

Adding an extra cell on the row beginning with 'Monitoring server'. Line 203.
The issue is that one table cell is missing (only 3 entries, when it is a 4 column table), causing the remaining table entries to flow back and into the wrong position.
I have assumed the missing entry should be a {cross}. Should be verified with the authors.

| Monitoring server
| {cross}
| {cross}

# Target branches

- [ ] Master (Default)
- [ ] Manager-4.3

# Links

### DIFF
--- a/modules/client-configuration/pages/supported-features-es.adoc
+++ b/modules/client-configuration/pages/supported-features-es.adoc
@@ -200,6 +200,7 @@ The icons in this table indicate:
 | Monitoring server
 | {cross}
 | {cross}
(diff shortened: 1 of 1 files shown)


keichwa (approved):
Thanks for spotting, Patrick!

jcayouette:
Thank you Patrick!

PR #2158, opened 2023-04-04
Adapt to Keycloak 21.0.1

Bischoff:
# Description

Update the SSO example  (previously: Keycloak 9.0.2; now: Keycloak 21.0.1).

Reorganize to match common sense (you cannot extract data from keycloak before installing it...).

# Target branches

- [x] Master (Default)
- [x] Manager-4.3
- [x] Manager-4.2

# Links

Fixes https://github.com/SUSE/spacewalk/issues/18861


keichwa (approved):
Thanks for updating the documentation, Eric!  If I got it right you re-used existing texts where ever possible (which is a good thing), just moved it where needed. Then you fixed obvious errors and documented changes.  This is perfect.
It would be nice if you'd add a short `.changelog` entry (toplevel directory).

PR #2159, opened 2023-04-04
Update server-y-yast.adoc

ptinley-vt:
# Description

Request to add two spaces at the start of line 51, before "If you are upgrading from 4.1 to 4.2, run the database migration script:" Without these spaces, list item 6 merges back into list item 5 in the translated books (CJK at least). This string also exists in server-y-zypper.adoc.
In that file the two spaces (an indent) are present and that is the format used in the PO file.
image

Without these space the translated files look like this:
image

# Target branches

- [ ] Master (Default)
- [ ] Manager-4.2
- [ ] Manager-4.2.12

# Links


jcayouette (approved):

jcayouette:
Thanks Patrick. I should probably ping upstream.

PR #2165, opened 2023-04-05
Update client-upgrades-product-migration.adoc

ptinley-vt:
# Description

Request to comment-out, or remove, 3 plus signs, that are erroneously appearing in the output text. These appear inside an IMPORTANT block. There are no lists in this block, so I believe there is no need for the list continuation + signs. They do not operate on the formatting and instead display in the body text.

Section 8.3.2.2, page 169.

image

# Target branches

- [ ] Master (Default)
- [ ] Manager-4.3
- [ ] Manager-4.2

# Links


0rnela:
Thank you.

This file has been fixed and reorganized as part of the ongoing work. New PR is pending review. 

Closing.

PR #2221, opened 2023-05-05
Note to clarify difference between Uyuni and SUMA

dvosburg:
Uyuni - 'zypper up'
SUMA - 'zypper patch'

# Description
Clarifies between Uyuni update and SUMA patch

# Target branches

- [X] Master (Default)
- [ ] Manager-4.3
- [ ] Manager-4.2

# Links

None

### DIFF
--- a/modules/installation-and-upgrade/pages/server-minor-upgrade-uyuni.adoc
+++ b/modules/installation-and-upgrade/pages/server-minor-upgrade-uyuni.adoc
@@ -42,13 +42,13 @@ zypper ref
 +
 
 ----
-zypper up 
+zypper up
 ----
 [NOTE]
 ====
-Uyuni is different from SUSE Manager in this step.  SUSE Manager uses 'zypper patch', but Uyuni requires 'zypper up'
+Uyuni is different from SUSE Manager in this step.
+SUSE Manager uses [command]``zypper patch``, but Uyuni requires [command]``zypper up``.
 ====
-
 +
 
 . If zypper reports that the Uyuni package won't be upgraded, run the command manually:


juliogonzalez (approved):
LGTM, but I am adding more reviewers from the doc squad, so they can tell :-)

modules/installation-and-upgrade/pages/server-minor-upgrade-uyuni.adoc
```diff
@@ -44,6 +44,11 @@ zypper ref
 ----
 zypper up 
 ----
+[NOTE]
+====
+Uyuni is different from SUSE Manager in this step.  SUSE Manager uses 'zypper patch', but Uyuni requires 'zypper up'
+====
+
 +
```
keichwa, line 52: This is WIP.  Proper ifdef will follow.  ATM, github does not let me a proper patch...

```suggestion
----
zypper patch
----
----
zypper up
----
[NOTE]
====
Uyuni is different from SUSE Manager in this step.
SUSE Manager uses 'zypper patch', but Uyuni requires 'zypper up'.
====

+
```

keichwa (approved):
This is just a proposal.  I think it would be enough if we'd add the note to uyuni documentation only.  But feel free to refuse or adjust!  Thanks, Don!

modules/installation-and-upgrade/pages/server-minor-upgrade-uyuni.adoc
juliogonzalez: I don't think ifdef is needed here. Please check the filename. Seems to be 100% specific to Uyuni.

keichwa: Ok, the less ifdefs the better!  I adjusted my suggestion to a minimum.

PR #2258, opened 2023-05-24
Uyuni cosmetic wording change #16856 keichwa

keichwa:
# Description

~https://github.com/SUSE/spacewalk/issues/16856!~ no longer applies.

# Target branches

- [x] Master (Default)
- [ ] Manager-4.3
- [ ] Manager-4.2

# Links


keichwa:
If you think that's confusing, I can revert the renaming of this PR and just add an xref to the "Major Uyuni Upgrade".

I'd like to avoid the "major/minor" wording for Uyuni, because we use the same wording "major/minor" for different grade in the case of SUMA.

deneb-alpha:
> If you think that's confusing, I can revert the renaming of this PR and just add an xref to the "Major Uyuni Upgrade".
> 
> I'd like to avoid the "major/minor" wording for Uyuni, because we use the same wording "major/minor" for different grade in the case of SUMA.

considering that  Uyuni follow a rolling release model, what about using `upgrade` vs `patch`?

From what I saw until now, we have only the upgrade to a new stable version and a patch that is installed on top of an existing released version.

@juliogonzalez am I missing something? 
@keichwa / @0rnela: what do you think?

juliogonzalez (commented):
What's exactly the problem with calling a minor upgrade, minor upgrade?

For people reading quickly, I am sure they will confuse "upgrade" and "major upgrade". Basically they will ignore the second one.

deneb-alpha:
> Uyuni is Uyuni, SUSE Manager is SUSE Manager. Did we see anyone confused by this?
> 
> > considering that Uyuni follows a rolling release model, what about using `upgrade` vs `patch`?
> 
> No, please no! We'll then have people using `zypper patch`.

I didn't considered this, but I agree there's this risk, so, I take back my idea.

> 
> And besides, Uyuni has patches, and it's a completely different thing: https://www.uyuni-project.org/pages/patches.html
> 
> > From what I saw until now, we have only the upgrade to a new stable version and a patch that is installed on top of an existing released version.
> 
> It's always an upgrade.
> 
> The only difference is if if the base OS needs to change and extra steps are needed (major) or not (minor)
> 

I think I got lost into jargon. my bad. 
I was thinking at the Uyuni naming YYYY.MM and I was considering as major a new release and  as a patch a security fix released on top of the existing stable. but indeed, you're right, major involves a base system update, while a minor only the update for the fixes  and changes we released. 

with that, I take back the idea to change the naming, unless @keichwa is aware of issues and misunderstanding reported due  to the  current way to call the different updates/upgrades.

PR #2263, opened 2023-05-25
add snippet for create_bootstrap_repo_register

keichwa:
# Description

https://github.com/SUSE/spacewalk/issues/21591

# Target branches

- Master (Default)

# Links

### DIFF
--- a/modules/client-configuration/pages/clients-pubcloud.adoc
+++ b/modules/client-configuration/pages/clients-pubcloud.adoc
@@ -97,9 +97,6 @@ When the bootstrap process has completed successfully, your client is listed at 
 * When registering clients launched from a public cloud image (for example, AWS AMI), you need to do some additional configuration to prevent them from over-writing each other.
   For more information about registering clones, see xref:administration:troubleshooting/tshoot-registerclones.adoc[].
 
-// 2023-05-31, ke: using snippet here?
-// see include::snippets/create_bootstrap_repo_register.adoc[]
-
 
 
 == Activation keys


modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -188,15 +188,7 @@ This window may be several hours after the client is registered.
...
-To register your {micro} clients, you need a bootstrap repository.
-By default, bootstrap repositories are automatically created, and regenerated daily for all synchronized products.
-You can manually create the bootstrap repository from the command prompt, using this command:
-
-----
-mgr-create-bootstrap-repo
-----
-
-For more information on registering your clients, see xref:client-configuration:registration-overview.adoc[].
+include::snippets/create_bootstrap_repo_register.adoc[]
```
juliogonzalez, line 191: I don't think this is the only place where we have this text? Probably we want to use the snippet in more places. Use grep. I found a lot of other adoc files.

juliogonzalez (changes requested):
LGTM, but the snippet should be used everywhere.

keichwa:
> LGTM, but the snipper should be used everywhere.

Yes, but I want to create a separate PR for all the other files.  But if you think it is better to have all in one PR I can do that.

modules/client-configuration/pages/clients-slemicro.adoc
keichwa: ok, thanks, it is ready for review now.

keichwa:
I've linked the output format from https://github.com/SUSE/spacewalk/issues/21591 (description).

modules/client-configuration/pages/clients-alibaba.adoc
```diff
@@ -114,16 +114,16 @@ For more information on activation keys, see xref:client-configuration:activatio
...
 == Register Clients
 
-{alibabaclo} clients are registered in the same way as all other clients.
-
+[[WARNING]]
+====
 Some {alibabaclo}{nbsp}2 instances will fail to register on the first try.
-
 This is due to a to a known bug in the {alibabaclo}{nbsp}2 image.
+The `python-urlgrabber3` package is provided both as a Python pip package and an RPM package, which can cause a conflict on the first attempt to register.
```
juliogonzalez, line 121: While I agree with this change, this is not related to this PR.

Please create a separate PR if you want to add this warning.

modules/client-configuration/pages/clients-pubcloud.adoc
```diff
@@ -97,6 +97,9 @@ When the bootstrap process has completed successfully, your client is listed at
 * When registering clients launched from a public cloud image (for example, AWS AMI), you need to do some additional configuration to prevent them from over-writing each other.
   For more information about registering clones, see xref:administration:troubleshooting/tshoot-registerclones.adoc[].
 
+// 2023-05-31, ke: using snippet here?
```
juliogonzalez, line 100: Do not leave a comment. This needs to be  decided. Either we include it, either we do not :-D

juliogonzalez (changes requested):

modules/client-configuration/pages/clients-alibaba.adoc
keichwa: ok, I reverted it.

Background info: The content was already in, I just added the `WARNING` markup.  But it is better to do this separately.

modules/client-configuration/pages/clients-pubcloud.adoc
keichwa: ok, I do not include it.

modules/client-configuration/pages/clients-oracle.adoc
```diff
@@ -195,12 +195,10 @@ ol67-gpg-pubkey-72F97B74EC551F0A3.key
 
 == Register Clients
 
-{oracle} clients are registered in the same way as all other clients.
-For more information, see xref:client-configuration:registration-overview.adoc[].
-
-
 //[WARNING]
```
juliogonzalez, line 198: Not related to this PR, but... time to remove this in another PR?

modules/client-configuration/pages/clients-rh-cdn.adoc
```diff
@@ -472,19 +472,10 @@ For the complete list of the {redhat} product signing keys, see https://access.r
...
-You can manually create the bootstrap repository from the command prompt, using this command:
-
-----
-mgr-create-bootstrap-repo
-----
-
-For more information on registering your clients, see xref:client-configuration:registration-overview.adoc[].
-
-
 //[WARNING]
```
juliogonzalez, line 475: Same as before

juliogonzalez (approved):
LGTM, but please have more people in docs reviewing, just in case.

When you merge, remember use the squash options, so you can cherrypick the comment back to 4.3, as I guess you want this fix there as well.

keichwa:
> LGTM, but please have more people in docs reviewing, just in case.

Yes, they are already listed as reviewers.  I guess they check PR requests daily, or almost daily.

> When you merge, remember use the squash options, so you can cherrypick the comment back to 4.3, as I guess you want this fix there as well.

ok, but I'd only backport it if one thinks that it is really needed.  IMO, it just something internal, almost invisible to the reader or user.

juliogonzalez:
> > When you merge, remember use the squash options, so you can cherrypick the comment back to 4.3, as I guess you want this fix there as well.
> 
> ok, but I'd only backport it if one thinks that it is really needed. IMO, it just something internal, almost invisible to the reader or user.

True, but it also makes maintenance easier if you want to add new OS. For example at some point there will be Debian 12, and my guess (pending PO's approval) is that 4.3 will get it.

keichwa:

> True, but it also makes maintenance easier if you want to add new
> OS. For example at some point there will be Debian 12, and my guess
> (pending PO's approval) is that 4.3 will get it.

Yes.  Or we'd then copy the new file with Debian 12 plus the snippet file
with the register info from master to 4.3, and we are done.

IMO, backporting this PR in advance would make sense, if we think we
probably will or must change the content of the snippet file with
register info.

But as said if someone wants the backport now, I'll happily do it.

-- 
Karl Eichwalder                       SUSE Software Solutions Germany GmbH
                                      Frankenstraße 146, D-90461 Nürnberg
Geschäftsführer: Ivo Totev, Andrew Myers, Andrew McDonald, Boudien Moerman
(HRB 36809, AG Nürnberg)

jcayouette (approved):

PR #2326, opened 2023-06-28
backport43 - File name change to reflect standardized naming format

0rnela:
# Description

Short summary of why you created this PR (if you added documentation, please add any relevant diagram)

# Target branches

- [x] Master (Default) https://github.com/uyuni-project/uyuni-docs/pull/2325
- [x] Manager-4.3
- [ ] Manager-4.2

# Links

https://github.com/SUSE/spacewalk/issues/21502


keichwa (commented):
We must make sure to have filename redirects on the web server(s).  Otherwise such a change will break user saved bookmarks.

0rnela:
> We must make sure to have filename redirects on the web server(s). Otherwise such a change will break user saved bookmarks.
> 
> Or keep the file and make it an xref to the new (backported) file.

@keichwa -- I took this into consideration. As you know, the building tool complains the "unused" files, which we could keep for some time, and then remove. But I believe it is not a high risk to simply remove this particular old file as it is not one of the critical processes that the users are more likely to bookmark, hence my decision to delete it, rather than to have an extra task of removing the file some months further down the line.

jcayouette (approved):
at first glance this seems ok. Thank you.

PR #2375, opened 2023-07-20
Add saltboot redeployment sub chapter

aaannz:
# Description

Added saltboot redeployment guide as a terminal deployment subchapter.

# Target branches

Which documentation version does this PR apply to?

- [ ] Master (Default)
- [X] Manager-4.3
- [ ] Manager-4.2

# Links

Fixes https://github.com/SUSE/spacewalk/issues/21884

PR #2379, opened 2023-07-24
4.3 - Additional changes for forced Saltboot redeployment

0rnela:
# Description

Fine-tuning PR https://github.com/uyuni-project/uyuni-docs/pull/2375 (Add saltboot redeployment sub chapter) further

# Target branches

- [ ] Master (Default)
- [x] Manager-4.3
- [ ] Manager-4.2

# Links


aaannz (approved):
Thanks! LGTM

By the way, when you want to rename files in git, it is better to use `git mv <old file> <new file>`, that way git tracks just the filename and path as a change and not deletion and adding of full file.

PR #2466, opened 2023-09-20
Fixed the support matrix in two Guides

0rnela:
# Description

The support matrix in the Install and Upgrade Guide and Client Config Guide were different in some entries, and were not identical either.

# Target branches

- [x] master (Default)
- [x] manager-4.3 - https://github.com/uyuni-project/uyuni-docs/pull/2495

# Links
- Fixes https://github.com/SUSE/spacewalk/issues/22551

### DIFF
--- a/modules/client-configuration/pages/supported-features.adoc
+++ b/modules/client-configuration/pages/supported-features.adoc
@@ -33,6 +33,10 @@ The operating system running on a client is supported by the organization that s
 // include::../../snippets/pages/supported-client-systems-snippet.adoc[]
 
 // Revert PR 1600 
+
+// 2023-09-20 OM: Make sure this table and the one in file
+// modules/installation-and-upgrade/pages/client-requirements.adoc are identical
+
 [[supported-client-systems-snippet]]
 
 ifeval::[{suma-content} == true]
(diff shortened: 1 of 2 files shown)


keichwa:
I think it would be good either to move the contents into a snippet file or simply delete the table in one of these files.

0rnela:
> I think it would be good either to move the contents into a snippet file or simply delete the table in one of these files.

@keichwa -- indeed, and there is in fact a snippet in a separate file already. I created it a while back so we could maintain the support matrix in one place ONLY. However, we can't seem to be able to call a snippet that is at the same level as modules in the file hierarchy. I have already added this to the list of issues to look at HackWeek, to see if anything had changed.

keichwa:
Ornela Marić ***@***.***> writes:

> @keichwa -- indeed, and there is in fact a snippet in a separate file
> already. I created it a while back so we could maintain the support
> matrix in one place ONLY. However, we can't seem to be able to call a
> snippet that is at the same level as modules in the file hierarchy. I
> have already added this to the list of issues to look at HackWeek, to
> see if anything had changed.

Thanks for the background info!

-- 
Karl Eichwalder                       SUSE Software Solutions Germany GmbH
                                      Frankenstraße 146, D-90461 Nürnberg
Geschäftsführer: Ivo Totev, Andrew Myers, Andrew McDonald, Boudien Moerman
(HRB 36809, AG Nürnberg)

PR #2477, opened 2023-09-26
Update clients-rh-rhui.adoc

ptinley-vt:
# Description

A missing '=' character on line 70 is messing up the formatting/layout of file and the produced PDF (also localised CJK books). There needs to be 4 = symbols in a sequence to close the note. There are only 3.

I've added the 4th and it fixes the layout issue.

# Target branches

- [ ] master (Default)
- [ ] manager-4.3.8

The file displays as below if the 4th '=' symbol is not in place.
image


keichwa (approved):
It is fixed in master, but still needs to be ported to the manager-4.3 branch.

PR #2509, opened 2023-10-11
Retail monitoring option

keichwa:
# Description

https://github.com/SUSE/spacewalk/issues/17497

# Target branches

Does this content need to be backported? If this is the case, create an issue for tracking it and add the link to this PR.

# Links
- Fixes #<insert spacewalk issue, if any>

### DIFF
--- a/modules/retail/pages/retail-components.adoc
+++ b/modules/retail/pages/retail-components.adoc
@@ -30,7 +30,8 @@ The branch server can also manage local networking, and provide DHCP and DNS ser
 
 [NOTE]
 ====
-For monitoring, you can install Prometheus server on the Branch servers, and, for visualization and analysis, Grafana with multiple data sources on a dedicated host.
+* For monitoring, you can install Prometheus server on the Branch servers.
+* For visualization and analysis, you can install Grafana with multiple data sources on a dedicated host.
 For more information about Prometheus and Grafana, see xref:administration:monitoring.adoc[].
 ====
 


modules/retail/pages/retail-components.adoc
```diff
@@ -20,13 +20,19 @@ They are used to securely build operating system images.
...
 == Branch Servers
 
 Branch servers should be physically located close to point-of-service terminals, such as in an individual store or branch office.
 Branch servers provide services for PXE boot, and act as an image cache, Salt broker, and proxy for software components (RPM packages).
 The branch server can also manage local networking, and provide DHCP and DNS services.
 
-
+[NOTE]
+====
+For monitoring, you can install Prometheus server on the Branch servers, and, for visualization and analysis, Grafana with multiple data sources on a dedicated host.
```
aaannz, line 33: I had to read the line multiple times to understand it. 😃 

Can it be split into two sentences, one for Prometheus and another for Grafana?

0rnela (approved):

modules/retail/pages/retail-components.adoc
0rnela, line 33:
```suggestion
* For monitoring, you can install Prometheus server on the Branch servers.
* For visualization and analysis, you can install Grafana with multiple data sources on a dedicated host.
```

0rnela (changes requested):

witekest:
The original issue was about Grafana not being available on the branch server. To that extend the change is good I think.

> Maybe @witekest can provide more details how to set it up and integrate with terminals.

The problem is we don't have a good procedure for discovering and configuring proxy clients. I have just created a [feature request](https://github.com/SUSE/spacewalk/issues/22756) for extending the current implementation.
Instead of documenting a workaround I'd prefer to wait until the new feature is implemented and then document it with all needed details.

modules/retail/pages/retail-components.adoc
keichwa: Yes, I completely agree.  Thanks, Ornela, for the suggestion!

keichwa:
Thanks for the background info, Witek!

PR #2527, opened 2023-10-18
Port sp-base macro from 4.3 branch #22764

keichwa:
# Description

https://github.com/SUSE/spacewalk/issues/22764.

# Target branches

- master

# Links
- Fixes #<insert spacewalk issue, if any>

### DIFF
--- a/modules/installation-and-upgrade/pages/install-vm.adoc
+++ b/modules/installation-and-upgrade/pages/install-vm.adoc
@@ -21,7 +21,7 @@ These are suitable for a quick test installation, such as a server with one clie
 If you want to use a production environment, review the requirements listed in xref:hardware-requirements.adoc[].
 ====
 
-In the following table replace {sles-version}-{sp-base} with the actual product version string.
+In the following table replace {sles-version} {sp-base} with the actual product version string.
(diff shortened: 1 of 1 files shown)


modules/installation-and-upgrade/pages/install-vm.adoc
```diff
@@ -21,7 +21,7 @@ These are suitable for a quick test installation, such as a server with one clie
 If you want to use a production environment, review the requirements listed in xref:hardware-requirements.adoc[].
 ====
 
-In the following table replace {sles-version}-{sp-version} with the actual product version string.
+In the following table replace {sles-version}-{sp-base} with the actual product version string.
```
0rnela, line 24:
```suggestion
In the following table replace {sles-version} {sp-base} with the actual product version string.
```

0rnela (approved):

modules/installation-and-upgrade/pages/install-vm.adoc
keichwa: I think the best would be if we'd tag this with `[literal]` here.  But that would be something for think about later. Applying your suggesting now

PR #2535, opened 2023-10-19
Corrected client tools channel name

0rnela:
# Description

Client tools channel name was incorrect.

# Target branches

- [x] master
- [ ] manager-4.3

# Links
- Fixes https://github.com/SUSE/spacewalk/issues/22803

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -4,7 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)
 
-- Corrected the client tools channel name in Client Configuration Guide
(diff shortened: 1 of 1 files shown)


.changelog
```diff
@@ -4,6 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)
 
+- Corrected the client tools channel name in Client Configuration Guide
```
nodeg, line 7:
```suggestion
- Corrected the client tools channel name in Client Configuration Guide for SLE Micro
```
I would add something like this, so we know it is only for SLE Micro.

nodeg (approved):
Small suggestion, other than that LGTM. Thank you, Ornela!

PR #2536, opened 2023-10-19
Add information about Uyuni repository version pinning

### LINKED ISSUE #2530: Information about version pinning missing
When installing Uyuni Server or Proxy, the installation documentation points to using the "latest and greatest" repository. In some cases, users might not want to use the very latest version of Uyuni, but the version before.

Also, when having a very outdated version of Uyuni, updating to the most recent version is likely to fail.

## Documentation Update Required

It would be great to have a not about version pinning in the "Install Uyuni Server with openSUSE" section.

Let me know if you basically think that this is a good idea. **I can create a PR for you**.

## What does the documentation say now?

It simply suggests using the very latest repository packages.

## What change is required?

A note could be added.

What content needs to be added?

The documentation should mention additional repository URLs deoending on architecture and type (Server, Proxy):

- https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Snapshots:/YYYY.MM/images/repo/Uyuni-Server-POOL-x86_64-Media1/
- https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Snapshots:/YYYY.MM/images/repo/Uyuni-Server-POOL-aarch64-Media1/
- https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Snapshots:/YYYY.MM/images/repo/Uyuni-Proxy-POOL-x86_64-Media1/
- https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Snapshots:/YYYY.MM/images/repo/Uyuni-Proxy-POOL-aarch64-Media1/

## Subject Matter Expert (SME)

I don't know.

## Version

Which version of the software is this feature/update expected to land in (if known)? Next?
Which version of the documentation must this feature/update land in (if known)? Next?
Does the content need to be backported to previous documentation versions (if known)? Would be useful, so yes.

## Internal SUSE Reporters

This repository is public.
If you are reporting an issue that needs to remain company confidential, please use the internal Bugzilla instance instead.

stdevel:
# Description

When installing Uyuni Server or Proxy, the installation documentation points to using the "latest and greatest" repository. In some cases, users might not want to use the very latest version of Uyuni, but the version before.
Also, when having a very outdated version of Uyuni, updating to the most recent version is likely to fail.

# Target branches

- [x] master
- [ ] manager-4.3

# Links

- Fixes #2530

### DIFF
--- a/.changelog
+++ b/.changelog
@@ -4,6 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)
 
+- Added information about Uyuni repository version pinning
 - Corrected the Uyuni version in the migration instructions in Installation
   and Upgrade Guide
 - Fixed delete channel procedure in Channel management section in


keichwa:
Thanks for taking action!  I think in the description, #2520 has to be replaced with #2530 where you rightly added a ref to this PR.

stdevel:
You're so right, @keichwa - just fixed the wrong identifier.

juliogonzalez:
@keichwa not sure if you really want this at the doc, or better at the site.

Your call, but just in case it's not only that Snapshots are not supported, but they also need to be used with the right OS version (not documented per se anywhere, but can be figured out from the Release Notes), and they are not tested in any way, so it could be broken at any moment (should not, but you can never know).

Not to mention, it could be old Snapshots do not work with the client tools.

PR #2602, opened 2023-11-08
separate update-cfg-files from pot target

keichwa:
This way, we can call `make update-cfg-files` directly from the toplevel `make` directory.

- [ ] When approved change our documentation build howto (wiki) accordingly.


0rnela (approved):

jcayouette (approved):
Cool! Nice change. Thank you, Karl.

PR #2604, opened 2023-11-08
for gh-pages

keichwa:
uyuni 2023.10; `en`, `ja`, `ko`, and `zh_CN`.


jcayouette:
Why is this EN only? All translations should be included.

keichwa:
> Why is this EN only? All translations should be included.

Ok, will try this, too!

keichwa:
> > Why is this EN only? All translations should be included.
> 
> Ok, will try this, too!

Done, pls check again!

deneb-alpha (approved):
Thanks

PR #2613, opened 2023-11-14
diversify section IDs

keichwa:
It's not fatal because the IDs are not in use.  But it will stop the build process from complaining.

- [ ] check whether to backport to 4.3


0rnela (approved):

0rnela:
Thank you @keichwa. I welcome any changes that leave unnecessary messages out of the building process.

PR #2614, opened 2023-11-15
Add liberty linux 7 and 8 jcayouette

jcayouette:
*Before you submit this PR make sure you have added an entry to **.changelog** file.*

# Description

Condensed supported features table and added missing Liberty 7 and 8. There is no reason to keep multiple table rows for each client version. It is better to organize them into single client and versions per row. This solves PDF table truncation, makes things easier on the eyes, and is sustainable.

# Target branches
It needs to be backported to manager-4.3

- [x] master
- [x] manager-4.3


keichwa (approved):
LGTM, but it would be good if 0rnela could take a look, too.

.changelog
```diff
@@ -4,6 +4,7 @@
 # - Fixed error in Bat section of Upgrade Guide (bsc#1234567)
 # For guidelines: https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)
 
+- Added Liberty Linux versions 7 and 8 to the supported features matrix
```
```suggestion
- Added Liberty Linux versions 7 and 8 to the supported features matrix in
  the Client Configuration Guide
```

0rnela (changes requested):
The changes related to supported clients need to go to several files. Please check: https://confluence.suse.com/display/SUSEMANAGER/Client+Support+Matrix+changes

jcayouette:
Updated second table including both Uyuni and suma results.

0rnela (approved):

PR #2648, opened 2023-11-28
Update Red Hat OVAL data URL and file in CentOS Clients Registration guide

simonflood:
# Description

According to Steve Meier's News page, in July Red Hat changed the mechanism for distributing OVAL data. This PR updates the URL and filename change in the CentOS Clients Registration guide.

# Target branches

* Which documentation version does this PR apply to?
* Does this content need to be backported? If this is the case, create an issue for tracking it and add the link to this PR.
* Whenever possible, cross-reference each backport PR here, so that all backports can be easily accessed from the description.

- [X] master
- [X] manager-4.3 - https://github.com/uyuni-project/uyuni-docs/pull/2657

# Links

https://cefs.steve-meier.de/

PR #2665, opened 2023-11-30
Correct CentOS 7 Updates/Extras channel names in CentOS Client Config guide

simonflood:
# Description

After submitting #2648 I've realised that the CentOS 7 Updates & Extras channels are named differently from those given for the java.allow_adding_patches_via_api line in rhn.conf and the errata-import.pl command. Instead of centos7-updates-x86_64 and centos7-extras-x86_64 the architecture is in the middle so they're centos7-x86_64-updates and centos7-x86_64-extras respectively.

Fixing these allows the errata information to be imported for the Updates and Extras channels.

# Target branches

- [X] master
- [X] manager-4.3

# Links


keichwa:
thanks for the fix, I'll merge it now.

Issue #2733, opened 2024-01-26
SUMA 4.3 docs have the wrong module release version


slsnow:
## Documentation Update Required

Incorrect version release version number in instructions

## What does the documentation say now?

https://documentation.suse.com/suma/4.3/en/suse-manager/installation-and-upgrade/pubcloud-setup.html#_activate_the_public_cloud_module

```
SUSEConnect -p sle-module-public-cloud/15.5/x86_64
```

## What change is required?

Should say 
```
SUSEConnect -p sle-module-public-cloud/15.4/x86_64
```

Also of note, in the public cloud now you can use an SCC registration code, but still use the registercloudguest infrastructure (yes for BYOS), like this:

```
registercloudguest -r YOUR_CODE
```
More details: https://documentation.suse.com/sle-public-cloud/all/html/public-cloud/cha-admin.html#sec-admin-register

Assuming you use an appropriate SUMA code, and you're using the SUMA image, everything you need should get activated. Using this, instead of going directly to the SCC, potentially has advantages. You should still verify that everything is activated before moving on.

SUSEConnect --status-text and everything still work after you've done the registercloudguest.

## Version

SUMA 4.3 docs. SUMA 4.3 is still on SLES 15 SP4, and hasn't moved to 15 SP5.

keichwa:
Thanks for the report!

cwickert:
This looks like a duplicate of #2700 

@keichwa, what is the status here? Should be trivial to fix.

keichwa:
> This looks like a duplicate of #2700
> 
> @keichwa, what is the status here? Should be trivial to fix.

Probably.  Now I try this one: https://github.com/uyuni-project/uyuni-docs/pull/2841

PR #2770, opened 2024-02-06
4.3: Fix SLE SP wrong references

juliogonzalez:
# Description

Fix SLE SP wrong references. We still had references to SP5 as base OS.

# Target branches

- master: https://github.com/uyuni-project/uyuni-docs/pull/2771
- 4.3: This PR

# Links
- https://github.com/uyuni-project/uyuni-docs/issues/2733


parameters.yml
```diff
@@ -174,11 +174,13 @@ asciidoc:
   - attribute: sp-base
     value: SP4
   - attribute:  sp-version
-    value: SP5 #use for normal text
+    value: SP4 #use for normal text
   - attribute: sp-vert
```
juliogonzalez, line 178: I wonder why we need `sp-vert` and `sp-version-l` if the value is 100% the same but... OK :-D

parameters.yml
```diff
@@ -174,11 +174,13 @@ asciidoc:
   - attribute: sp-base
     value: SP4
   - attribute:  sp-version
-    value: SP5 #use for normal text
+    value: SP4 #use for normal text
```
juliogonzalez, line 177: Same as below, not clear to me why we need `sp-base` and `sp-version.

parameters.yml
keichwa: > I wonder why we need `sp-vert` and `sp-version-l` if the value is 100% the same but... OK :-D

long time ago, I defined `sp-version-l`.  It some incarnation of `sp-vert` or `sp-ver` in long format.  I do not remember exactly the values.  Maybe, "SP 5" vs. "Service Pack 5" or just "5" vs. "SP 5" ;)  We often do not know how to properly phase out a variable no longer needed.

parameters.yml
keichwa: One for the SP version SUSE Manager is based on (4.3 SP4), and the other latest supported SLE client (4.3 supports SP5)
IIRC, we want to switch to sp-server and sp-client for clarity.  I proposed it, Joseph agreed, but I hesitated to change it actually.

keichwa (approved):
Your changes look syntactically correct, but I'm not an expert with version numbers in general.  We'll need at least a second tic here.

parameters.yml
juliogonzalez: So what should the be right value then? Whatever entity is used for clients should have a clear name for it.. or at least a clear comment. Does this need to stay as SP5?

parameters.yml
jcayouette: lol not sure why I did not notice this before. We should keep these simple for sure. Phase it out! Phase it out! :)

jcayouette (commented):
See comments.

juliogonzalez:
> See comments.

@jcayouette I don't understand what to exactly about them. Do I need to rollback some values to SP4, or are they OK as they are in the PR?

If a refactor is needed, that would be for another PR :-)

jcayouette:
@juliogonzalez Your changes look good to me. Thank you!

jcayouette (approved):

PR #2779, opened 2024-02-12
Automatically build and publish doc (as zips) for master and manager-4.3

juliogonzalez:
# Description

- Automatically build and publish doc (as zips) for master and manager-4.3, on each push or run of `update_gettext.yml`. The files become available at the GitHub action page.
- Stop adding the `.key` files at `update_gettext.yml`
- Remove `.github/workflows/main_uyuni.yml`. We now do it with this new workflow.
- Remove `.github/workflows/build-and-release-docs.yml`. It was trying to do what I am doing now, but with releases, which will not work

This workflow can help answering questions such as:
- does `master` or `manager-4.3` build?
- is change X included in the build?
- is change X enabled in master for SUSE Manager and not only for Uyuni?
- are the `if`s correctly working in `master`?

You can see this running in my fork at https://github.com/juliogonzalez/uyuni-docs/actions/runs/7869591730

# Target branches

* Which product version this PR applies to: **ALL, but workflows work only from `master`, so no need to backport**
* Does this PR need to be backported? **NO**

# Links
- None


.github/workflows/build-and-archive-docs.yml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: Configure environment
+      run: |
+        alias docker='podman'
+        mkdir outputs
+
+    - name: Build the Uyuni documentation
+      run: |
+        cd uyuni-docs-helper
+        ./uyuni-docs-helper -r ${{ github.ref_name }} -o /tmp -c antora-uyuni-en -p uyuni
```
juliogonzalez, line 37: I think I am using the right target here, as this builds the doc for the site, and the PDFs, but please confirm (just in case).

.github/workflows/build-and-archive-docs.yml
```diff
@@ -0,0 +1,62 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: documentation-uyuni-${{ github.ref_name }}
+        path: uyuni/
+      if: github.ref_name == 'master'
+
+    - name: Build the SUSE Manager documentation
+      run: |
+        cd uyuni-docs-helper
+        ./uyuni-docs-helper -r ${{ github.ref_name }} -o /tmp -c antora-suma-en -p suma
```
juliogonzalez, line 53: Same as before

.github/workflows/build-and-archive-docs.yml
jcayouette: The target is correct. :+1:

.github/workflows/build-and-archive-docs.yml
jcayouette: Correct :+1:

jcayouette (approved):

jcayouette:
@juliogonzalez Really, a big thanks for this Julio. Excellent work.

PR #2784, opened 2024-02-13
Workflow cleanup

juliogonzalez:
# Description

Cleanup workflow as discussed:
- Building ES/CZ PDF translations is removed, not in scope for now
- Rest of languages are now consolidated into a single job: translation_tests_pdf.yml, reason being that we keep less actions running in parallel. They also now use the container image and `uyuni-docs-helper`
- Checkstyle and linting disabled (they can still be triggered manually, but they don't run automatically), so they don't fail on forks (as for now both actions are broken)
- `feature_tests_uyuni.yml` removed, as it was testing building the PDFs for simplified chinese, and we already have that in the translation tests.
- Release workflows removed, we agreed to handle that with branches, so I will incorporate this into `build-and-archive-docs.yml` later.
- Rework the README (disabled actions are still commented, but since most actions run for both `master` and `manager-4.3` we now show the results as tables)
- Finally, the syntax and format of workflows is unified (they all require only one job for now, called `run`).

@jcayouette do we always need to run the PDF test for non-english, or can it just be triggered when certain files are changed by a commit? Same question for updating the translation files :-)

See current rest results at https://github.com/juliogonzalez/uyuni-docs (they run against my fork, and the badges come from my fork, the code of this PR is adapted to take the badges from the main branches)

Tests: 

imagen

# Target branches

* Which product version this PR applies to: **Uyuni and SUSE Manager 5.0 and 4.3**
* Does this PR need to be backported? **NO, github actions are always configured only on the main branch**

# Links
None


jcayouette:
"@jcayouette do we always need to run the PDF test for non-english, or can it just be triggered when certain files are changed by a commit? Same question for updating the translation files :-)"

Sure we can choose to run these when language translation files are updated? We can have a discussion about this.

jcayouette (approved):
Looks great Julio, thank you!

juliogonzalez:
> Sure we can choose to run these when language translation files are updated? We can have a discussion about this.

Where are those files? I can just setup a file filter so the actions are only launched when they are changed :-)

PR #2789, opened 2024-02-13
Workflows to generate release zips

juliogonzalez:
# Description

In the  same way I created an action to build the site doc for Uyuni and SUSE Manager for development branches, this new action does something similar for release branches.

More specifically:

- Pushes to branches matching `uyuni-20[0-9][0-9].[0-9][0-9]` will build a zip file for the Uyuni site (all languages) and a zip for the Uyuni package (English only)
 - Pushes to branches matching `manager-4.3-MU-4.3.[0-9][0-9]?` will build a zip file for the SUSE manager site (all languages) and a zip for the SUSE Manager package (English only)
 - We should define the format for the alphas/betas/RCs, so I can add them to the workflow, as right now we this doesn't build them.

Results can be checked at https://github.com/uyuni-project/uyuni-docs/actions/workflows/build_and_archive_release_docs.yml

For this, I created two fake release branches `uyuni-2099.12` and `manager-4.3-MU-4.3.99` that I will remove after this PR is merged.

As an extra, this PR changes the other workflow and create the new one using `_` instead of `-`

# Target branches

* Which product version this PR applies to (Uyuni, SUMA 4.3, SUMA MU X.Y.Z, or SUMA development version).  **All, but GitHub actions only matter on the main branch, and will apply to all relevant branches**
* Does this PR need to be backported? **No, see above**

# Links
None

### DIFF
--- a/.github/workflows/build_and_archive_devel_docs.yml
+++ b/.github/workflows/build_and_archive_devel_docs.yml
@@ -1,4 +1,4 @@
-name: Build and archive documentation from branches
+name: Build and archive documentation from development branches
 
 on:
   push:
--- a/.github/workflows/build_and_archive_release_docs.yml
+++ b/.github/workflows/build_and_archive_release_docs.yml
@@ -1,0 +1,97 @@
+name: Build and archive documentation from release branches
+
+on:
+  push:
+    branches:
+      - 'uyuni-20[0-9][0-9].[0-9][0-9]' # Only master will build both Uyuni and SUSE Manager documentations
+      - 'manager-4.3-MU-4.3.[0-9][0-9]?' # Other branches will only build SUSE Manager documentation
+      - 'manager-5.0-[alpha|beta|rc|gmc|gm]*' # Other branches will only build SUSE Manager documentation
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout this repository
+      uses: actions/checkout@v4
+      with:
+        path: uyuni-docs
+
+    - name: Checkout uyuni-docs-helper repository
+      uses: actions/checkout@v4
+      with:
+        path: uyuni-docs-helper
+        repository: uyuni-project/uyuni-docs-helper
+
+    - name: Configure environment
+      run: |
+        alias docker='podman'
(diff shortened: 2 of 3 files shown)


.github/workflows/build_and_archive_devel_docs.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and archive documentation from branches
+name: Build and archive documentation from development branches
```
juliogonzalez, line 1: Adjusted, to avoid confusion.

.github/workflows/build_and_archive_release_docs.yml
```diff
@@ -0,0 +1,96 @@
+name: Build and archive documentation from release branches
+
+on:
+  push:
+    branches:
+      - 'uyuni-20[0-9][0-9].[0-9][0-9]' # Only master will build both Uyuni and SUSE Manager documentations
+      - 'manager-4.3-MU-4.3.[0-9][0-9]?' # Other branches will only build SUSE Manager documentation
```
juliogonzalez, line 7: We need to define the format for alpha/beta/rc/gmc branches

juliogonzalez: I suggest:
```
Manager-5.0-[Alpha|Beta|RC|GMC|GM]*
```

keichwa: In case it is case-sensitive, I'd like to go for "alpha|beta".  The rest can stay.

keichwa (commented):
Thanks, just a remark.

jcayouette (approved):

.github/workflows/build_and_archive_release_docs.yml
juliogonzalez: Done, I added:
```
      - 'manager-5.0-[alpha|beta|rc|gmc|gm]*' # Other branches will only build SUSE Manager documentation

juliogonzalez: Didn't fly, so I had to create and merge https://github.com/uyuni-project/uyuni-docs/pull/2794

PR #2852, opened 2024-03-08
Moved Client Onboarding to Cl.Conf Guide

0rnela:
Moved client onboarding to Cl.Conf Guide.

Part of CW reorganization / finding more suitable place for the existing files in other books / guides.


0rnela:
The file Client Onboarding, when moved from Common Workflows to Cl. Config Guide, can be found here:
ClientOnboarding

0rnela:
In conversation with @admd it has been decided that we will keep the CW book as is, i.e. we won't move the existing files to otehr books.

PR #2854, opened 2024-03-11
Moved Content Lifecycle Management WF to Admin Guide

0rnela:
CW book , as a collection of random files, can be reorganized so that some of the files are put where they belong more naturally.

This PR moves CLM workflow to the Admin Guide.


0rnela:
In conversation with @admd it has been decided that we will keep the CW book as is, i.e. we won't move the existing files to otehr books.

PR #2856, opened 2024-03-11
Moved Product Upgrade via WebUI WF to Client Conf. Guide

0rnela:
CW book , as a collection of random files, can be reorganized so that some of the files are put where they belong more naturally.

This PR moves Product Upgrade via WebUI workflow to the Client Config Guide.


0rnela:
In conversation with @admd it has been decided that we will keep the CW book as is, i.e. we won't move the existing files to other books.

PR #2952, opened 2024-04-18
Hub install container #24130 keichwa

keichwa:
https://github.com/SUSE/spacewalk/issues/24130


jcayouette (approved):

srbarrios (commented):
LGTM. But only @nadvornik can confirm :)

modules/specialized-guides/pages/large-deployments/hub-reqs.adoc
```diff
@@ -4,22 +4,34 @@
...
 == Peripheral Servers
 
+Deploy the peripheral servers with [command]``mgradm``.
+For more information about deploying with [command]``mgradm``, see xref:installation-and-upgrade:container-deployment/deployment-as-a-container.adoc[].
+
 Peripheral servers must be registered to the Hub Server as clients.
-When you register the peripheral servers, assign them the appropriate ``SUSE Manager Server`` software channel as their base channel.
-Additionally, they must be registered to the Hub Server directly, do not use a proxy.
+When you register the peripheral servers, assign them the appropriate [guimenu]``{productname} Server`` software channel as their base channel.
+The peripheral servers must be registered to the Hub Server directly, do not use a proxy.
```
nadvornik, line 34: Peripheral server must be registered as _foreign_ clients, that means the standard registration methods can't be used. Also, there are no channels involved.

To register the peripheral server, use mgradm command on the peripheral server host:
```
mgradm hub register --api-server <hub fqdn> --api-user <hub admin> --api-password <hub admin password>
```

Optionally, it is possible to register the podman container host as standard salt minion and manage it from SUMA.

nadvornik (changes requested):

modules/specialized-guides/pages/large-deployments/hub-reqs.adoc
srbarrios: Question @nadvornik , if we onboard both the container and the host, it will trigger a merge of registered systems as it happens with proxy container and proxy host? If so, that would affect somehow to Hub functionality?

nadvornik: There is no merging, they will stay separate.
Only the foreign client will have the "peripheral server" entitlement so the Hub API will use only this one.

nadvornik (approved):
LGTM

CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Use mgradm for Hub Server deployment in Large Deployment Guide
```
0rnela, line 1:
```suggestion
- Documented (Or Described) the use of  mgradm tool for Hub Server deployment in Large Deployment Guide
```

PR #2981, opened 2024-05-02
Fixes for Admin Guide / Mounting point correction

0rnela:
Clarify how we mount it, in a containerized server.
Use mgradm term and mgradm restart.

- master

# Links
- This PR tracks issue  https://github.com/SUSE/spacewalk/issues/24234 && https://github.com/SUSE/spacewalk/issues/24193

### DIFF
--- a/modules/administration/pages/disconnected-setup.adoc
+++ b/modules/administration/pages/disconnected-setup.adoc
@@ -191,7 +191,7 @@ mgr-sync add channel channel-label
 . Install the server container with 
 +
 ----
-mgradm install podman --mirrorPath /media/disk
+mgradm install podman --mirror /media/disk
 ----
 +
 . Restart the Tomcat service:


modules/administration/pages/disconnected-setup.adoc
```diff
@@ -33,7 +33,7 @@ We recommend you set up a dedicated RMT instance for each {productname} installa
 . On the RMT instance, install the RMT package:
 +
 ----
-zypper in rmt-server
+transactional-update pkg in rmt-server
```
cbosdo, line 36: Do we ask the RMT server to be on a SLE Micro? I understand that the RMT / SMT server should be on a different machine, so there is no need to change from `zypper` to `transactional-update pkg`

modules/administration/pages/disconnected-setup.adoc
```diff
@@ -105,7 +105,8 @@ For more details on installing and configuring SMT, see https://documentation.su
 smt-sync --createdbreplacementfile /tmp/dbrepl.xml
 ----
 . Export the synchronized data to your removable storage.
-    In this example, the storage medium is mounted at [path]``/mnt/usb``:
+//    In this example, the storage medium is mounted at [path]``/mnt/usb``:
+    TO BE CHECKED WITH DEV: In this example, the storage medium is mounted in the containerized {productname} Server on [path]``/mnt/usb``:
```
cbosdo, line 109: This part is to be executed on a different machine that has access to the repos and internet, so no Uyuni container involved at that point.

The server container should be installed with the `mgradm install podman --mirrorPath /media/disk` rather than adding `server.susemanager.fromdir = /media/disk` to `/etc/rhn/rhn.conf`. At the moment we have no way to change this option after the installation.

The `systemctl` and `mgr-sync` commands in the last procedure should be prepended with `mgrctl exec --  ` to be executed in the container.

0rnela:
@cbosdo I have tried to incorporate your comments and would appreciate your feedback.

modules/administration/pages/disconnected-setup.adoc
```diff
@@ -182,6 +183,35 @@ mgr-sync refresh
...
 ----
+////
+
+// NEW: 
+
+.Procedure: Synchronizing a Disconnected Server
+. Install the server container with 
++
+----
+mgradm install podman --mirrorPath /media/disk
```
cbosdo, line 194: I have a pending PR (https://github.com/uyuni-project/uyuni-tools/pull/324) changing the `--mirrorPath` to `--mirror`. You may want to use directly the new one in this doc.
```suggestion
mgradm install podman --mirror /media/disk
```

cbosdo (approved):
You probably want to adjust `--mirrorPath` now before needing another PR in a few days / hours.

PR #2997, opened 2024-05-07
Fixes in the Admin Guide, ISS v2

0rnela:
# Description

 Remind the user to use "mgrctl term" before running steps inside the server container
 
Backport targets:

- master

# Links
- This PR tracks issue  https://github.com/SUSE/spacewalk/issues/24268 && https://github.com/SUSE/spacewalk/issues/24193.


jcayouette:
I am not sure how the ISS packaging works with containers. Are packages persisted when installed?  They should be stored on some shared volume but maybe run this by Cedric for verification?

PR #2999, opened 2024-05-08
Ccg imbz and others review #24193

keichwa:
https://github.com/SUSE/spacewalk/issues/24193


srbarrios (approved):
Approving.
I think it would be good to describe better what this is fixing, I got we are fixing the issues in some rows and columns in some tables, but it's not obvious by the title or the description :)

jcayouette (approved):

PR #3024, opened 2024-05-15
43 - Removed Unified Installed from the Retail Guide

0rnela:
Removed Unified Installed from the Retail Guide.

- [x] master https://github.com/uyuni-project/uyuni-docs/pull/3021
- [x]  manager 4.3


aaannz:
You could before we moved to the VM image. But now only supported way is VM based install.

aaannz:
But let's check with @admd if we should remove this in 4.3 branch as well, or keep it there even thou unsupported.

admd:
@0rnela 
For 4.3, instead of removing it completely, please add a note on top that this is unsupported and only supported way is VM based install.

jcayouette (approved):

keichwa (changes requested):
Hi Ornela, in the last days, Abid commented above.

0rnela:
Closing -see https://github.com/uyuni-project/uyuni-docs/pull/3100.

PR #3025, opened 2024-05-15
containerized commands centos (beta 2 review)

keichwa:
see https://github.com/SUSE/spacewalk/issues/24328

# Some hints

Consider adding an entry to the `CHANGELOG.md` file in the toplevel directory.
Cosmetic changes such as fixing typos do not need log entries (nevertheless it is important to fix typos, etc.)!
In the `manager-4.3`, the hidden `.changelog` file with a leading dot is still in use.

Add Description, Target branches, and related Links below the following section titles.

In the Description, just enter a summary of why you created this PR (and, if available, add any relevant diagram).

# Description

Short summary of why you created this PR (if you added documentation, please add any relevant diagram).

# Target branches

- master
- 4.3

# Links

### DIFF
--- a/modules/client-configuration/pages/clients-centos.adoc
+++ b/modules/client-configuration/pages/clients-centos.adoc
@@ -158,8 +158,14 @@ include::snippets/create_bootstrap_repo_register.adoc[]
 
 
 
+ifeval::[{uyuni-content} == true]
 
 == Manage Errata
+
+[NOTE]
+====
+This section is only applicable to non-containerized Uyuni and it will be removed after {centos} 7 will go end-of-life.
+====
 
 When you update {centos} clients, the packages do not include metadata about the updates.
 You can use a third-party errata service to obtain this information.
@@ -167,12 +173,12 @@ You can use a third-party errata service to obtain this information.
 [WARNING]
 ====
 
-ifeval::[{suma-content} == true]
-
-The third-party errata service described here, CEFS, is provided and maintained by the community.
-It is not supported by {suse}.
-
-endif::[]
+// ifeval::[{suma-content} == true]
+//
+// The third-party errata service described here, CEFS, is provided and maintained by the community.
+// It is not supported by {suse}.
+//
+// endif::[]
 
 The authors of CEFS provide patches or errata on a best-effort basis, in the hope they are useful but with no guarantees of correctness or currency.
 This could mean that the patch dates could be incorrect, and in at least one case, the published data was shown to be more than a month old.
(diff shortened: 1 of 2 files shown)


mbussolotto (changes requested):
I don't think this procedure could work. I think we need to pre-install `perl-Text-Unidecode` in server-image, then:
- `mgrctl exec -- systemctl restart tomcat` should be subtiture by `mgradm restart`
- `/usr/local/bin/cent-errata.sh` should be moved to a persisted location

@cbosdo @rjmateus  JFYI, let me know if I missed anything

modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -188,30 +188,36 @@ Please consider alternatives to using this service, such as independently verify
...
 +
 ----
-SUSEConnect --product sle-module-development-tools/15.2/x86_64
+mgrctl exec -- SUSEConnect --product sle-module-development-tools/15.5/x86_64
 ----
 . Install errata service dependencies:
 +
 ----
-zypper in  perl-Text-Unidecode
+transactional-update pkg install perl-Text-Unidecode
```
srbarrios, line 199: Remember that after any action with transaction-update, we must reboot the system to make the change real.

modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -236,7 +242,7 @@ export SPACEWALK_USER='<adminname>';export SPACEWALK_PASS='<password>'
 . Set up a cron job to run the script daily:
 +
 ----
-ln -s /usr/local/bin/cent-errata.sh /etc/cron.daily
+mgrctl exec -- ln -s /usr/local/bin/cent-errata.sh /etc/cron.daily
```
cbosdo, line 245: Adding cron files in the container won't fly. However the `errata-import.pl` script that the job runs uses the XML-RPC API and could run on the host.

In this case we should just document to add this to the host and not in the container.

This means that the script file shouldn't be created in the container but in the host.

modules/client-configuration/pages/clients-centos.adoc
cbosdo: This would be fine to run the script (and the cron job) on the host.

modules/client-configuration/pages/clients-centos.adoc
```diff
@@ -188,30 +188,36 @@ Please consider alternatives to using this service, such as independently verify
 
 .Procedure: Installing an Errata Service
 
-. On the {productname} Server, from the command prompt, as root, add the ``sle-module-development-tools`` module:
+. On the {productname} Server host, from the command prompt, as root, add the [systemitem]``sle-module-development-tools`` module:
 +
 ----
-SUSEConnect --product sle-module-development-tools/15.2/x86_64
+mgrctl exec -- SUSEConnect --product sle-module-development-tools/15.5/x86_64
```
cbosdo, line 194: To run the script on the host, the host needs to have the Development tools module. It should be checked if that works on SLE Micro 5.5 thought

keichwa: Ok, I'll change the description accordingly but need someone who can verify it.

rjmateus:
Outcome of the discussion/research: Keep it for uyuni and remove the all section for SUSE Manager.
@admd anything to adD?

admd:
@keichwa these changes will not work!

This workaround to get the errata information was needed for centos, we only have CentOS 7 right now which is going EOL in June. So I would say keep the original text but only Uyuni and maybe add a note in this section that this is only applicable to non-containerized Uyuni and this section will be removed after CentOS 7 will go EOL.

keichwa:
> ... So I would say keep the original text but only for Uyuni and maybe add a note in this section that this is only applicable to non-containerized Uyuni and this section will be removed after CentOS 7 will go EOL.

Thanks, makes a lot of sense.  Done.

PR #3058, opened 2024-06-05
Remove the need to manually add the container utils in deployment page - installation guide

raulillo82:
Remove the need to manually add the container utils in deployment page - installation guide

# Some hints

Consider adding an entry to the `CHANGELOG.md` file in the toplevel directory.
Cosmetic changes such as fixing typos do not need log entries (nevertheless it is important to fix typos, etc.)!
In the `manager-4.3`, the hidden `.changelog` file with a leading dot is still in use.

Add Description, Target branches, and related Links below the following section titles.

In the Description, just enter a summary of why you created this PR (and, if available, add any relevant diagram).

# Description

Remove references to manually installing the container tools, with GMC they come automatically when adding the extension.

# Target branches

* Which product version this PR applies to (SUMA 5.0)
* Does this PR need to be backported? I don't think so, Uyuni seems to be different.

- master -> No backports.

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/23646

### DIFF
--- a/modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-suma.adoc
+++ b/modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-suma.adoc
@@ -173,7 +173,6 @@ transactional-update
 ----
 
 . Reboot.
-. Log in as root.
 
 
 == Deploy with mgradm


modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-suma.adoc
```diff
@@ -174,17 +174,6 @@ transactional-update
 
 . Reboot.
 . Log in as root.
-. Install the container utilities:
```
juliogonzalez, line 177: I don't think we need the previous line, 176?

juliogonzalez (changes requested):
Something small.

Also, before merging this, please track the outcome of https://suse.slack.com/archives/C02CLB6AGJ3/p1717576708848709 as it looks to me we'll need to add a warning.

modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-suma.adoc
raulillo82: Removed

0rnela (approved):

juliogonzalez (approved):
After we discussed in the daily today, LGTM.

jcayouette:
@raulillo82 I want to be certain we have this correct. This page is for ISO deployment it is not for images. We should remove the `podman` package requirement and keep the installation for utilities. If I understand correctly this step is not required for images, not the ISO on top of SLE Micro 5.5?

juliogonzalez:
> @raulillo82 I want to be certain we have this correct. This page is for ISO deployment on SLE Micro 5.5, It is not for images. We should remove the `podman` package requirement and keep the installation command for utilities. If I understand correctly this step is not required for images, but IS required for ISO deployment on top of SLE Micro 5.5?

Apparently this is not needed in neither case, unless you are deploying on top of SLE Micro 5.5 on public cloud, which is what I did and it's not supported.

Seems the ISO already installs podman, and the SLE Micro 5.5 images for KVM and non-public cloud include podman as well.

deneb-alpha (approved):
LGTM

jcayouette (approved):

PR #3087, opened 2024-06-17
Removal of unnecessary step

0rnela:
Removed the step from the procedure, as per

https://github.com/SUSE/spacewalk/issues/24532


0rnela:
@jcayouette -- I do not see this file in the navigation bar. Can you please advise where it should go?

keichwa (approved):

jcayouette (approved):

PR #3100, opened 2024-06-20
[WIP] Updating retail unified installer

0rnela:
(https://github.com/uyuni-project/uyuni-docs/pull/3024#issuecomment-2124430383)

### DIFF
--- a/modules/retail/pages/retail-install-unified.adoc
+++ b/modules/retail/pages/retail-install-unified.adoc
@@ -14,7 +14,7 @@ For information on registering with {scc}, retrieving your organization credenti
 [WARNING]
 ====
 This method is no longer supported.
-The supported way is going forward is VM based install - **link needed here**.
+The supported way is going forward is VM based install.
 ====
 
 [NOTE]


0rnela:
RECAP

**INTRO**
Retail installation is described in:

- SUMA 4.3 https://documentation.suse.com/suma/4.3/en/suse-manager/retail/retail-install-unified.html
- SUMA 5.0 https://documentation.suse.com/suma/5.0/en/suse-manager/retail/retail-install-setup.html

**PROGRESS**
We had issue:
https://github.com/uyuni-project/uyuni-docs/pull/3024
which became
https://github.com/uyuni-project/uyuni-docs/pull/3100

**DILEMMA**
With the comment in one of the discussions that that we now provide only VM based installation in 4.3, I need to clarify if SUMA 4.3 needs any extra content apart from warning that Unified Installer is no longer supported - if YES, which I expect because we need to redirect the customer to the alternative, 
than **_what is that content?_**

aaannz:
Indeed images are now the primary installation source/deployment.

For retail that means after suma deployment as usual, they will need to install suma_retail pattern (`zypper install -t pattern suma_retail`) and that's it.
BranchServer installation is the same like proxy deployment, but use RBS SCC key to register correct product.

Issue #2705, opened 2024-01-04
recommendation for worker_threads in the section specialized guides > Large Deployment Guide > Tunig Large Installations should be changed


thsnielsen:
## Documentation Update Required
yes update required. The current advice could be damaging for large environments

https://documentation.suse.com/suma/4.3/en/suse-manager/specialized-guides/large-deployments/tuning.html#worker-threads

## What does the documentation say now?
Value recommendation 8-200
...
Example worker_threads: 50

Description of existing documentation.
Include screenshot if required.
![image](https://github.com/uyuni-project/uyuni-docs/assets/16719354/f08eb83b-7eef-446b-b279-6646ce7f87de)

## What change is required?
The recommendation should be lowered to f.ex.
Value recommendation 8-<max number of cpu's in the server -1>
...
Example worker_threads: 15

What content needs to be added?
nothing needs to be added

## Subject Matter Expert (SME)
Victor Zhestkov
see this presentation at 14min25sec  in : 
https://www.youtube.com/watch?v=dmuh4gfrEnQ&t=845s

## Version

- [x] 4.3
- [x] master - https://github.com/uyuni-project/uyuni-docs/pull/3155
- [x] 5.0

mcalmer:
@0rnela this is important. Would be good if you could find a few minutes for this change

keichwa:
merged and backported to 5.0 and 4.3.

PR #3155, opened 2024-07-03
Improve tuning guide

vzhestkov:
# Description

Improve large scale guide and add extra section about Salt Connectivity

# Target branches

- master
- 5.0

# Links

### DIFF
--- a/modules/specialized-guides/pages/salt/salt-connectivity.adoc
+++ b/modules/specialized-guides/pages/salt/salt-connectivity.adoc
@@ -14,7 +14,7 @@ To understand such a connection issue check if the client is not reachable from 
 If this is the case, it is recommended to set [literal]``tcp_keepalive`` parameters.
 
 The parameters with the values from the example below could be used as a starting point to look for the better combination, which could prevent cases of connection loss without restoring for some environments.
-It's recommended to put the parameters in the separate drop-in configuration file like [path]``/etc/venv-salt-minion/minion.d/tuning-keepalives.conf`` or [path]``/etc/salt/minion.d/tuning-keepalives.conf``, depending on the minion type used on the client side.
+It is recommended to put the parameters in the separate drop-in configuration file like [path]``/etc/venv-salt-minion/minion.d/tuning-keepalives.conf`` or [path]``/etc/salt/minion.d/tuning-keepalives.conf``, depending on the minion type used on the client side.
 
 .Example: Keepalive parameters example values
 


keichwa:
Thanks for the PR.  I'll propose some cosmetic fixes soon.

It also adresses https://github.com/uyuni-project/uyuni-docs/issues/2705 and that is a good thing!

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+- Added Salt Connectivity section to the Specialized guides
+- Improved large deployments guide with better values and
+  extra parameters added.
```
keichwa, line 3:
```suggestion
- Added Salt Connectivity section to the Specialized Guides
- Improved Large Deployments Guide with better tuning values and
  extra parameters added
```

modules/specialized-guides/pages/large-deployments/tuning.adoc
```diff
@@ -306,14 +306,14 @@ This section contains information about the available parameters.
 [cols="1,1"]
 |===
 | Description          | Before any action is executed on a client, a presence ping is executed to make sure the client is reachable.
-                         This parameter sets the amount of time before a second command (`find_job`) is sent to the client to verify its presence.
+                         This parameter sets the amount of time before a second command (in most cases `state.apply` or any other salt function) is sent to the client to verify its presence.
```
keichwa, line 309:
```suggestion
                         This parameter sets the amount of time before a second command (in most cases `state.apply` or any other {salt} function) is sent to the client to verify its presence.
```

modules/specialized-guides/pages/large-deployments/tuning.adoc
```diff
@@ -327,15 +327,15 @@ This section contains information about the available parameters.
 [cols="1,1"]
 |===
 | Description          | Before any action is executed on a client, a presence ping is executed to make sure the client is reachable.
-                         After xref:java-salt-presence-ping-timeout[`java.salt_presence_ping_timeout`] seconds have elapsed without a response, a second command (`find_job`) is sent to the client for a final check.
-                         This parameter sets the number of seconds after the second command after which the client is definitely considered offline.
+                         After xref:java-salt-presence-ping-timeout[`java.salt_presence_ping_timeout`] seconds have elapsed without a response, a second command (in most cases `state.apply` or any other salt function) is sent to the client and if there is no response from the clinet for the amount of seconds specified with this parameter one more call (`saltutil.find_job`) is sent for a final check.
```
keichwa, line 330:
```suggestion
                         After xref:java-salt-presence-ping-timeout[`java.salt_presence_ping_timeout`] seconds have elapsed without a response, a second command (in most cases `state.apply` or any other {salt} function) is sent to the client and if there is no response from the client for the amount of seconds specified with this parameter one more call (`saltutil.find_job`) is sent for a final check.
```

modules/specialized-guides/pages/large-deployments/tuning.adoc
```diff
@@ -327,15 +327,15 @@ This section contains information about the available parameters.
...
+                         This parameter sets the number of seconds after the second command after which the client is definitely considered timedout.
```
keichwa, line 331:
```suggestion
                         This parameter sets the number of seconds after the second command after which the client is definitely considered timeout.
```

modules/specialized-guides/pages/large-deployments/tuning.adoc
```diff
@@ -578,18 +578,54 @@ If you see lines containing ``LOG:  temporary file:`` try raising this parameter
...
-| Tune when            | xref:client-count[Client count] increases significantly, xref:thread-pool[`thread_pool`] increases significantly, or `SaltReqTimeoutError` or `Message timed out` errors appear in `/var/log/salt/master`.
+                         Setting this value to very high values could cause opposite effect as the workers will compete to each other for the CPU resources and the performance could be dropped significantly.
+| Tune when            | xref:client-count[Client count] increases significantly, xref:thread-pool[`thread_pool`] increases significantly, or `SaltReqTimeoutError` or `Message timed out` errors appear in `/var/log/salt/master` could be a sign of too low or too high value of this parameter.
 | Value default        | 8
-| Value recommendation | 8-200
+| Value recommendation | 8-32, depending on the number of the CPU cores available for the server, it's recommended to keep the value slightly less than the number of CPU cores.
```
keichwa, line 584:
```suggestion
| Value recommendation | 8-32, depending on the number of the CPU cores available for the server, it is recommended to keep the value slightly less than the number of CPU cores.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+Depending on the size and spreadness of the environment where {productname} is used some connectivity issues are possible.
+There are no common recommendations for all the possible use cases as the environments could be very different expecially if public clouds instances are involved.
+
+
+
+[[minions-connectivity]]
+== Minions Connectivity
+
+In case if clients are losing the connection to the {productname} Server (or {productname} Proxy if involved),
+they are unreachable from the {productname} Server web UI or using the CLI.
```
keichwa, line 13:
```suggestion
In case clients are losing the connection to the {productname} Server (or {productname} Proxy if involved), they are unreachable from the {productname} Server {webui} or using the CLI.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+To ensure if it's the case, it's better to check if the client is not reachable from the {productname} Server with [command]``salt MINION_ID test.ping``,
+while [command]``venv-salt-call test.ping`` or ([command]``salt-call test.ping`` in case if non-bundle salt is used on the client)
+is working fine on the client side, it is recommended to set `tcp_keepalive` parameters.
```
keichwa, line 16: Sentence is too long.  I'm not sure if I got it right.  Maybe, we can try sth like this:

```suggestion
In such a case, first check the connection from the {productname} Server with [command]``salt MINION_ID test.ping``.
[command]``venv-salt-call test.ping`` or ([command]``salt-call test.ping`` in case non-bundle {salt} is used on the client) is working fine on the client side, especially if `tcp_keepalive` parameters are set.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+The parameters with the values from the example below could be used as a starting point to look for the better combination, which could work better for certain environment.
```
keichwa, line 18:
```suggestion
The parameters with the values from the example below could be used as a starting point to look for the better combination, which could improve performance for some environments.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+It's recommended to put the parameters in the separate drop-in config file like [path]``/etc/venv-salt-minion/minion.d/tuning-keepalives.conf`` or [path]``/etc/salt/minion.d/tuning-keepalives.conf``, depending on the minion type used on the client side.
```
keichwa, line 19:
```suggestion
It's recommended to put the parameters in the separate drop-in configuration file like [path]``/etc/venv-salt-minion/minion.d/tuning-keepalives.conf`` or [path]``/etc/salt/minion.d/tuning-keepalives.conf``, depending on the minion type used on the client side.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+
+[[proxies-connectivity]]
+== Proxies Connectivity
+
+[command]``salt-broker`` service is used on {productname} Proxies to forward `salt` traffic between the {productname} Server
+and [command]``salt-minion`` services used on the client side.
+It's possible that [command]``salt-broker`` and all the clients behind it could be affected with the same issue as the clients
+directly connected to the {productname} Server.
+The issue could be fixed with the same parameters as recommended for the minions, but specified in [path]``/etc/salt/broker`` on each {productname} Proxy.
```
keichwa, line 62:
```suggestion
[command]``salt-broker`` service is used on {productname} Proxies to forward `salt` traffic between the {productname} Server and [command]``salt-minion`` services used on the client side.
It is possible that [command]``salt-broker`` and all the clients behind it could be affected with the same issue as the clients directly connected to the {productname} Server.
The issue could be fixed with the same parameters as recommended for the minions, but specified in [path]``/etc/salt/broker`` on each {productname} Proxy.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+
+The other possible issue which {productname} Proxy can be affected with is the case when the connectivity to the {productname} Server
+is lost for quite long interval, so the clients behind it started to retry the authentication to the [command]``salt-master`` service on the {productname} Server.
+This situation could be potentially dangerous as it could lead to collecting large amount of ZeroMQ messages with authentication attemps
+in the interal buffer of ZeroMQ sockets used inside the [command]``salt-broker`` service, so that on restoring the connection to the [command]``salt-master``
+all of the messages will be pushed to it. It could lead to the issues on {productname} Server side with [command]``salt-master`` service,
+as it could be impossible to serve all the cached requests in apropriate time or even to the complete denial of the service.
```
keichwa, line 70:
```suggestion
all of the messages will be pushed to it.
It could lead to the issues on {productname} Server side with [command]``salt-master`` service, as it could be impossible to serve all the cached requests in apropriate time or even to the complete denial of the service.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+
+To avoid such situation a set of extra parameters was introduced, the most important one is ``wait_for_backend``, which sould be set to ``True``,
+it prevents opening the sockets for the clients behind the proxy while the connectivity to the [command]``salt-master`` service is not established.
```
keichwa, line 73:
```suggestion
To avoid such situation a set of extra parameters was introduced.
The most important one is [option]``wait_for_backend``, which should be set to [literal]``True``.
This prevents opening the sockets for the clients behind the proxy while the connectivity to the [command]``salt-master`` service is not established.
```

modules/specialized-guides/pages/salt/salt-connectivity.adoc
```diff
@@ -0,0 +1,156 @@
...
+In this case the messages from the clients are not collected in the internal buffers.
+``drop_after_retries`` is setting the number of retries before closing the sockets to drop the cached messages.
+The other parameters could help to fine tune the behaviour for the environment.
+
+[NOTE]
+====
+Setting timouts, intevals and ``drop_after_retries`` to lower values are making [command]``salt-broker`` service more agressive on detecting
+the connection loss to the [command]``salt-master``, so that it puts the clients behind the proxy to the conditions closer as they are connected
```
keichwa, line 81:
```suggestion
Setting timeouts, intervals and ``drop_after_retries`` to lower values are making [command]``salt-broker`` service more aggressive on detecting the connection loss to the [command]``salt-master``, so that it puts the clients behind the proxy to the conditions closer as they are connected
```

keichwa (changes requested):
You can submit several suggestions in one go if you start from the "Files changed" tab, and then add them to batches.

modules/specialized-guides/pages/salt/salt-connectivity.adoc
vzhestkov: Looks like the meaning has changed here. The idea here that the minion service can lose the connection while didn't get any notification about it, so minion is still holds dead tcp socket which can't be used and it's not trying to reconnect. To ensure if it's the case one command from the server should fail while the commands `...salt-call` from the client is not failing, and in this case it makes sense to add `tcp_keepalive...` parameters to the minion config to avoid such cases.

modules/specialized-guides/pages/salt/salt-connectivity.adoc
vzhestkov: It's not about the performance, but mostly about improving the connection to prevent the cases of connection loss with no restoring.

modules/specialized-guides/pages/salt/salt-connectivity.adoc
vzhestkov:
```suggestion
[command]``salt-broker`` service is used on {productname} Proxies to forward `salt` traffic between the {productname} Server and [command]``salt-minion`` service used on the client side.
It is possible that [command]``salt-broker`` and all the clients behind it could be affected with the same issue as the clients directly connected to the {productname} Server.
The issue could be fixed with the same parameters as recommended for the minions, but specified in [path]``/etc/salt/broker`` on each {productname} Proxy.
```

vzhestkov: Just 1 small change from your suggestion, just noticed not really good typo

vzhestkov:
@keichwa I put all of these to the single PR as it's mostly the result of collecting the issues which could happen in combination, as we already have at least 2 similar cases. I just afraid I can get lost among all of the PRs I created regarding these cases.

modules/specialized-guides/pages/salt/salt-connectivity.adoc
keichwa: Ok, I see.  Maybe, "could stabilize connection"  or "could stabilize server-client connection" or "could make connection stable".  Some with "avoid interruption"?  If you think all this is misleading, we can keep your initial wording "work better".  (I wanted to get rid of the two "betters" in one long sentence.)

modules/specialized-guides/pages/salt/salt-connectivity.adoc
keichwa: I think I understand now want it is about.  Maybe, I'll come up with a proposal later.

keichwa: Next try:
```suggestion
To debug such a situation, check whether both communication directions between server and client are working.
On the server, check with [command]``salt MINION_ID test.ping`` that the server can reach the client.
If it fails, check on the client with [command]``venv-salt-call test.ping`` whether the server is reachable from the client.
If the communication from the client is working fine, it is recommended to set `tcp_keepalive` parameters on the client.

[NOTE]
====
In case the salt bundle is not used on the client, run [command]``salt-call test.ping`` instead of [command]``venv-salt-call test.ping``.
====
```
Maybe, we should also add the technical details about the dead TCP socket?  Maybe, instead of "communication" better the simpler term "connection"?

modules/specialized-guides/pages/salt/salt-connectivity.adoc
keichwa: I now think it is better to use your words.

keichwa (changes requested):
I changed my proposals.

modules/specialized-guides/pages/salt/salt-connectivity.adoc
keichwa: Yes, is the most important thing here.  With the following proposal I try to keep the test scenario, and mostly only change the intro phrase.  Then I propose the solution in a short final sentence.  We could add the background info you provide, starting from "2 parallel channels with the same direction open. And the idea of the check is to ... etc.".

Or we just merge it as-is now, and then ask Joseph for help.

keichwa (approved):

vzhestkov:
@keichwa thanks for the suggested fixes :+1: and sorry for being nitpicky for some of them, I'm tryin to prevent any confusions from the users side.

PR #3168, opened 2024-07-09
migration chapter: replace source/destination terms with old/new resp. version numbers

keichwa:
https://github.com/SUSE/spacewalk/issues/24697 **<== here you'll also find a link to the rendered output!**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-- Consolidated terminology around the old 4.3 server and the migration
-  target in Installation and Upgrade Guide
+- Consolidated terminology around the old server and the migration target
+  in Installation and Upgrade Guide
 - Added more information about the Prometheus alert manager in
   Administration Guide
 - Add storage extra configuration for migration step in Installation and


keichwa:
Tricky.  This covers suma and uyuni.

uyuni has the special issue that "standard" and legacy migration are identical.  "standard" migration (migrate-server-uyuni.html) probably must be replaced with "migrate old containerized server installation [= uyuni **pre-**2024.07] to new containerized server installation [= uyuni 2024.07]"

keichwa:
> uyuni has the special issue that "standard" and legacy migration are identical. "standard" migration (migrate-server-uyuni.html) probably must be replaced with "migrate old containerized server installation [= uyuni **pre-**2024.07] to new containerized server installation [= uyuni 2024.07]"

Crap.

"standard" now should probably cover "normal" server installation to the new containerized server installation.

"legacy" is migrating an old "normal" server installation to a new server, but still not containerized.

It's confusing.  I guess we first must create proper issues cards where we define what we want to have documented in the uyuni case... ;)

PR #3192, opened 2024-07-17
Replaced old {micro} parameter with new one {sle-micro} for better clarity

0rnela:
# Description

Analysis of usage of SL Micro and SLE Micro. 

Part 1 - using renamed parameter, for better clarity.

# Target branches

- master
- 4.3 https://github.com/uyuni-project/uyuni-docs/pull/3195
- 5.0 https://github. com/uyuni-project/uyuni-docs/pull/3197

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/24701

PR #3197, opened 2024-07-18
5.0 - Replaced old {micro} parameter with new one {sle-micro} for better clarity

0rnela:
# Description

Analysis of usage of SL Micro and SLE Micro. 

Part 1 - using renamed parameter, for better clarity.

# Target branches

- master https://github.com/uyuni-project/uyuni-docs/pull/3192
- 4.3 #3195
- 5.0

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/24701


keichwa (approved):
I'm not an expert with these names, but if it still builds it is probably fine.

PR #3204, opened 2024-07-22
removes duplicate migration file

jcayouette:
As title suggests removes duplicate legacy migration method. We list it under the legacy section.


keichwa (approved):
thanks, LGTM!

PR #3256, opened 2024-08-12
update for uyuni-2024.08 ghpages

jcayouette:
Content update for Uyuni 2024.08

PR #3277, opened 2024-08-23
Update for uyuni-2024.08 ghpages

juliogonzalez:
Content update for Uyuni 2024.08

We can't use https://github.com/uyuni-project/uyuni-docs/pull/3256 because:
- We had to fix things
- We are unable to build the doc again for all languages, a japanese fails to build (I created a card about it), so we built only for English and remove the other languages for now.


keichwa:
Marina Latini ***@***.***> writes:

> Merged #3277 into gh-pages.

Thank you - Julio for generating, Marina for merging!

-- 
Karl Eichwalder

PR #3366, opened 2024-10-15
Added info for mgr-ssl-cert-setup (bcs#1229079)

0rnela:
# Description

Missing instructions in the documentation, when running the commands in the container.  I have added the same text in the form of admonition as in the other documents where the same instruction applies.

# Target branches

- master
- 5.0 https://github.com/uyuni-project/uyuni-docs/pull/3371

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/25051
- https://bugzilla.suse.com/show_bug.cgi?id=1229079#c20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-- Added information for running mgr-ssl-cert-setup with 5.0 and containers
-  (bcs#1229079)
+- Added information for running mgr-ssl-cert-setup in Administration Guide
+  (bsc#1229079)
 - Added reference to Inter-Server Synchronization in Administration Guide
   (bsc#1230943)
 - Documented consistently Leap Micro as Uyuni container host system in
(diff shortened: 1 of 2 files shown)


modules/administration/pages/ssl-certs-imported.adoc
```diff
@@ -70,6 +68,12 @@ Certificates signed with different CAs do not match.
 
 == Replace Certificates
 
+
+[IMPORTANT]
+====
+Use [literal]``mgrctl term`` before running steps inside the server container.
```
mcalmer, line 74: I think the key point here is, that the customers do not know that they need to run this inside of the container.
So maybe we rephrase this a bit:

>   The following procedure need to run inside of the container. Use `mgrctl term` to open a terminal in the container before you execute it.

or something like this.

keichwa (approved):
with mc's proposal.

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+- Added information for running mgr-ssl-cert-setup with 5.0 and containers
+  (bcs#1229079)
```
keichwa, line 2:
```suggestion
- Added information for running mgr-ssl-cert-setup with containers
  (bsc#1229079)
```
Better w/o version number.  Uyuni is also affected ;)

PR #3394, opened 2024-10-24
5.0 - Fixed links after page removal

0rnela:
# Description

Supported clients file is now shown only in the Client Config guide. The corresponding file with duplicate content has been removed from the Installation Guide, but the old file was still referenced. This has now been taken care of. Remaining links and any other references to the removed file have been addressed too. 

# Target branches

- master 
- 5.0 https://github.com/uyuni-project/uyuni-docs/pull/3394

# Links
- This PR tracks issue  - none. I found this one while working on other issues.


0rnela:
Will be replaced with the tidy new PR.

PR #3389, opened 2024-10-23
minor fixes, discovered while working on matching proxy deployment

keichwa:
https://github.com/uyuni-project/uyuni-docs/pull/3385

Backport target:

- 5.0

PR #3385, opened 2024-10-23
VMware image deployment documentation for Proxy

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/25264

- https://bugzilla.suse.com/show_bug.cgi?id=1227852
- https://github.com/SUSE/spacewalk/issues/25264

PR #3397, opened 2024-10-28
Backport 5.0: iau server vm image cosmetic#3385

keichwa:
https://github.com/uyuni-project/uyuni-docs/pull/3389

### DIFF
--- a/modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-vm-suma.adoc
+++ b/modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-vm-suma.adoc
@@ -117,17 +117,17 @@ image::root-password.png[]
 
 
 
-== Register {sl-micro} and {productname} {productnumber} Server
+== Register {sle-micro} and {productname} {productnumber} Server
 
 // note for 5.0 backport: 5.0 should have sle-micro
 
 
-.Procedure: Registering {sl-micro} and {productname} {productnumber}
+.Procedure: Registering {sle-micro} and {productname} {productnumber}
 . Boot the virtual machine.
 . Log in as `root`.
 +
 // note for 5.0 backport: 5.0 should have sle-micro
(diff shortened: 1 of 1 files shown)


modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-vm-suma.adoc
```diff
@@ -168,7 +174,7 @@ mgr-storage-server /dev/nvme1n1 /dev/nvme2n1
 +
 [NOTE]
 ====
-This command will create the persistent storage volumes at [path]``/var/lib/containers/storage/volumes``.
+This command will move the persistent storage volumes at [path]``/var/lib/containers/storage/volumes`` to specified storage devices.
```
jcayouette, line 177: From what I understand this should be creates rather than move. This tool is creating these volumes. Its not really moving anything. Perhaps someone can verify?

keichwa: Probably both.  Because we now do it after deployment, it creates the storage device, then moves any already existing data (volume) to the storage device,  and mounts it (as the volume).  I'm neither sure about the sequence of these commands nor about the terminology we have to use here.

Ondrej proposed such a change here: https://github.com/uyuni-project/uyuni-docs/pull/3385#discussion_r1812321720

keichwa: @jcayouette can we keep the phrase as-is now or shall we address both ways more details.  In any case, I'd propose to merge this now, and create a separate PR as needed.  The new PR then could all places where we now have such a note and turn it into a snippet (or something similar).

jcayouette: Im fine with your change. Thank you karl.

PR #3415, opened 2024-11-05
De-register first from SCC or RMT/SMT

keichwa:
- https://github.com/SUSE/spacewalk/issues/23002
- https://bugzilla.suse.com/show_bug.cgi?id=1216946

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 - Added reminder note to unregister before registration to Client
   Configuration Guide (bsc#1216946)
+- Added admonition about podman related IP forwarding configuration to
+  Requirements in Installation and Upgrade Guide (bsc#1224318)
 - Updated Hub chapter in Large Deployments Guide (bsc#1215815)
 - Add registry.suse.com to the list of required URLs in the Network
   Requirements section of the Installation and Upgrade Guide
--- a/l10n-weblate/installation-and-upgrade/cs.po
+++ b/l10n-weblate/installation-and-upgrade/cs.po
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2024-11-04 07:55+0000\n"
+"POT-Creation-Date: 2024-11-06 16:17+0000\n"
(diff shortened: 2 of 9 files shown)


modules/client-configuration/pages/registration-overview-suse.adoc
```diff
@@ -19,6 +19,9 @@ You must also have created an activation key.
 For more information about creating activation keys, see xref:client-configuration:activation-keys.adoc[].
 
 
+If the client was registered to SCC, SMT, or RMT previously, remove such a registration with [command]``SUSEConnect`` before migrating the client to {productname} Server.
```
mcalmer, line 22: I think we should be more explicit and write down the full commands to de-register. Afaik 2 commands are needed. De-register and cleanup. We need to check.

mcalmer: https://www.suse.com/de-de/support/kb/doc/?id=000019054

keichwa: ok, makes sense.  I'll add the basic commands and add all link to the kb where the user can find all the details (cleanup and verification commands).  If you think we need all the details in our product documentation, I'd create a troubleshooting section in the Admin Guide.

modules/client-configuration/pages/registration-overview-suse.adoc
```diff
@@ -19,9 +19,36 @@ You must also have created an activation key.
...
+
+If the client was registered to SCC, SMT, or RMT previously, remove such a registration with [command]``SUSEConnect`` before migrating the client to {productname} Server.
+If you forget about removing the old registration first, it can happen that repositories not needed anymore getting automatically added to the client even after the client has been migrated to a newer version of a {suse} operating system using {productname}.
+
+* To de-register a {sle} 11 client, on the client, as root, execute:
++
+
+----
+suse_register -E
+----
```
mcalmer, line 43: I think we do not need SLES11. It is not supported in SUMA (officially) so we should not document anything about it

mcalmer (changes requested):
Just a bit too much added :-)

mcalmer (approved):

PR #3421, opened 2024-11-07
organization name is case sensitive

keichwa:
https://github.com/SUSE/spacewalk/issues/23579


modules/administration/pages/iss_v2.adoc
```diff
@@ -72,4 +72,5 @@ Properties file location can be overridden with parameter ``--serverConfig``.
 == Known Limitations
 
 * Source and target servers need to be on the same version.
-* Export and import organization should have the same name.
+* Export and import organization should have exactly the same name.
+  The organization name is case sensitive.
```
0rnela, line 76:
```suggestion
  * The organization name is case sensitive.
```

0rnela: I'd make the case sensitive requirement more pronounced by turning into a bullet-point (no need to the admonition though).

0rnela (changes requested):

rjpmestre (approved):

modules/administration/pages/iss_v2.adoc
keichwa: I think this is not needed.

PR #3426, opened 2024-11-11
Remove quickstart jcayouette

jcayouette:
Removes quickstart from both Uyuni and SUMA by commenting out the books in the params file. We can adjust the code for final removal in the future.


0rnela (approved):

0rnela (changes requested):
@jcayouette -- as Karl said plus please keep in mind the previous PR for QS removal.

PR #3428, opened 2024-11-11
admin guide: use mgrctl on container host

keichwa:
https://github.com/SUSE/spacewalk/issues/25390

### DIFF
--- a/modules/administration/pages/troubleshooting/tshoot-corruptrepo.adoc
+++ b/modules/administration/pages/troubleshooting/tshoot-corruptrepo.adoc
@@ -27,7 +27,7 @@ With an new repository data file, updates should operate as expected.
 .Procedure: Resolving Corrupt Repository Data
 . Remove all files from [path]``/var/cache/rhn/repodata/<channel-label>``.
     If you do not know the channel label, you can find it in the {productname} {webui}, by navigating to menu:Software[Channels > Channel Label].
-. On the container host, execute from the command line to regenerate the file in the container:
+. On the container host, from the command line, execute the following command to regenerate the file in the container:
 +
 ----
 mgrctl exec -ti -- spacecmd softwarechannel_regenerateyumcache <channel-label>


m-czernek (approved):
+1, this looks good to me.

Btw I wouldn't change anything here, but it's interesting that we say `on the container host, regenerate ...`.

In reality, `mgrctl exec` opens a terminal session and executes the command (in this case, `spacecmd ...`) inside of the container, so we actually regenerate the file in the container.

You're probably aware of the above, and I don't think we need to change anything here (probably too low-level information for the docs). But perhaps elsewhere, this distinction might be useful.

modules/administration/pages/troubleshooting/tshoot-corruptrepo.adoc
```diff
@@ -27,8 +27,8 @@ With an new repository data file, updates should operate as expected.
 .Procedure: Resolving Corrupt Repository Data
 . Remove all files from [path]``/var/cache/rhn/repodata/<channel-label>``.
     If you do not know the channel label, you can find it in the {productname} {webui}, by navigating to menu:Software[Channels > Channel Label].
-. Regenerate the file from the command line:
+. On the container host, regenerate the file from the command line:
```
keichwa, line 30: Yes, that's misleading. Trying to find a short but better phrase:
```suggestion
. On the container host, execute from the command line to regenerate the file in the container:
```

m-czernek: A quick note; you're missing a noun in your suggestion. Execute what?

I'd suggest something like: `... execute the following command from the command line to ...`

Or alternatively, `On the container host, regenerate the file in the container from the command line:`

(Sorry, a passer-by comment :) )

modules/administration/pages/troubleshooting/tshoot-corruptrepo.adoc
```diff
@@ -27,8 +27,8 @@ With an new repository data file, updates should operate as expected.
 .Procedure: Resolving Corrupt Repository Data
 . Remove all files from [path]``/var/cache/rhn/repodata/<channel-label>``.
     If you do not know the channel label, you can find it in the {productname} {webui}, by navigating to menu:Software[Channels > Channel Label].
-. Regenerate the file from the command line:
+. On the container host, execute from the command line to regenerate the file in the container:
```
keichwa, line 30:
```suggestion
. On the container host, from the command line, execute the following command to regenerate the file in the container:
```

PR #3451, opened 2024-11-18
Fix Server/Proxy repository enablement instructions

juliogonzalez:
# Description

- The URLs should be added with `$(arch)` so it works for the architecture of the host. This was the case in the past, but seems it got lost when the instructions were rewritten for containers.
- We also need to add the repo name, without it `zypper ar` fails with:
  ```
  transactional update # zypper ar https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Stable/images/repo/Uyuni-Server-POOL-x86_64-Media1/
  If only one argument is used, it must be a URI pointing to a .repo file.
  addrepo (ar) [OPTIONS] <URI> <ALIAS>
  addrepo (ar) [OPTIONS] <FILE.repo>
  ```
- A couple small extra fixes: quickstart and nonquistart were not consistante: `uyuni-storage-setup-server` added to quickstart, but `netavark` removed from non-quickstart (gets installed automatically)

**NOTE:** I understand soon we'll tell users to install the product again, not individual packages, but until that's done, let's add the storage package and let's remove netavark.

# Target branches

* Which product version this PR applies to: **Uyuni**
* Does this PR need to be backported? **NO**

# Links
Related: https://github.com/uyuni-project/uyuni-project.github.io/pull/144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-- Update OpenSUSE Leap micro 5.5 download link
+- Fixed Uyuni repository link and removed netavark from the installation
+  command
+- Updated OpenSUSE Leap micro 5.5 download link
 - Added documentation on deploying SUSE Manager Proxy in Public
   Cloud in Large Deployment Guide
 - Added admonition about disabling data synchronization with SCC in


0rnela (approved):

juliogonzalez:
Do not merge yet. The doc is still wrong, as it tells to use zypper to install packages on transactional systems.

juliogonzalez:
> Do not merge yet. The doc is still wrong, as it tells to use zypper to install packages on transactional systems.

Ignore this comment. We use zypper because we're inside a snapshot already. All good.

keichwa (approved):

PR #3482, opened 2024-11-22
Adding a hint on reducing downtime for migration with the usage of the parameter --prepare

Martin-Weiss:
# Description

Adding a hint for using --prepare for the migration to leave the source up and running and use initial / delta / final migration approach.

# Target branches

Uyuni and SUMA 5.x

### DIFF
--- a/modules/installation-and-upgrade/pages/container-deployment/uyuni/migrate-uyuni-to-a-container.adoc
+++ b/modules/installation-and-upgrade/pages/container-deployment/uyuni/migrate-uyuni-to-a-container.adoc
@@ -42,7 +42,10 @@ After the migration, it will be necessary to manually update the DHCP and DNS re
 
 [IMPORTANT]
 ====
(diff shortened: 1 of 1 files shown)


modules/installation-and-upgrade/pages/container-deployment/suma/server-migration-suma.adoc
```diff
@@ -60,6 +59,11 @@ After the migration, it will be necessary to manually update the DHCP and DNS re
 
 === Initial Preparation on the Old 4.3 Server
 
+[IMPORTANT]
+====
+The migration can take a very long time depending on the amount of data that needs to be replicated. To reduce downtime it is possible to run the migration multiple times in a process of "initial replication", "re-replication", "final replication and switch over" while all the services on the source can stay up and running and only during the final migration the processes on the source need to be stopped. For all "non-final" replications add the parameter [command]``--prepare`` needs to be added to to prevent the automatic stopping the services on the source server. I.e. use [command]``mgradm migrate podman <oldserver.fqdn> --prepare`` to prevent the automatic stopping the services on the source.
```
cbosdo, line 64:
```suggestion
The migration can take a very long time depending on the amount of data that needs to be replicated. To reduce downtime it is possible to run the migration multiple times in a process of "initial replication", "re-replication", "final replication and switch over" while all the services on the source can stay up and running and only during the final migration the processes on the source need to be stopped. For all "non-final" replications the parameter [command]``--prepare`` needs to be added to to prevent the automatic stopping the services on the source server. I.e. use [command]``mgradm migrate podman <oldserver.fqdn> --prepare`` to prevent the automatic stopping the services on the source.
```

cbosdo (commented):
Suggested a fix for one of the sentences, but LGTM after it

keichwa (approved):

0rnela (approved):

0rnela:
Thank you @Martin-Weiss.

PR #3485, opened 2024-11-25
5.0 - Admonition formatted in line with documentation  guidelines.

0rnela:
Minor follow-up work after mergeing https://github.com/uyuni-project/uyuni-docs/pull/3482.

# Target branches

- master / Uyuni https://github.com/uyuni-project/uyuni-docs/pull/3483
- 5.0


keichwa (approved):
I think it would be good if we'd replace "source" or "source server" with "old server"/"old 4.3 server".

0rnela:
> I think it would be good if we'd replace "source" or "source server" with "old server"/"old 4.3 server".


Thanks @keichwa, it makes sense. I'll do it.

PR #3493, opened 2024-11-27
Shim power s390x #23424

keichwa:
https://github.com/SUSE/spacewalk/issues/23424


0rnela (approved):

aaannz (approved):
I am not sure I would understand what is meant by this, but I hope this is just temporary and we'll make shim available also for ppc and s390 archs.

PR #3532, opened 2024-12-02
Replacing product version with the parameter  (part 1)

0rnela:
# Description

As part of the preparations for the product name change from SUSE Manager to SUSE Multi-Linux Manager, the use of parameter {productnumber} has been extended to all the instances in the documentation where we still had plain text instead.

This is Part 1 of the effort: Administration Guide
 
Some minor regular clean-up and normalization is done along the way too.

# Target branches

Parametarization will be implemented for both master and manager-5.0 branches, but the product name change will only happen for 5.1 branch (i.e. master) 

- master 
- 5.0 

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/25782

### DIFF
--- a/modules/administration/pages/content-lifecycle-examples.adoc
+++ b/modules/administration/pages/content-lifecycle-examples.adoc
@@ -228,7 +228,7 @@ AppStream repositories are natively supported throughout the {webui}.
 endif::[]
 
 ifeval::[{suma-content} == true]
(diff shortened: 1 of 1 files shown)


0rnela:
There are no instances where 5.0 can be replaced with parameter productnumber.

PR #3539, opened 2024-12-09
4.3: check DB password

keichwa:
https://bugzilla.suse.com/show_bug.cgi?id=1232836
https://github.com/SUSE/spacewalk/issues/25877


meaksh (commented):
I'm not really sure that it really makes sense to add this note to our docs.

The documentation says the user must copy these mentioned directories and files as part of the backup and restore procedure. So the copying is expected this is done, and thefore the `report_db_password` value should be also copied.

I think we should not add a note to our documentation to cover a case when the user is not following the procedure the documentation describes, unless there was a special procedure in our docs for this particular case, but I don't see it.

keichwa:
Thanks for feedack!
I'll ask the reporter what we can do to encourage users to follow our procedures more careful.

PR #3554, opened 2024-12-16
Replace mgradm cp w/ mgrctl cp in migration docs

netzvieh:
# Description

mgradm does not have a cp subcommand. Should be mgrctl instead.

# Target branches

* Which product version this PR applies to: Uyuni, SUMA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+- Replaced mgradm with mrgctl in Installation and Upgrade Guide
 - Corrected metadata signing section in Administration Guide
 - Changes proxy helm installation to use package from OS channel in Installation and
   Upgrade Guide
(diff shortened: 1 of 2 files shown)


keichwa (commented):
Thank you for spotting and fixing this issue!

One more or less just cosmetic issue: There is no need to fix the translation files (.po and .pot), even in this case it looks straight forward.  But these files will be changed externally (via weblate) and will thus be overriden by the translators.  Can you revert these changes?
 
Then it would be nice if you could add a line to the toplevel `CHANGELOG.md` file (just `Replaced mgradm with mrgctl in Installation and Upgrade Guide`, or something similar).

netzvieh:
I didn't want to leave them out, since they referenced this line of code and I'm not familiar with weblate. I removed the changes from the commit and also added your proposed line to the changelog.

keichwa (approved):
Thanks!  Now it is perfect for our purposes ;)  I think it isn't sufficiently documented how we deal with translation files in this repository. Thanks for taking the time to fix this issue and to update the PR!  I'll merge it later today or tomorrow.  Maybe, a co-editor wants to take a look, too.

keichwa:
@netzvieh do you want to backport (5.0) it yourself?  If not, I can do it for you and let you review it.

keichwa:
> @netzvieh do you want to backport (5.0) it yourself? If not, I can do it for you and let you review it.

Please ignore it!  We do not need this in the 5.0 branch because this is relevant for uyuni only.

PR #3556, opened 2024-12-16
more info on software channels and client tools

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/25767 (**updated!**).

The other clients will be check as part of https://github.com/SUSE/spacewalk/issues/26069.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 - Documented client tools details for Uyuni client registration in
   Client Configuration Guide
+- Corrected server SSL self-signed certificates renewal procedure
+  in Administration Guide (bsc#1235696)
 - Updated external Link in Client Configuration Guide (bsc#1235825)
 - Added admonition that NFS does not support SELinux labeling and should
-  not be used
+  not be used in Installation and Upgrade Guide
 - Updated product migration in Client Configuration Guide: added SUSE
   Linux Enterprise Server to SUSE Linux Enterprise Server for SAP
   Applications and extensions enabled automatically
--- a/l10n-weblate/administration/administration.pot
+++ b/l10n-weblate/administration/administration.pot
@@ -7,7 +7,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2025-01-20 11:44+0000\n"
+"POT-Creation-Date: 2025-01-22 09:13+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -15828,7 +15828,7 @@ msgstr ""
 
 #. type: Plain text
 #: modules/administration/pages/ssl-certs-selfsigned.adoc:10
-#: modules/administration/pages/ssl-certs-selfsigned.adoc:21
+#: modules/administration/pages/ssl-certs-selfsigned.adoc:22
 msgid ""
 "Ensure that the [systemitem]``set-cname`` parameter is the fully qualified "
 "domain name of your {productname} Server.  You can use the the "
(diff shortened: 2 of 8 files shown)


m-czernek:
IMO the whole doc starts making sense only when you finish reading the whole page. As a person that's trying to solve a particular issue, I would like a summary of the process before we delve into implementation details of the procedure.

Funnily enough, in [0], we provide a summary about what is SLEM, which to me seems unnecessary. I would welcome something like:

```
To register a new client to Uyuni, first synchronize software channels for the new client's operating system and architecture by using either Web UI or CLI.
# Then, for Uyuni only
Then, synchronize the client tool repository for the newly added operating system by using CLI.
```

Minor point: maybe we want to also show how to list all client tool repositories? Users can execute `spacewalk-common-channels -l` to list all available repositories.

[0] https://w3.suse.de/~ke/uyuni/uyuni-clienttools-3556/uyuni/client-configuration/clients-sle.html

modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -3,27 +3,39 @@
...
+
+* Synchronize software channels for the new client's operating system and architecture by using either {webui} or command line interface.
+
+ifeval::[{uyuni-content} == true]
+
+* Synchronize the client tool repository for the newly added operating systems by using command line interface.
+
+endif::[]
+
+* Run the actual registration.
```
m-czernek, line 25: We have multiple ways to register the client, how about:

```suggestion
* Register the client with {productname} Server by using the {webui} or the command line interface.
```

Note: I see you use `command line interface` throughout the text, but I wonder if it should be a `command-line interface`. The way I previously used this, `command-line` is a compound adjective that modifies the word `interface` (which is why it requires a hyphen).

In my suggestion above, I use `command line interface` for consistency's sake.

modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -3,27 +3,39 @@
...
+
+* Additional sections to check and verify the registration procedure.
```
m-czernek, line 27: This feels like it's missing a verb.

Did you want to write maybe something like:

```suggestion
Additional sections describe ways to check and verify the registration procedure.
```

modules/client-configuration/pages/clients-slemicro.adoc
```diff
@@ -164,6 +176,11 @@ include::snippets/addchannels_vendor_cli.adoc[]
 
 ifeval::[{uyuni-content} == true]
 
+== Client Tools
+
+Without client tools onboarding clients will fail.
+Client tools you can only add from the command line.
```
m-czernek, line 182: I'm not sure about this particular sentence. The wording is wrong (it should be `You can add client tools only from the command line`), but IMHO it's unnecessary here? I might remove this sentence.

When we can do one thing multiple times, then I'd recommend adding a preface saying `You can do X by using Y, Z, and Alpha`. The reason for that is that a customer might see multiple ways to do something, but they might think that they need to do it multiple times.

If we only have one procedure type, I think it's fine not to preface the methods for doing it.

Just my 2c.

m-czernek (approved):
Thank you for the changes, this looks better! Adding a few comments :) 

I think there's still room for improvement regarding the way the guide looks in general, i.e. improving the flow such that a reader can scan the page and know the general steps, but that'd be a larger refactor, not really reasonable for a PR review like this.

So +1 from me; I'd love to find time at some point to create a concept PR that'll simplify these pages.

modules/client-configuration/pages/clients-slemicro.adoc
keichwa: Thanks!  I think I now understood the thing with the dash!  Actually, it is also in Daria's wonderful https://documentation.suse.com/style/current/single-html/docu_styleguide/#sec-terminology . The Style Guide I must read more often!

modules/client-configuration/pages/clients-slemicro.adoc
keichwa: This makes sense.  The next generation could even think about separating webui and CLI into two separate guides.

keichwa: separate issue created

PR #3562, opened 2024-12-18
For 4.3: ssl certs cleanup #1216968 keichwa

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/23003

https://bugzilla.suse.com/show_bug.cgi?id=1216968

### DIFF
--- a/modules/administration/pages/ssl-certs-imported.adoc
+++ b/modules/administration/pages/ssl-certs-imported.adoc
@@ -128,13 +128,16 @@ spacewalk-service start
 
 === On the Proxy
 
-If you are using a proxy, you need to generate a server certificate RPM for each proxy, using their hostnames and cnames.
-You should use [command]``mgr-ssl-cert-setup`` also on a {productname} Proxy to replace the certificates.
-Because the {productname} Proxy does not have a postgreSQL database, just the dedicated proxy restrart command is sufficient:
+
+If you are using a proxy, you need to generate a server certificate for the proxy on the {productname} server and copy the files to the proxy.
+Alternatively, you let your own Certificate Authority generate these files and provide them on the proxy.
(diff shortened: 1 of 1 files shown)


modules/administration/pages/ssl-certs-imported.adoc
```diff
@@ -117,13 +124,23 @@ systemctl restart postgresql.service
 spacewalk-service start
 ----
 
-If you are using a proxy, you need to generate a server certificate RPM for each proxy, using their host names and cnames.
+
+
+=== On the Proxy
+
+If you are using a proxy, you need to generate a server certificate RPM for each proxy, using their hostnames and cnames.
 You should use [command]``mgr-ssl-cert-setup`` also on a {productname} Proxy to replace the certificates.
```
mcalmer, line 132: You should not generate the "RPM". You just need to get the server certificate and key for the proxy.
If the customer generated it on the server, he should do the same for the proxy. If he get them from his own CA, he just need to make the files available on the proxy. After that, they should use `mgr-ssl-cert-setup` similar as on the server as described above.

mcalmer: Maybe something like:
If you are using a proxy, you need to generate a server certificate for the proxy on the server and copy the files to the proxy. Alternative you let your own Certificate Authority generate these files and provide them on the proxy. 
After the files are available, use `mgr-ssl-cert-setup` .....

mcalmer (changes requested):

modules/administration/pages/ssl-certs-imported.adoc
keichwa: ok, now understood, I think. Using your proposal.

mcalmer (approved):

PR #3564, opened 2024-12-19
CLM filters for kernel package

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/26090.

https://bugzilla.suse.com/show_bug.cgi?id=1234202

### DIFF
--- a/modules/administration/pages/content-lifecycle-examples.adoc
+++ b/modules/administration/pages/content-lifecycle-examples.adoc
@@ -177,18 +177,21 @@ You need to exclude all patches which contain a higher version of [package]``ker
 . Click btn:[Attach/Detach Filters].
 . Select [guimenu]``Exclude kernel greater than 4.12.14-150.17.1``, and click btn:[Save].
 
+You need to repeat this procedure for the [package]``kernel-default-base`` package.
+
 
 When you click btn:[Build], a new environment is created.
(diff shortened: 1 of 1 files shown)


cbbayburt:
Thanks for looking into it, Karl.

Actually, the only change is needed in the `Procedure: Excluding Packages with a Higher Kernel Version` part in `content-lifecycle-examples.adoc`.

This procedure shows how to exclude `kernel-default` from the channels. As described in the bug, the same procedure must be done for `kernel-default-base` as well. So the procedure must be repeated for `kernel-default-base`.

`content-lifecycle.adoc` describes an alternative procedure and is not affected by this, so no change is needed there.

cbbayburt (changes requested):
Thanks for looking into it, Karl.

Actually, the only change is needed in the `Procedure: Excluding Packages with a Higher Kernel Version` part in `content-lifecycle-examples.adoc`.

This procedure shows how to exclude `kernel-default` from the channels. As described in the bug, the same procedure must be done for `kernel-default-base` as well. So the procedure must be repeated for `kernel-default-base`.

`content-lifecycle.adoc` describes an alternative procedure and is not affected by this, so no change is needed there.

cbbayburt (approved):
Thanks for the update Karl. Looks good to me 👍

PR #3571, opened 2025-01-02
Removed spacewalk-koan from the list of required packages

keichwa:
https://github.com/SUSE/spacewalk/issues/26118
https://github.com/uyuni-project/uyuni/issues/8905

### DIFF
--- a/modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
+++ b/modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
@@ -1,11 +1,12 @@
 [[troubleshooting-autoinstallation]]
 = Troubleshooting Autoinstallation
 
-If you use the obsolete plain {salt} (salt-minion) implementation for communication, you on your own must make sure that all the needed software (software packages) is properly available in configured client channels.
+If you use the plain {salt} (salt-minion) implementation for communication, you on your own must make sure that all the needed software (software packages) is properly available in configured client channels.
+It is not recommended to use this implementation together with {productname}.
 
 With the default {salt} bundle (venv-salt-minion) implementation you only need a client tools channel as a child channel that is compatible with your client base channel.
 All required packages will be part of the {salt} bundle.
 
 * Check that the client tools software channel related to the base channel in your autoinstallation profile is available to your organization and your user.
 * Check that the tools channel is available to your {productname} as a child channel.
-* Only if using the obsolete plain {salt} (salt-minion) implementation, also check that the required packages and any dependencies are available in the associated channels.
+* Only if using the plain {salt} (salt-minion) implementation, also check that the required packages and any dependencies are available in the associated channels.


modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
```diff
@@ -8,7 +8,6 @@ For autoinstallation to work, these packages are required:
 * [package]``pyOpenSSL``
 * [package]``rhnlib``
 * [package]``libxml2-python``
-* [package]``spacewalk-koan``
```
mcalmer, line 11: I think in 5.0 none of these packages are required. 
I think we only need venv-salt-minion and its dependencies. 
Maybe we should just point to the client tools channel to be available

mcalmer (changes requested):

modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
keichwa: I think you proposal makes sense.  I tried to fix the focus of the TS section.  I hope my phrases are understandable.  Can you please check again?

modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
```diff
@@ -1,17 +1,11 @@
...
-* [package]``pyOpenSSL``
-* [package]``rhnlib``
-* [package]``libxml2-python``
-* [package]``spacewalk-koan``
-
-To resolve this issue, check these things first:
-
-* Check that the tools software channel related to the base channel in your autoinstallation profile is available to your organization and your user.
+* Check that the client tools software channel related to the base channel in your autoinstallation profile is available to your organization and your user.
 * Check that the tools channel is available to your {productname} as a child channel.
```
mcalmer, line 10: Are these 2 maybe the same?

modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
```diff
@@ -1,17 +1,11 @@
 [[troubleshooting-autoinstallation]]
 = Troubleshooting Autoinstallation
 
-Depending on your base channel, new autoinstallation profiles might be subscribed to a channel that is missing required packages.
+If you use the obsolete plain {salt} (salt-minion) implementation for communication, you on your own must make sure that all the needed software (software packages) is properly available in configured client channels.
```
mcalmer, line 4: I am not sure about the work "obsolete" . Only together with SUSE Manager we do not use this package anymore. But for SLES it is still a regular supported package. But I am unsure how to phrase this correctly

modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
keichwa: yes, I think it is the same.  I think with the first item we want to say the organization and the user must have permissions to use it and with the second that the Manager must have it available/ready to be used.  The second might be superfluous or, maybe, it should come first.

modules/administration/pages/troubleshooting/tshoot-autoinstallation.adoc
keichwa: ok, thanks for bg info.  I tried to change it that way.

mcalmer (approved):

PR #3595, opened 2025-01-13
clean up and make install docs on proxy consistent

keichwa:
**Find link to rendered output** in https://github.com/SUSE/spacewalk/issues/26181

replace outdated and obsolete content with a stub to provide an entry point for links.

https://bugzilla.suse.com/show_bug.cgi?id=1235567
https://github.com/SUSE/spacewalk/issues/26181

### DIFF
--- a/modules/installation-and-upgrade/pages/container-management/persistent-container-volumes.adoc
+++ b/modules/installation-and-upgrade/pages/container-management/persistent-container-volumes.adoc
@@ -165,9 +165,9 @@ The following  volumes are stored under the **Podman** default storage location 
 |Volume Name | Volume Directory
 
 | **uyuni-proxy-rhn-cache**
-| [path]``/var/???/???``
+| [path]``/var/cache/rhn``
 
 
 | **uyuni-proxy-squid-cache**
-| [path]``/var/???/???``
+| [path]``/var/cache/squid``
 |===


0rnela (approved):

modules/installation-and-upgrade/pages/container-deployment/suma/proxy-deployment-suma.adoc
```diff
@@ -82,72 +81,76 @@ This table shows the hardware requirements for deploying {productname} Proxy.
 
 |===
 
+// FIXME: ^^^ should we better just mention
+//            /var/lib/containers/storage/volumes here above? ^^^
```
aaannz, line 85: Paths above are wrong. I though we already fixed those.

We should not mention individual volumes because in fact only one volume is used, the rest are basically empty or for config files.
But more importantly `mgr-storage-proxy` works only over whole volume directory.

So please remove any mentioned of `[path]``/var/lib/containers/storage/volumes/srv-www`` ` and `[path]``/var/lib/containers/storage/volumes/var-cache`` ` and replace it with only one path `[path]``/var/lib/containers/storage/volumes`` ` which should be 100GB minimum.

modules/installation-and-upgrade/pages/container-deployment/transfer_proxy_config.adoc
```diff
@@ -5,19 +5,20 @@ This archive needs to be made available on the proxy container host.
 
 
 .Procedure: Copying the Proxy Configuration
-. Copy the files from the Server container to the Server host OS:
+
+. If notalready done, copy the configuration archive ([literal]``config.tar.gz``) generated in the previous step from the server container to the server host:
```
aaannz, line 9:
```suggestion
. If not already done, copy the configuration archive ([literal]``config.tar.gz``) generated in the previous step from the server container to the server host:
```

aaannz (commented):
Just two points for now.

However there are additional new dialogs for 5.1 images and also with 5.1 we now have encrypted images as well.

This should be covered as well once Alpha2 images are released.

modules/installation-and-upgrade/pages/container-deployment/suma/proxy-deployment-suma.adoc
keichwa: Yes, but I missed that here.  I only fixed the requirements in `installation-and-upgrade/pages/hardware-requirements.adoc`.  I unfortunately do not know where we book-people to have the requirements documented...

If contents in `installation-and-upgrade/pages/hardware-requirements.adoc` is also good enough for the vm section for proxy deployments, I'd just reference that reqs file and remove it from the deployment files.

keichwa: BTW, I created a new section for proxy in `persistent-container-volumes.adoc`.  Does it make sense?  If yes, can you please replace `???` with actual directories from inside the container?

PR #3608, opened 2025-01-15
Add hint not to use NFS

stdevel:
# Description

I've stumbled upon that NFS doesn't support SELinux file labeling and therefore can't be used for container volume mounts. There is already a hint regarding PostgreSQL and NFS - but it also can't be used for the packages (`/var/spacewalk`).

This PR adds appropriate hints for the SUMA and Uyuni installation files.

# Target branches

- master
- 5.0

# Links
-


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Added hint that NFS doesn't support SELinux labeling and shouldn't be used
```
```suggestion
- Added hint that NFS does not support SELinux labeling and should not be used
```

modules/installation-and-upgrade/pages/hardware-requirements.adoc
```diff
@@ -253,6 +253,11 @@ We recommend you use XFS as the filesystem type for all volumes.
 The size of the disk for repositories storage is dependent on the number of distributions and channels you intend to manage with {productname}.
 See the tables in this section for guides to estimate the size required.
 
+[IMPORTANT]
+====
+Do not use NFS for storage as it doesn't support SELinux file labeling.
```
keichwa, line 258:
```suggestion
Do not use NFS for storage because it does not support SELinux file labeling.
```

modules/installation-and-upgrade/pages/uyuni-install-requirements.adoc
```diff
@@ -6,6 +6,10 @@ endif::[]
 
 The following tables specify the minimum server and proxy requirements.
 
+[IMPORTANT]
+====
+Do not use NFS for storage as it doesn't support SELinux file labeling.
```
keichwa, line 11:
```suggestion
Do not use NFS for storage because it does not support SELinux file labeling.
```

keichwa (changes requested):

cbosdo (approved):
This may be mitigated. I'm currently looking at a case  with NFS. It seems to work with `/var/spacewalk` on NFS, but you need to setup `security_label` option on the export... which requires an selinux-enabled NFS server and NFS 4.2+. It seems that there are still drawbacks for that... so I'm fine recommending not to use NFS at all.

stdevel:
Nice to hear that you're having a look at it. I had the same situation before creating this PR. Unfortunately a customer was using a NFS applicance that has not SELinux-capable NFS server. But having some more technical notes (maybe on a dedicated page?) about this is definitely very useful.

keichwa (approved):

cbosdo:
> Nice to hear that you're having a look at it. I had the same situation before creating this PR. Unfortunately a customer was using a NFS applicance that has not SELinux-capable NFS server. But having some more technical notes (maybe on a dedicated page?) about this is definitely very useful.

I'm still far from having a clear view on all that. I still have side effects in some cases.

PR #3611, opened 2025-01-15
Removed images (Admin Guide, pt.2) 

0rnela:
As part of documentation cleanup for better maintenance, images are identified as candidates for removal when their content and purpose are already well described in the corresponding procedures.

This PR is for images in the Confidential Computing document.  

5.1 (master) branch only.


keichwa (approved):
You also changed a link to an external resource. I assume this happened intentionally.

PR #3648, opened 2025-01-29
Backup and Restore chapter regarding containerization

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/25047


0rnela (approved):

modules/administration/pages/backup-restore.adoc
```diff
@@ -13,6 +13,10 @@ Regardless of the backup method you use, you must have available at least three
 Running out of space can result in backups failing, so check this often.
 ====
 
+The following section describes backing up data within the {productname} container.
+On the container host, use [literal]``mgrctl term`` to open a terminal in the container before you execute commands.
```
rjmateus, line 17: I think not all the commands are to be runned inside the container. So we should be careful with this sentence in here.

modules/administration/pages/backup-restore.adoc
```diff
@@ -415,8 +443,10 @@ public.log                            | 32 kB
 It is possible to move the database to another location.
 For example, if database storage space is running low.
 
-
-
+////
```
rjmateus, line 446: I don't think we should keep this procedure anymore. It can even cause problems with SELinux label in the way it's described in here.

modules/administration/pages/backup-restore.adoc
```diff
@@ -473,8 +503,12 @@ rcpostgresql start
 ----
 spacewalk-service start
 ----
+////
 
 
+////
+// FIXME: does this still make sense?
```
rjmateus, line 510: this section needs to be dropped or re-written. What we have in here is not true anymore.

rjmateus (commented):
some comments on this one. Thank you for looking for it Karl

modules/administration/pages/backup-restore.adoc
keichwa: Yes, you are right., we'd better remove that statement here.  All the procedures in the following sections are now adjusted as needed.

PR #3651, opened 2025-01-29
Admonition added to warn about using latest MU

0rnela:
# Description

To set expectations if a customer report a bug in an old MU version, as in such case, we request customers to update to the latest MU version of that product.

To address that, the existing copy was reshuffled and tidied up, and the new copy added to the existing admonition.

# Target branches

- master
- 5.0 https://github.com/uyuni-project/uyuni-docs/pull/3652
- 4.3 https://github.com/uyuni-project/uyuni-docs/pull/3653


keichwa (approved):

keichwa (changes requested):
I revoked my approval.  The exact statement is to be clarified with the product owner.  The statement must fit for all clients.  The most important thing (IMO) is to say that the

> client is supported by the organization that supplies the operating system

0rnela:
@admd can you please confirm if the text in the admonition can be merged? The rest of the changes in this PR are just simplification and cleanup of the already existing text.

keichwa (approved):
wording is ok for me.  but sooner or later we must clarify what "support" means where it occurs.  Also see my proposal: https://github.com/SUSE/spacewalk/issues/18593 --but I cannot do this.

PR #3680, opened 2025-02-05
Changed Uyuni base OS version download address

0rnela:
# Description

We stated 6.1 as the base OS, but linked to 5.5.

Master/Uyuni only.


keichwa (commented):
I'd leave out the version number from the URL, and just specify:

https://download.opensuse.org/distribution/leap-micro/

PR #3690, opened 2025-02-10
update commands to allow renaming a containerized server

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/25390

PR #3707, opened 2025-02-17
Backport 5.0: update commands to allow renaming a containerized server (#3690)

keichwa:
* update commands to allow renaming a containerized server
*  https://github.com/SUSE/spacewalk/issues/25390
* differentiate between NC 4.3 proxy and containerized one
* Update modules/administration/pages/troubleshooting/tshoot-hostname-rename.adoc
Co-authored-by: Ondřej Holeček <oholecek@suse.com>
Co-authored-by: Ornela Marić <omaric@suse.com>

### DIFF
--- a/modules/administration/pages/troubleshooting/tshoot-hostname-rename.adoc
+++ b/modules/administration/pages/troubleshooting/tshoot-hostname-rename.adoc
@@ -38,6 +38,10 @@ This is because the changes have not been made in the database, which prevents t
 If you need to change the hostname of the {productname} Server, you can do so using the [command]``spacewalk-hostname-rename`` script.
 This script updates the settings in the {postgresql} database and the internal structures of {productname}.
 
+
+
+=== Server Configuration
+
 The [command]``spacewalk-hostname-rename`` script is part of the [package]``spacewalk-utils`` package.
 
 The only mandatory parameter for the script is the newly configured IP address of the {productname} Server.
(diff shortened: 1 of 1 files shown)


modules/administration/pages/troubleshooting/tshoot-hostname-rename.adoc
```diff
@@ -45,45 +49,49 @@ The only mandatory parameter for the script is the newly configured IP address o
...
 ----
 
-+
-Or  the [systemitem]``salt-minion``  service (if not using the {salt} bundle):
+. Restart the server.
+  On the container host, execute the restart command:
 +
 ----
-systemctl restart salt-minion
+mgradm restart
```
aaannz, line 73: I did not notice this on the original PR, but this does not make sense.
Why in the one procedure we are mixing SMLM server restart and changing salt config of salt clients?
I think the restart step was not a restart step of the SMLM server, but salt client, so the `systemctl restart venv-salt-minion` should be there instead of `mgradm restart`.

I have no experience renaming the server, but this is confusing.

At least there should be two procedures -> one for changes on the SMLM server and another what to do with directly connected salt clients.

keichwa: I ported fixes from master

aaannz (approved):

PR #3739, opened 2025-02-26
5.0 - New admonition for outdated client

0rnela:
# Description

This is an extension of a [previous PR](https://github.com/uyuni-project/uyuni-docs/pull/3652).

We are informing the users that only the latest minor release of each supported major version is fully supported.

The text has been adapted from the release notes.

Backport targets (edit as needed):

- master https://github.com/uyuni-project/uyuni-docs/pull/3738
- 5.0
- 4.3 https://github.com/uyuni-project/uyuni-docs/pull/3740

Related issue: https://github.com/SUSE/spacewalk/issues/25503

PR #3738, opened 2025-02-26
New admonition for outdated clients

0rnela:
# Description

This is an extension of a [previous PR](https://github.com/uyuni-project/uyuni-docs/pull/3651).

We are informing the users that only the latest minor release of each supported major version is fully supported.

The text has been adapted from the release notes.

Backport targets (edit as needed):

- master
- 5.0 https://github.com/uyuni-project/uyuni-docs/pull/3739
- 4.3 https://github.com/uyuni-project/uyuni-docs/pull/3740

Related issue: https://github.com/SUSE/spacewalk/issues/25503

PR #3740, opened 2025-02-26
4.3 - New admonition for outdated clients

0rnela:
# Description

This is an extension of a [previous PR](https://github.com/uyuni-project/uyuni-docs/pull/3653).

We are informing the users that only the latest minor release of each supported major version is fully supported.

The text has been adapted from the release notes.

- master https://github.com/uyuni-project/uyuni-docs/pull/3738
- 5.0 https://github.com/uyuni-project/uyuni-docs/pull/3739
- 4.3

Related issue: https://github.com/SUSE/spacewalk/issues/25503

### DIFF
--- a/modules/client-configuration/pages/supported-features.adoc
+++ b/modules/client-configuration/pages/supported-features.adoc
@@ -23,6 +23,14 @@ Supported client operating systems are listed in this table. The icons in the ta
 * {cross} clients running this operating system are not supported by {suse}
 * {question} clients are under consideration, and may or may not be supported at a later date.
 
+[IMPORTANT]
+====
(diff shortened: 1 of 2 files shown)


jcayouette (changes requested):
I would update the text to something like this:

```
{productname} is compatible with multiple major versions of non-SUSE client operating systems. However, full support is provided only for the latest minor release of each supported major version.

While older minor releases may continue to function, support for them is limited to a reasonable-effort basis. If issues occur on an outdated minor release, users will be required to update to the latest minor version before receiving further support.
```

keichwa (approved):

PR #3750, opened 2025-03-04
adjust monitoring for container host/server container

keichwa:
https://github.com/SUSE/spacewalk/issues/24778

- master
- manager-5.1 https://github.com/uyuni-project/uyuni-docs/pull/4392


modules/administration/pages/monitoring.adoc
```diff
@@ -579,15 +605,72 @@ This section demonstrates how to generate client/server certificates for Prometh
 
 .Procedure: Creating server/client TLS certificate
 
-. On the {productname} Server, at the command prompt, run following command:
+. At the command prompt of the {productname} container host, as root, run the following commands:
+
++
```
srbarrios, line 610:
```suggestion
[source,shell]
----
mgrctl term
----
```

modules/administration/pages/monitoring.adoc
```diff
@@ -579,15 +605,72 @@ This section demonstrates how to generate client/server certificates for Prometh
...
+
+--
+
+.. To generate certificate files, run the following command.
+
++
+
+Ensure that the [systemitem]``set-cname`` parameter is the fully qualified domain name (FQDN) of your Salt client.
+You can use the the [systemitem]``set-cname`` parameter multiple times if you require multiple aliases:
```
srbarrios, line 619: Actually, I'm not fully sure, but I think that's just optional.
Better ask @witekest

modules/administration/pages/monitoring.adoc
```diff
@@ -579,15 +605,72 @@ This section demonstrates how to generate client/server certificates for Prometh
...
+
 +
+
 ----
-rhn-ssl-tool --gen-server --dir="/root/ssl-build" --set-country="COUNTRY" \
---set-state="STATE" --set-city="CITY" --set-org="ORGANIZATION" \
---set-org-unit="ORGANIZATION UNIT" --set-email="name@example.com" \
---set-hostname="minion.example.com" --set-cname="minion.example.com" --no-rpm
+mgrctl exec -ti -- mgr-ssl-tool --gen-server --dir="/root/ssl-build" --set-country="COUNTRY" \
```
srbarrios, line 624: Nice, so if you do that way, we don't really need to use `mgrctl term`

modules/administration/pages/monitoring.adoc
```diff
@@ -579,15 +605,72 @@ This section demonstrates how to generate client/server certificates for Prometh
...
+.. Copy [path]``server.crt`` and [path]``server.key`` files from the host to the monitoring client:
+
++
+
+----
+ssh minion.example.com 'mkdir /etc/ssl/mlm-server-certs'
+scp /root/server.* minion.example.com:/etc/ssl/mlm-server-certs
+ssh minion.example.com 'chmod go+r /etc/ssl/mlm-server-certs/server.*; ls -la /etc/ssl/mlm-server-certs'
+----
+
```
srbarrios, line 660: Before start with the configuration of formulas, I'm looking into what I wrote in my card, and I see this section:

----

From the minion:

```
suma-bv-50-min-sles15sp4:/etc/ssl # mkdir suma-server-certs
suma-bv-50-min-sles15sp4:/etc/ssl # cd suma-server-certs/
suma-bv-50-min-sles15sp4:/etc/ssl/suma-server-certs # cp /root/server.* .
suma-bv-50-min-sles15sp4:/etc/ssl/suma-server-certs # chmod go+r server.*
suma-bv-50-min-sles15sp4:/etc/ssl/suma-server-certs # ls -la
total 16
drwxr-xr-x 2 root root   42 jul  5 11:46 .
drwxr-xr-x 6 root root  183 jul  5 11:45 ..
-rw-r--r-- 1 root root 8335 jul  5 11:46 server.crt
-rw-r--r-- 1 root root 3268 jul  5 11:46 server.key
```

And we use that path on the minion when we pass the server certificate and key into the Prometheus Formula of the minion.

I would say that's missing right?

modules/administration/pages/monitoring.adoc
keichwa: I now only recommend `mgrctl term`  if the user must enter more commands in a row inside the container :^)

PR #3752, opened 2025-03-05
containerize fallout: run parted on host OS in Admin tshoot

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/26575 (updated).

* https://bugzilla.suse.com/show_bug.cgi?id=1237535
* https://github.com/SUSE/spacewalk/issues/26575


modules/administration/pages/troubleshooting/tshoot-container-full-disk.adoc
```diff
@@ -24,24 +24,31 @@ If more detailed instructions are required, put them in a "Resolving" procedure:
...
-* [literal]``(parted)`` [command]``resizepart NUMBER 100%`` where [literal]``NUMBER`` is the partition number shown by [command]``print`` command (for example, [literal]``1`` if [literal]``/dev/vdb1``)
-* [literal]``(parted)`` [command]``quit``
+. Ensure the partition uses all available space.
+  Run the following commands:
+
+--
+.. [command]``parted /dev/vdb``
+.. [literal]``(parted)`` [command]``print``
+.. [literal]``(parted)`` [command]``resizepart NUMBER 100%`` where [literal]``NUMBER`` is the partition number shown by [command]``print`` command (for example, [literal]``1`` if [literal]``/dev/vdb1``)
+.. [literal]``(parted)`` [command]``quit``
```
aaannz, line 41: This is again more complicated by our making 😄

Initial `mgr-storage-server` created partition so this procedure works.

However later it was modified (to support LVM) that it does not have any partition, so this step can be skipped.

So the procedure:
- is there a partition -> call parted and resize it.
- no partition, no need to call parted and continue next

modules/administration/pages/troubleshooting/tshoot-container-full-disk.adoc
```diff
@@ -24,24 +24,42 @@ If more detailed instructions are required, put them in a "Resolving" procedure:
...
+. Increase the disk size.
+  Actions to take depend on the installation scenario.
+
+. If there is a partition and
+
++
+
+** modified later on, skip this step.
+
+** it was initially created with [command]``mgr-storage-server`` or [command]``mgr-storage-proxy``, and not modified later on, ensure the partition uses all available space.
```
aaannz, line 40: I would not understand what this means.
How about:

```suggestion
. If the disk is partitioned (for example there is a [path]``/dev/vdb1`` for disk [path]``/dev/vdb`` run the following commands:
```

PR #3774, opened 2025-03-11
Removed Ubuntu 20.04 from the list of supported clients

keichwa:
**For link to rendered output,** see https://github.com/SUSE/spacewalk/issues/26604.

* https://github.com/SUSE/spacewalk/issues/26604
* https://bugzilla.suse.com/show_activity.cgi?id=1238481

### DIFF
--- a/modules/client-configuration/pages/clients-ubuntu.adoc
+++ b/modules/client-configuration/pages/clients-ubuntu.adoc
@@ -1,13 +1,13 @@
 [[clients-ubuntu]]
 = Registering {ubuntu} 22.04 and 24.04 Clients
 
-This section contains information about registering clients running {ubuntu} 22.04 LTS and 24.04 LTS operating systems.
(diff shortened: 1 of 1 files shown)


modules/client-configuration/pages/clients-ubuntu.adoc
```diff
@@ -1,20 +1,20 @@
 [[clients-ubuntu]]
-= Registering {ubuntu} 20.04, 22.04 and 24.04 Clients
+= Registering {ubuntu} 22.04 and 24.04 Clients
```
0rnela, line 2: I would remove client versions from the title as well.  We don't have them for any other OS. Master only, no need for back-porting if it is time consuming.

keichwa: ok, done. I kept it somewhere else, but can remove it if that's better.

PR #3830, opened 2025-03-26
Restructure Server Install scenario micro/sles

keichwa:
* https://bugzilla.suse.com/show_bug.cgi?id=1239801
* https://github.com/SUSE/spacewalk/issues/26739


admd:
@keichwa can we move this note https://github.com/uyuni-project/uyuni-docs/pull/3830/files#diff-39d653c70b8d9efdfe3ec4bbb910df7d35eb4e5edd9a78073249ff6927d4d1b0R395-L354 , under 3.1 section which is specific to SLE Micro

keichwa:
> @keichwa can we move this note https://github.com/uyuni-project/uyuni-docs/pull/3830/files#diff-39d653c70b8d9efdfe3ec4bbb910df7d35eb4e5edd9a78073249ff6927d4d1b0R395-L354 , under 3.1 section which is specific to SLE Micro

Thanks for spotting.  Done.  Also for 5.0.4.

modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-suma.adoc
```diff
@@ -213,7 +229,99 @@ transactional-update
...
+// FIXME: where is the following "Reboot" actually required?
+
+. Reboot
+. Log in as root and install podman
+
++
+
+[source,shell]
+----
+zypper install podman
```
nodeg, line 306:
```suggestion
zypper install podman mgradm mgradm-bash-completion
```
I think we are missing `mgradm` and `mgradm-bash-completion`, was we have for `mgrpxy` for the proxy. Otherwise, we cannot continue later with deploying SUMA/MLM with `mgradm`.
See https://github.com/uyuni-project/uyuni-docs/pull/3834/files#diff-183022748e3c5a3480fb8789a0c5b574382f2f7a051537ca50ab2ca2edbbfa4eR220

nodeg (changes requested):

modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-suma.adoc
nodeg: I did a test with `zypper se mgradm` and apparently those are already installed. So not sure if we then need them.

admd: We should maybe add a note in there that they get automatically installed but if not, do this.

nodeg: This is also a good solution.

keichwa: I think they were not automatically with an earlier release or MU. A note or remark as abid proposes sounds good.  Maybe, we can relax that statement later.

modules/installation-and-upgrade/pages/container-deployment/suma/server-deployment-suma.adoc
```diff
@@ -198,9 +214,9 @@ transactional-update register -p SUSE-Manager-Server/5.0/x86_64 -r <reg_code>
 . Reboot.
 
 
-=== Update the System
+==== Update the System
 
-.Procedure
+.Procedure: Updating the System
 . Log in as *root*.
 . Run **transactional-update**:
```
nodeg, line 221: Just out of curiosity, why do we not mention to install `mgradm` and `mgradm-bash-completion` here, like we do for the proxy?
image
image

keichwa: I think it now will get installed automatically.  Trying to ping @mcalmer for clarification: Do we still need to install `mgradm` and completion tools with the first `transactional-update` run after preparing Micro as the server host OS? On both, 5.0.4 and 5.1.x?

PR #3881, opened 2025-04-14
Update Server Upgrade: micro vs. sles/leap

keichwa:
**For link to HTML output,** see https://github.com/SUSE/spacewalk/issues/26987

https://github.com/SUSE/spacewalk/issues/26987

Will be superseded with #3911.

### DIFF
--- a/modules/installation-and-upgrade/pages/container-management/updating-server-containers.adoc
+++ b/modules/installation-and-upgrade/pages/container-management/updating-server-containers.adoc
@@ -22,15 +22,13 @@ zypper ref
 
 +
 
-On a transactional system such as {sl-micro} or {leapmicro}::
+On a transactional system such as {sl-micro} or {leapmicro}: ::
 
 +
 
 --
 
-.. Apply available updates with [command]``transactional-update``:
-
-+
+Apply available updates with [command]``transactional-update``:
 
 [source,shell]
 ----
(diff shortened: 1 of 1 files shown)


modules/installation-and-upgrade/pages/container-management/updating-server-containers.adoc
```diff
@@ -1,26 +1,72 @@
 = {productname} Server Upgrade
 
+// FIXME: Shouldn't we better say:
+//        "... to upgrade the host operating system first."
```
admd, line 4: +1. I agree, we should recommend to update the host OS 1st that will result in update of our tooling as well.

modules/installation-and-upgrade/pages/container-management/updating-server-containers.adoc
```diff
@@ -1,26 +1,72 @@
...
 ----
 zypper ref
 ----
-. Apply available updates with [command]``transactional-update``:
+
+. Depending on the host operating system, proceed with these steps:
+
++
+
+On a transactional system such as {sl-micro} or {leapmicro}: ::
```
admd, line 25: As this is againts master and we don't have branch yet for 5.1, this could create confusion for MLM users who might think that MLM is also supported on leapmicro

modules/installation-and-upgrade/pages/container-management/updating-server-containers.adoc
```diff
@@ -1,26 +1,72 @@
...
+
+[source,shell]
 ----
 transactional-update
 ----
+
+--
+
+
+On {sles} or {leap}: ::
```
admd, line 41: same here, As this is againts master and we don't have branch yet for 5.1, this could create confusion for MLM users who might think that MLM is also supported on leap

admd (changes requested):

PR #3896, opened 2025-04-17
Removed air-gapped deployment from Uyuni guide

0rnela:
# Description

# Target branches

- master

### DIFF
--- a/modules/installation-and-upgrade/nav-installation-and-upgrade-guide.adoc
+++ b/modules/installation-and-upgrade/nav-installation-and-upgrade-guide.adoc
@@ -29,7 +29,7 @@ ifeval::[{uyuni-content} == true]
 ** Deployment and Installation
 *** xref:install-server.adoc[Server]
 **** xref:container-deployment/uyuni/server-deployment-uyuni.adoc[Server Deployment on {leapmicro}]
-**** xref:container-deployment/uyuni/server-air-gapped-deployment-uyuni.adoc[Air-gapped Deployment]
+//**** xref:container-deployment/uyuni/server-air-gapped-deployment-uyuni.adoc[Air-gapped Deployment]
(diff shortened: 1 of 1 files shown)


juliogonzalez:
@0rnela this is strictly speaking not correct.

I think the page is wrong for both SUSE Multi-Linux Manager 5.1 and the future Uyuni, as we now provide RPMs with the images (for Uyuni those will come with the upcoming version).

So unless I am wrong, the instructions should mention the RPMs. And of course a single image is not enough anymore, as the postgresql container uses a separate image. This is valid for SUSE Multi-Linux Manager 5.1 and the future Uyuni

So even if I am wrong about the RPMs, this probably means this page:
- Needs to stay and maybe be duplicated (one for Uyuni one for SUSE Multi-Linux Manager) or needs IFs, so different image names and different registries are used for Uyuni or SUSE Multi-Linux Manager
- Needs changes to include the new postgresql image, or to use the RPMs

Please coordinate with Hexagon about this one.

juliogonzalez:
https://documentation.suse.com/suma/5.0/en/suse-manager/installation-and-upgrade/container-deployment/suma/server-air-gapped-deployment-suma.html does not mention skopeo or such things anymore, but the RPMs.

Some IMHO at master we need two pages, or ifs to use the  same method with slightly different things for Uyuni and SUSE Multi-Linux Manager (such as the RPM names)

jcayouette (approved):

juliogonzalez:
@jcayouette are you sure your approval is correct? See the comment I left. I don't see any changes since I sent the comment.

0rnela:
This PR is still a draft, therefore not ready for review.

jcayouette (changes requested):
Needs additional work.

PR #3903, opened 2025-04-22
Document renaming the journald log folder when changing machine id

cbosdo:
# Description

Journald expects the `/var/log/journal/xxx` folder to match the machine ID. This this isn't the case no log will be persisted.

# Target branches

- master
- 5.0
- 4.3

# Links
- This PR tracks issue https://bugzilla.suse.com/show_bug.cgi?id=1241286


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- Document renaming the journal folder when changing machine-id.
```
```suggestion
- Documented renaming the journal folder when changing machine ID in
  Administration Guide
```

modules/administration/pages/troubleshooting/tshoot-registerclones.adoc
```diff
@@ -52,6 +52,13 @@ dbus-uuidgen --ensure
 systemd-machine-id-setup
 ----
 
+If the cloned machine also has a folder in [path]``/var/log/journal/`` it also needs to be renamed with the new machine ID.
+Without this, [command]``journalctl`` would not persist any log and [command]``podman logs`` would not show anything.
+
```
keichwa, line 57:
```suggestion
. If the cloned machine also has a folder in [path]``/var/log/journal/`` it needs to be renamed accordingly to the new machine ID.
If names do not match, [command]``journalctl`` would not persist any log and [command]``podman logs`` would not show anything.

+

```
I do not understand what "would not persist any log" means in this context. (`enter` or `find` a log entry?)

keichwa (changes requested):
@cbosdo can you please help with clarifying?

modules/administration/pages/troubleshooting/tshoot-registerclones.adoc
cbosdo: it means that `journalctl -u` for any systemd service would not given anything: logs would be lost. Same for `podman logs <yourcontainer>` or `mgrpxy logs`: they would return nothing and the containers logs would be lost.

PR #3956, opened 2025-05-14
ISS files clean-up (follow-up to addition of Hub Online Synchronization)

0rnela:
# Description

- References to ISS v1 removed entirely.
- Clean-up and normalization of the files related to ISS v2.
- Links to Hub Online Syncing checked.
- 

# Target branches

- master

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/27009

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 - Removed references to Inter-Server Synchronization version 1 from
-  the documentation
+  documentation
 - Clarified that NFS with Cobbler is not supported (bsc#1240666)
 - Removed SCAP file size limit from Reference Guide
 - Added initial version of hub online synchronization to Large Deployments


modules/administration/pages/iss_v2.adoc
```diff
@@ -1,24 +1,18 @@
 [[iss2]]
-= Inter-Server Synchronization - Version 2
+= Inter-Server Synchronization version 2
```
keichwa, line 2: It looks as if you no longer want to use title style capitalization.  If that's the case, we should decide about it and then stick with it ;)

keichwa (approved):

modules/administration/pages/iss_v2.adoc
0rnela: We are switching back to sentence-style, in line with SLES documentation.

PR #3982, opened 2025-05-22
gh-pages update for uyuni-2025.05

0rnela:
gh-pages update for uyuni-2025.05


jcayouette:
Nice work, looks good.

PR #3970, opened 2025-05-19
Update network-requirements.adoc

mlaguilharre:
https://github.com/SUSE/spacewalk/issues/27330 (tracker)

Need to add settings to achieve server configuration behind http osi level 7 proxy.

- master
- 5.0
- 4.3

# Links
- This PR tracks issue #<insert spacewalk issue, if any>
- Related development PR #<insert PR link, if any>

PR #3988, opened 2025-05-26
improve layout and wording

keichwa:
It belongs to https://github.com/uyuni-project/uyuni-docs/pull/3970.


keichwa:
> @ke - it would be good to point to the PR this is an improvement to, so we can trace these changes from there too. Otherwise, it is unattached to anything, and more importantly - if backports of the original work are needed, then all these subsequent improvements can be missed.

Yes, I'll do! For the 5.0 backport, I already attached a proper changelog entry ;)

PR #3983, opened 2025-05-22
Rebranding: changing extensions of the product during the registration

szachovy:
# Description

I was unable to pull the Beta 2 images because I wasn't registered with SCC. I had been following https://w3.suse.de/~ke/mlm/mlm-migr-3962/docs/installation-and-upgrade/container-deployment/mlm/server-migration-mlm.html#_prepare_suse_multi_linux_manager_5_1_server_host, but it is no longer active. Instead, I used the following https://bugzilla.suse.com/show_bug.cgi?id=1243403#c8, and it worked:

For SL Micro 6.1:

```
~ # transactional-update --quiet register --list-extensions
AVAILABLE EXTENSIONS AND MODULES

    SUSE Linux Micro Extras 6.1 x86_64
    Activate with: transactional-update register -p SL-Micro-Extras/6.1/x86_64
    
    SUSE Multi-Linux Manager Proxy Extension 5.1 x86_64 (BETA)
    Activate with: transactional-update register -p Multi-Linux-Manager-Proxy/5.1/x86_64 -r ADDITIONAL REGCODE
    
    SUSE Multi-Linux Manager Server Extension 5.1 x86_64 (BETA)
    Activate with: transactional-update register -p Multi-Linux-Manager-Server/5.1/x86_64 -r ADDITIONAL REGCODE
    
    SUSE Multi-Linux Manager Retail Branch Server Extension 5.1 x86_64 (BETA)
    Activate with: transactional-update register -p Multi-Linux-Manager-Retail-Branch-Server/5.1/x86_64 -r ADDITIONAL REGCODE
```

For SLE 15 SP7 it's the same:

```
        Containers Module 15 SP7 x86_64 (Activated)
        Deactivate with: suseconnect -d -p sle-module-containers/15.7/x86_64
        
            SUSE Multi-Linux Manager Proxy Extension for SLE 5.1 x86_64 (BETA)
            Activate with: suseconnect -p Multi-Linux-Manager-Proxy-SLE/5.1/x86_64 -r ADDITIONAL REGCODE
            
            SUSE Multi-Linux Manager Server Extension for SLE 5.1 x86_64 (BETA)
            Activate with: suseconnect -p Multi-Linux-Manager-Server-SLE/5.1/x86_64 -r ADDITIONAL REGCODE
            
            SUSE Multi-Linux Manager Retail Branch Server Extension for SLE 5.1 x86_64 (BETA)
[...]

PR #3990, opened 2025-05-26
use new productname for registration

keichwa:
If approved this might supersede https://github.com/uyuni-project/uyuni-docs/pull/3983.


keichwa:
> Ok, I will close mine as soon as it gets merged #3983.
> 
> Approving, but please look at the GH actions error.

I think that's false alarm ATM.  Thanks for approving.  When Ornela is done with the uyuni release, we can merge it.

PR #3991, opened 2025-05-26
Re-arrange some steps for better understanding

admd:
# Description

Re-arrange some steps for better understanding. Also added a note about upgrading 4.3 server before starting the migration process.

# Target branches

* Which product version this PR applies to (Uyuni, SUMA 5.0 and 5.1).  This information can be helpful if `ifeval` statements are needed to publish it for certain products only.

- master
- 5.0

# Links

### DIFF
--- a/modules/installation-and-upgrade/pages/hardware-requirements.adoc
+++ b/modules/installation-and-upgrade/pages/hardware-requirements.adoc
@@ -279,6 +279,7 @@ These scenarios are not supported.
 
 On the {productname} Server, use this command to find all available storage devices:
 
+[source, shell]
 ----
 hwinfo --disk | grep -E "Device File:"
 ----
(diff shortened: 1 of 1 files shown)


modules/installation-and-upgrade/pages/container-deployment/mlm/server-migration-mlm.adoc
```diff
@@ -181,25 +181,25 @@ For example on {productname} server:
...
 spacewalk-service stop
 ----
-.. Stop the PostgreSQL service on 4.3 Server:
+. Stop the PostgreSQL service on 4.3 Server:
 +
 ----
 systemctl stop postgresql
 ----
-.. Perform the migration on {productname} server
+. Perform the migration on {productname} server
```
admd, line 197: We need to check if rendering would be fine here. I added second level of navigation as with 1st it was looking bit strange but then I didn't check the rendered version. I only checked with atom asciidoc plugin.

keichwa (approved):

PR #3995, opened 2025-05-27
Manager 5.0: proxy check reboot #27329 keichwa

keichwa:
**Link to HTML,** see https://github.com/SUSE/spacewalk/issues/27329

- https://github.com/SUSE/spacewalk/issues/27329

### DIFF
--- a/modules/installation-and-upgrade/pages/container-deployment/actkey-bootstrap-proxy-suma.adoc
+++ b/modules/installation-and-upgrade/pages/container-deployment/actkey-bootstrap-proxy-suma.adoc
@@ -1,13 +1,5 @@
 == Create an Activation Key for the Proxy
 
-// FIXME: Maybe, we'd better remove this WARNING again, or
-//        add more info.
-
-[WARNING]
-====
-If deploying on {sles}, you can skip some reboots.
-====
-
 .Procedure: Creating an Activation Key
 . Navigate to menu:Systems[Activation Keys], and click btn:[Create key].
 . Create an activation key for the proxy host with {sle-micro} {microversion} or {sles} {bci-suma} as the parent channel.
@@ -25,12 +17,9 @@ If deploying on {sles}, you can skip some reboots.
 . Click btn:[Bootstrap].
 . Wait for the bootstrap process to complete successfully.
   Check the menu:Salt[] menu and confirm the {salt} key is listed and accepted.
-// FIXME: Can we skip the following two reboots on both Micro *and* SLES?
-+
-. Reboot the proxy host.
-. Select the host from the menu:System[] list and trigger a second reboot after all events are finished to conclude the onboarding.
+. Reboot the proxy host if the operating system is {sle-micro}.
+. Select the host from the menu:System[] list and trigger a second reboot in case of {sle-micro} after all events are finished to conclude the onboarding.
 
-// FIXME: Do we nee this procedure on both Micro *and* SLES?
 .Procedure: Updating the Proxy Host
 . Select the host from the menu:Systems[] list and apply all patches to update it.
-. Reboot the proxy host.
+. Reboot the proxy host if the operating system is {sle-micro}.


modules/installation-and-upgrade/pages/container-deployment/actkey-bootstrap-proxy-suma.adoc
```diff
@@ -1,8 +1,16 @@
 == Create an Activation Key for the Proxy
 
+// FIXME: Maybe, we'd better remove this WARNING again, or
+//        add more info.
+
+[WARNING]
+====
+If deploying on {sles}, you can skip some reboots.
+====
```
admd, line 9: such vogue statements doesn't help user. We should mention on the exact step if that step is applicable on sales or not.

modules/installation-and-upgrade/pages/container-deployment/actkey-bootstrap-proxy-suma.adoc
```diff
@@ -17,12 +25,12 @@
 . Click btn:[Bootstrap].
 . Wait for the bootstrap process to complete successfully.
   Check the menu:Salt[] menu and confirm the {salt} key is listed and accepted.
+// FIXME: Can we skip the following two reboots on both Micro *and* SLES?
```
admd, line 28: We need to verify this from our QE squad. Howerver, I don't believe these 2 and then next on line 36 after updating the proxy Host are needed for sales. So we should rather like it like

`Reboot the proxy host if sle micro ` or something like this.

For now I would say, make the change by adding the if sle micro explicitly in the docs. QE can then test based on those steps and let us know if that's correct or we need any modification.

@nodeg FYI

modules/installation-and-upgrade/pages/container-deployment/prepare-micro-host.adoc
```diff
@@ -1,22 +1,28 @@
...
-. Boot or reboot your system.
+
+
+=== Install {sle-micro} {microversion}
 
 For more information about preparing your machines (virtual or physical), see link:https://documentation.suse.com/sle-micro/5.5/html/SLE-Micro-all/book-deployment-slemicro.html[{sle-micro} {microversion} Deployment Guide].
 
 .Procedure: Installing {sle-micro} {microversion}
+. Insert the DVD or USB flash drive (USB disk or key) containing the installation image for {sle-micro} {microversion}.
+. Boot or reboot your system.
```
admd, line 17: I don't believe, this step is needed.

modules/installation-and-upgrade/pages/container-deployment/actkey-bootstrap-proxy-suma.adoc
nodeg: IIRC the reboots were only introduced and necessary for transactional systems like SL Micro and the way they work. When using e.g. SLES as base OS, no reboot is required since changes can be made right away.

modules/installation-and-upgrade/pages/container-deployment/prepare-micro-host.adoc
keichwa: But isn't that the way to kick-off the installation (at least on-prem)?

keichwa:
At the other places, steps are kept and only FIXMEs are removed.

modules/installation-and-upgrade/pages/container-deployment/prepare-micro-host.adoc
admd: ah correct, sorry I was just focusing on reboot part. Yes, you are right.

admd (approved):
LGTM now. Thank you Karl.

modules/installation-and-upgrade/pages/container-deployment/prepare-micro-host.adoc
keichwa: ;)

PR #4016, opened 2025-06-03
warn about root with SSH password configuration

keichwa:
**For a link to HTML output,** see https://github.com/SUSE/spacewalk/issues/27367

- https://bugzilla.suse.com/show_bug.cgi?id=1243569
- https://github.com/SUSE/spacewalk/issues/27367


CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
+- (bsc#1243569)
```
```suggestion
- Documented root SSH login configuration with password in Installation
  and Upgrade Guide (bsc#1243569)
```

modules/installation-and-upgrade/pages/container-deployment/prepare-micro-host.adoc
```diff
@@ -133,3 +133,54 @@ If you prefer the {sl-micro} default behavior, enable the timer by running the f
 systemctl enable --now transactional-update.timer
 ----
 ====
+
+
+
+=== Allow SSH root login with a password
```
admd, line 139: We should rather add this in troubleshooting and add a link here.

Also, we should explain why is this needed which is obvious we should write it down that it is required because root SSH login configuration with password is disabled in Sl micro 6.1 and I believe it will be same in 6.2 as well.

Also this is not limited to proxy but any 6.1/6.2 client will have this.

Also the SLE Micro 5.5 systems which will when be migrated to 6.1/6.2 will suddenly lose this access which was working before the migration. We should add all this info.

keichwa: Thanks for background info!  In the dedicated SL Micro documentation I found a cockpit related statement only.  Now I checked the release notes, and it has a more general statement that we can resuse or refer somehow-- https://www.suse.com/releasenotes/x86_64/SL-Micro/6.1/index.html:

> 2.1.5 Password-based remote root login not allowed

> For enhanced security, new installations of SL Micro 6.1 do not allow password-based remote root login anymore, which also affects Cockpit. The login to Cockpit now requires to create and use an unprivileged user and to optionally set up 2-factor authentication with one-time passwords. The firstboot-installer as well as ignition and the online tool fuel-ignition support this setup.

> For more information about the procedure and other options see https://documentation.suse.com/sle-micro/6.1/html/Micro-cockpit/index.html#cockpit-authentication-basics.

modules/installation-and-upgrade/pages/container-deployment/prepare-micro-host.adoc
```diff
@@ -133,3 +133,54 @@ If you prefer the {sl-micro} default behavior, enable the timer by running the f
...
+
+--
+
+.. In [path]``/usr/etc/ssh/sshd_config`` file, set:
+
++
+
+----
+PermitRootLogin yes
+PasswordAuthentication yes
```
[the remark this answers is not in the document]
keichwa: ok, thanks, good news :)

keichwa:
Here now is such a troubleshooting article: https://w3.suse.de/~ke/mlm/mlm-ssh-4016/docs/administration/troubleshooting/tshoot-remote-root-on-micro.html
I'll wait with adding xrefs to the deploy sections until Joseph has the migration chapters ready.

0rnela (approved):

PR #4020, opened 2025-06-04
Adjust postgresql paths for database tuning

aaannz:
# Description

This PR updates path for the database tuning.

However it needs some info added somewhere that these links are accessible from the container host and not from the container itself.

# Target branches

- master

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/26916


keichwa (commented):
In https://github.com/SUSE/spacewalk/issues/26916, we have this statement:
> The tuning of postgresql.conf should be done in the DB container. We want to create a mgrctl command to connect to DB and perform this action.
I'm wondering whether we should mention it here as well?

keichwa (approved):

0rnela (approved):

PR #4027, opened 2025-06-10
Ansible for Control Node comes with distribution

keichwa:
**For link to (updated) HTML,** see https://github.com/SUSE/spacewalk/issues/27441

- https://bugzilla.suse.com/show_bug.cgi?id=1244125
- https://github.com/SUSE/spacewalk/issues/27441

### DIFF
--- a/modules/administration/pages/ansible-setup-control-node.adoc
+++ b/modules/administration/pages/ansible-setup-control-node.adoc
@@ -36,12 +36,12 @@ If you want to install a newer Ansible on a {sle} 15 SP4 or SP5 client, you mus
 
 [NOTE]
 ====
-With newer Ansible, systems with an old Python version cannot be managed anymore.
+Newer versions of Ansible no longer support managing nodes with outdated Python versions.
+If a managed node still defaults to an older Python version, you may encounter connection errors or failures during playbook runs.
+To address this, user should upgrade Python on the managed node, if possible and set the correct Python interpreter in the Ansible inventory or configuration.
 ====
 
 
-//Procedure: Installing the SCAP Security Guide Package
-include::snippets/install_scap_security_guide_package.adoc[]
 
 
 [[configure-ansible-inventory-files]]


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+- Documented the only Client Tool for SLE 15 have the Ansible package
```
0rnela, line 1: Please rephrase. The entry is unclear.

keichwa: thanks, I hope it is better now!

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -3,7 +3,14 @@
 
 To set up an Ansible control node, execute the following steps from the {productname} {webui}.
 
-.Procedure: Setting up Ansible Control Node
+[NOTE]
+====
+The following procedure applies for {sle} 15 system only.
```
admd, line 8: We are not shipping ansible in any of our client tools anymore in 5.1. So we should adjust it accordingly. For every distribution, ansible needs to come from the distrubution.

For 15SP6 and SP7, it should be coming be coming from `Systems Management Module` which has dependency on `Python 3 Module` so both of these modules needs to be enabled if user wants to install ansible package on SP6/SP7 system in order to make them control node.

For SP4 and SP5, if user want to install new Ansible, they need to enable only Python 3 Module ` , this is what that bug was about where you were tagged.

We also need to add a note with newer ansible, systems with old python cannot be managed anymore.

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -15,7 +22,7 @@ To set up an Ansible control node, execute the following steps from the {product
 
 [NOTE]
 ====
-The {productname} client tools contain the [package]``ansible`` package.
+On {sle} 15, the {productname} client tools contain the [package]``ansible`` package.
```
admd, line 25: This needs to be removed or modified with what I have described above.

modules/administration/pages/ansible-setup-control-node.adoc
keichwa: Thanks.  I added all this. I think here we shouldn't document Ansible version related limitations that detailed.  Focus here should be setting up a (new) Ansible control with MLM 5.1.

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -3,20 +3,22 @@
 
 To set up an Ansible control node, execute the following steps from the {productname} {webui}.
 
-.Procedure: Setting up Ansible Control Node
+[NOTE]
+====
+To configure a {productname} client the Ansible software must be available for operating system of the client.
+This usually means Ansible must come from the client's software distribution.
+For example, Ansible is available for {sle} 15 SP6 or SP7.
+====
```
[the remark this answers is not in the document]
admd: I added a suggestion but feel free to modify it

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -25,6 +27,16 @@ The {productname} client tools contain the [package]``ansible`` package.
 
 . Select the menu:Events[] tab and verify the status of the highstate.
 
+[NOTE]
+====
+If you want to install a newer Ansible on a {sle} 15 SP4 or SP5 client, you must enable only the [literal]``Python 3 Module``.
```
admd, line 32: remove the 'only'

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -25,6 +27,16 @@ The {productname} client tools contain the [package]``ansible`` package.
...
+====
+
+[NOTE]
+====
+With newer Ansible, systems with an old Python version cannot be managed anymore.
+====
```
admd, line 38: We can probably extend it with something like this 

Newer versions of Ansible no longer support managing nodes with outdated Python versions 
If a managed node still defaults to an older Python version, you may encounter connection errors or failures during playbook runs. To address this, user should  Upgrade Python on the managed node, if possible and Set the correct Python interpreter in your Ansible inventory or configuration.

admd: @meaksh can you please verify this

admd (changes requested):
//Procedure: Installing the SCAP Security Guide Package
include::snippets/install_scap_security_guide_package.adoc[]

This doesn't belong here and should be removed.

modules/administration/pages/ansible-setup-control-node.adoc
meaksh: that's correct :+1:

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+- Documented that only Client Tools for SLE 15 channel contains the
+  Ansible package (bsc#1244125)
```
meaksh, line 2: Is this actually the right changelog? shouldn't this be the opposite:

```suggestion
- Documented that not only Client Tools for SLE 15 channel contains the
  Ansible package (bsc#1244125)
```

modules/administration/pages/ansible-setup-control-node.adoc
keichwa, line 32:
```suggestion
If you want to install a newer Ansible on a {sle} 15 SP4 or SP5 client, you must enable the [literal]``Python 3 Module``.
```

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -3,20 +3,22 @@
...
+. In the {productname} {webui}, navigate to menu:Admin[Setup Wizard > Products], verify that [literal]``SUSE Linux Enterprise Server 15 SP6 x86_64`` (or later) with the [literal]``Systems Management Module`` and the required [literal]``Python 3 Module`` are selected and synchronized.
 
-[NOTE]
-====
-The {productname} client tools contain the [package]``ansible`` package.
-====
+. Deploy a {sle} 15 SP6 (or later) client.
+
+. In the {productname} {webui}, navigate to the menu:Systems[Overview] page of the client.
+  Select menu:Software[Software Channels] and subscribe the client to the [literal]``SUSE Linux Enterprise Server 15 SP5 x86_64``, [literal]``Systems Management Module`` and  [literal]``Python 3 Module`` channels.
```
meaksh, line 23: I think there is a typo here, I would suggest:
```suggestion
  Select menu:Software[Software Channels] and subscribe the client to the [literal]``SUSE Linux Enterprise Server 15 SP6 x86_64`` (or later SP), [literal]``Systems Management Module`` and  [literal]``Python 3 Module`` channels.
```

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -3,20 +3,22 @@
...
 
-.Procedure: Setting up Ansible Control Node
+[NOTE]
+====
+To configure a {productname} client the Ansible software must be available for operating system of the client.
+This usually means Ansible must come from the client's software distribution.
+For example, Ansible is available for {sle} 15 SP6 or SP7.
+====
 
-. In the {productname} {webui}, navigate to menu:Admin[Setup Wizard > Products], verify that [literal]``SUSE Linux Enterprise Server 15 SP5 x86_64`` and [literal]``{productname} Client Tools for SLE 15 x86_64`` are selected and synchronized.
```
meaksh, line 8: Shouldn't we mention that the old ansible version can be found in SLE15 client tools? We are dropping all the information about this in the page

meaksh (changes requested):

modules/administration/pages/ansible-setup-control-node.adoc
```diff
@@ -25,6 +29,16 @@ The {productname} client tools contain the [package]``ansible`` package.
...
+====
+
+[NOTE]
+====
+With newer Ansible, systems with an old Python version cannot be managed anymore.
+====
+
 
 //Procedure: Installing the SCAP Security Guide Package
 include::snippets/install_scap_security_guide_package.adoc[]
```
keichwa, line 44:
```suggestion
```

modules/administration/pages/ansible-setup-control-node.adoc
keichwa: ok

modules/administration/pages/ansible-setup-control-node.adoc
admd: We have new set of SLE15 client tools in 5.1 and those client tools don't have ansible. 

For 5.0, yes, we need to keep the info about the client tools.

meaksh: Hmmm, I still see `ansible` at:
- https://build.suse.de/package/show/Devel:Galaxy:Manager:Head:MLMTools-Beta-SLE15/ansible
- https://build.suse.de/package/show/Devel:Galaxy:Manager:5.1:MLMTools-SLE15/ansible

Aren't these the new client tools? Maybe the package are present even if not really submitted.. we would need some clarifications.

keichwa: I'll merge this one now.  If needed we must create a follow-up PR.

admd: @meaksh sorry missed your comment before. I don't know why ansible package is displayed in IBS but I don't see that here https://updates.suse.de/SUSE/Products/MultiLinuxManagerTools/SLE-15/x86_64/product/ and it isn't expected. I will check with @deneb-alpha

PR #4034, opened 2025-06-12
Fixed remediation procedure in Ansible

keichwa:
**For link to refreshed HTML output,** see https://github.com/SUSE/spacewalk/issues/27497.

- https://bugzilla.suse.com/show_bug.cgi?id=1244338
- https://github.com/SUSE/spacewalk/issues/27497

### DIFF
--- a/modules/administration/pages/ansible-compliance-as-code.adoc
+++ b/modules/administration/pages/ansible-compliance-as-code.adoc
@@ -18,20 +18,7 @@ The following procedure will guide you through running remediation using an Ansi
 
 .Procedure: Run Remediation Using an Ansible Playbook
 
-. From the control node system menu select menu:Ansible[Playbooks].
-
-. In [guimenu]``Add a Playbook directory`` field, enter the path to the OpenSCAP playbooks, for example:
-
-+
-
-[source,shell]
-----
-/usr/share/scap-security-guide/ansible
-----
-
-. Confirm with btn:[Save].
-
-. Navigate Go menu:Ansible[Playbooks], expand the folder you just declared, and click a playbook, for example:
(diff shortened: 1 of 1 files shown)


modules/administration/pages/ansible-compliance-as-code.adoc
```diff
@@ -14,19 +14,43 @@ include::snippets/install_scap_security_guide_package.adoc[]
...
++
+
+[source,shell]
+----
+/usr/share/scap-security-guide/ansible
+----
+
+. Confirm with btn:[Save].
+
+. Navigate Go menu:Ansible[Playbooks], expand the folder you just declared, and click a playbook, for example:
```
cbbayburt, line 34: In 5.1, we are adding the default Playbook directory by default, so this step is not required.

In case the user has the playbook in a different directory, they can follow the link to "Setup Ansible Control Node" to find out how to add it.

It makes sense to have this step added for 5.0 and before though.

modules/administration/pages/ansible-compliance-as-code.adoc
```diff
@@ -14,19 +14,43 @@ include::snippets/install_scap_security_guide_package.adoc[]
...
+
++
+
+[source,shell]
+----
+sle15-playbook-stig.yml
+----
+
+. To run the playbook, select the [guimenu]``OS Inventory Path`` for the clients, for example:
```
cbbayburt, line 43: This is the actual name of the label in the UI
```suggestion
. To run the playbook, select the [guimenu]``Inventory Path`` for the clients, for example:
```

cbbayburt (changes requested):
Thanks for the update, @keichwa.

Below are a couple of suggestions.

cbbayburt (approved):

PR #4041, opened 2025-06-16
Backup & Restore clarifications

keichwa:
- https://github.com/SUSE/spacewalk/issues/27527
- https://bugzilla.suse.com/show_bug.cgi?id=1244524

### DIFF
--- a/modules/administration/pages/backup-restore.adoc
+++ b/modules/administration/pages/backup-restore.adoc
@@ -44,7 +44,9 @@ A full backup of the {productname} consists of backing up the following componen
 
 [WARNING]
 ====
-The {productname} service is automatically restarted during creation of a full backup, the downtime can be significant.
+The {productname} service is automatically stopped for the time it takes to create a full backup.
+The downtime can be significant.
+After backup is done, service is automatically restarted.
 ====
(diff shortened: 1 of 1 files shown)


modules/administration/pages/backup-restore.adoc
```diff
@@ -44,7 +44,7 @@ A full backup of the {productname} consists of backing up the following componen
 
 [WARNING]
 ====
-The {productname} service is automatically stopped during creation of a full backup.
+The {productname} service is automatically restarted during creation of a full backup, the downtime can be significant.
```
aaannz, line 47: I would prefer something like:

```suggestion
The {productname} service is automatically stopped for the duration of creating of a full backup, the downtime can be significant.
After backup is done, service is automatically restarted.
```

It, in my opinion, better explains what is going on and that it is not an instant restart.

keichwa: Alternate suggestion by keichwa:
```suggestion
The {productname} service is automatically stopped for the time it takes to create a full backup.
The downtime can be significant.
After backup is done, service is automatically restarted.
```

PR #4062, opened 2025-06-24
id deduplicated

keichwa:
Maybe, we do not need this list at all...


keichwa:
> @ke - approved. But if you think we don't need the list, please remove it.

ok, thanks. but I'm not that familiar with the actions feature.

PR #4081, opened 2025-06-26
backport 5.0.5: Fix how to change the hostname of the server

keichwa:
Even if the procedure is close enough from what was documented after the script fix, it still needs adjustments since the user has to pass the FQDN rather than the IP.


keichwa:
@cbosdo I guess if possible, we want to have this also published as part of the MU-5.0.5 documentation release.

0rnela (approved):

cbosdo (approved):

PR #4083, opened 2025-06-26
Renaming files to reflect they are snippets

0rnela:
# Description

Part of the unused or orphan files cleanup, this PR introduces temporary solution for a selection files that are snippets, but are currently not stored in the snippet directory. T

# Target branches

- master (5.1 and Uyuni only)

# Links


keichwa (changes requested):
Please wait with this until we have branched `manager-5.1`.

keichwa (approved):
ok, as long as you have verified things will continue to work as expected, esp. includes and cross-ref (if there are any).

PR #4112, opened 2025-07-02
use swapfile

keichwa:
**For link to HTML,** see https://github.com/SUSE/spacewalk/issues/27542

### DIFF
--- a/modules/installation-and-upgrade/pages/hardware-requirements.adoc
+++ b/modules/installation-and-upgrade/pages/hardware-requirements.adoc
@@ -73,7 +73,7 @@ For more information about filesystem and partitioning details, see xref:install
 | Minimum 50 GB
 
 | Swap space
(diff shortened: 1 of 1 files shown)


aaannz:
Can we also add an example snippet how to prepare such swap file?

This snippet creates 8GiB swap file, users needs to check the available storage and also quick storage should be used for better performance.

Note that creating swap file on btrfs filesystem prevent creating of a snapshot of that volume. `/var` in the example below is already excluded from the snapper snapshots so it is safe to use `/var/swap`.

```bash
## setup swapfile at /var/swap
# run following as a root user

# allocate 8GiB for swap file
fallocate -l 8G /var/swap
# ensure CoW is disabled for the swap file
chattr +C /var/swap
# allow only root access
chmod 600 /var/swap
# make swap file based on allocated file
mkswap /var/swap
# activate swap use for the running system
swapon /var/swap
# activate swap during the next boots
echo "/var/swap swap swap defaults 0 0" >> /etc/fstab
```

keichwa:
> Can we also add an example snippet how to prepare such swap file?

Thanks for the proposal and esp. for the draft info!  I added a section, shortened the text in the tables, and referenced the new section.

With a separate PR, I can change the Retail tables accordingly.

PR #4113, opened 2025-07-02
4.3 - Added info about missing package

0rnela:
# Description

The bug identified the package missing on proxy if monitoring was configured on it.

# Target branches
- 4.3

# Links
- Bugzilla https://bugzilla.suse.com/show_bug.cgi?id=1191142


keichwa (changes requested):
Sth went wrong here.

PR #4119, opened 2025-07-02
4.3 - Removed system getcrashfiles

0rnela:
# Description

The method has been removed from the product, but had remained in the docs.

# Target branches

- 4.3

# Links
- Bugzilla https://bugzilla.suse.com/show_bug.cgi?id=1209060


keichwa (approved):

0rnela:
> I recommend we remove spacecmd from the reference guide entirely. The command-line tool already includes built-in help, and I'm confident the existing spacecmd documentation is outdated. As long as we mention the tool in the 'Tools' section, that should be sufficient.

@jcayouette - this is a quick fix for 4.3. For the task you propose it's probably best to create an issue, get Abid's approval, and then we can go ahead.

PR #4159, opened 2025-07-09
Improved step in the procedure

0rnela:
# Description

From the bug report: 

"Point 1 says 'At the command prompt in the SUSE Manager Server...'
With containers, it might not be clear for everybody that this means logging into the container. I suggest to make this more clear."

# Target branches

- master
- 5.0 https://github.com/uyuni-project/uyuni-docs/pull/4158

# Links
- Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1233494


0rnela:
> Maybe, already fixed in the meantime. You must run `mgrctl` on the container host (as documented). This PR is not needed.

Fair point. Thanks.

PR #4181, opened 2025-07-17
5.0 bsc1246422 add third party channels

0rnela:
# Description

https://bugzilla.suse.com/show_bug.cgi?id=1246422

# Target branches

- master
- 5.0

### DIFF
--- a/modules/administration/pages/third-party-channels.adoc
+++ b/modules/administration/pages/third-party-channels.adoc
@@ -24,8 +24,8 @@ To import a GPG key, use the following command syntax, executing the command on 
 # mgradm gpg add https://public.dhe.ibm.com/software/server/POWER/Linux/yum/OSS/SLES/15/ppc64le/repodata/repomd.xml.key
 ----
 
-To list the keys currently held in the {productname} GPG database, run the following command from inside the server container:
+To list the keys currently held in the {productname} GPG database, run the following command:
 
 ----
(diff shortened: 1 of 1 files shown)


cbosdo (commented):
How comes that your PR adds content and doesn't change the existing page as suggested by the report?

mcalmer:
Something you could also add to the end.

What you described here, is how to get the "repo-sync" done when metadata are signed or the signature of the packages are checked.

But when installing a package on the client, you may need to trust the GPG key also on the client.
This requires, that the GPG key is also available there.
We already have decscribed how to achieve this:

https://www.uyuni-project.org/uyuni-docs/en/uyuni/client-configuration/gpg-keys.html#_trust_gpg_keys_on_clients

Maybe a reference to this place would be good

modules/administration/nav-administration-guide.adoc
```diff
@@ -32,6 +32,7 @@ endif::[]
 *** xref:custom-channels.adoc[Custom Channels]
 
 ifeval::[{suma-content} == true]
+*** xref:third-party-channels.adoc[Third-Party Channels]
```
mcalmer, line 35: Why only in "suma-content". 
I think Uyuni user have the same problems

0rnela: Thanks, changed.

0rnela:
New PR created.

PR #4189, opened 2025-07-22
disable HSTS

keichwa:
**Find link to HTML in:** https://github.com/SUSE/spacewalk/issues/27857

https://github.com/SUSE/spacewalk/issues/27857


rjmateus:
I tried to tested it, but in our reference server I was unable to check that the HSTS was enabled by default. I think we need a few more tests

PR #4244, opened 2025-08-14
Fixed procedure formatting

0rnela:
# Description

Docs building kept showing Warning "unterminated quote block". The cause was Proxy migrating procedure 5.0 to 5.1 had formatting issue.

# Target branches

- master
- 5.1 https://github.com/uyuni-project/uyuni-docs/pull/4245

PR #4245, opened 2025-08-14
5.1 - Fixed procedure formatting

0rnela:
# Description

Docs building kept showing Warning "unterminated quote block". The cause was Proxy migrating procedure 5.0 to 5.1 had formatting issue.

# Target branches

- master https://github.com/uyuni-project/uyuni-docs/pull/4244
- 5.1


jcayouette (commented):
This is how procedures should be formatted. With the new branding procedures now how additional styles applied through the role and example block. `____`. 

I plan to use AI to adjust all our procedures to match this new style. So do not worry about adjusting these manually.

```
[role=procedure]
____
. Task 1
. Task 2
. Task 3
____

```

PR #4404, opened 2025-10-17
Remove the option about editing the sshd_config file

admd:
# Description

Remove the option about editing the sshd_config file.

# Target branches

- master
- 5.1 https://github.com/uyuni-project/uyuni-docs/pull/4413

# Links
- This PR tracks bug https://bugzilla.suse.com/show_bug.cgi?id=1252106 / issue https://github.com/SUSE/spacewalk/issues/28643


aaannz (commented):
We don't have to remove it completely. It's even in a FIXME here.

Instead of `/usr/..` path, just suggest to create a config drop-in `/etc/ssh/sshd_config.d/permit_root.conf` with mentioned content.

There is also no need to restart a server unlike when installing a package.

admd:
Thanl you @aaannz , can you please take another look.

PR #4431, opened 2025-10-24
Add missing product refresh to proxy migration

nodeg:
# Description

In order to do
image
you need to first refresh the product catalog. Otherwise there is no Proxy Extension 5.1 and no client tools for SL Micro 6.1 shown. At least if you did the server migration right before.

I also added a note to the common workflow that Abid created as per his request. In the meantime I also removed all of the trailing whitespaces at the end of the file.

# Target branches

- master
- 5.1 (this PR)

# Links
- https://documentation.suse.com/multi-linux-manager/5.1/en/docs/installation-and-upgrade/container-deployment/mlm/migrations/proxy/proxy-mlm-50-51.html#_distribution_upgrade_and_proxy_migration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Added product refresh step under proxy migration in
+  Administration Guide and Common Workflows
 - Added SSL certificates requirement before migrations
 - Updated workflow for switching to new client tools channels
   in Common Workflows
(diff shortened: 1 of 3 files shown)


admd (approved):
Good point. Thank you

nodeg:
image

image

PR #4529, opened 2025-11-13
Added missing configuration step to snippet for generating proxy configuration

0rnela:
# Description

Added step for copying the RHN-ORG-TRUSTED-SSL-CERT and RHN-ORG-PRIVATE-SSL-KEY to container before generating the proxy configuration via CLI.

# Target branches

- master
- 5.1
- 5.0

# Links
- This PR tracks issue  https://github.com/SUSE/spacewalk/issues/28396


jcayouette (approved):

0rnela:
There needs to be more discussion about this fix. the big reported had initially agreed with the proposal, but engineering had remarks.

This is now work in progress.

PR #4643, opened 2026-01-21
Update hashbang in build scripts

KucharczykL:
# Description

On some systems /bin/bash doesn't exist (NixOS, for one) so using the `env` binary is a safe way to find out where `bash` is located.

Tested on:
* NixOS
* Debian 12
* Ubuntu 25.04
* openSUSE Leap 15.6
* openSUSE Tumbleweed

# Target branches

- master
- 5.1
- 5.0
- 4.3

### DIFF
--- a/cleanup_pdfs.sh
+++ b/cleanup_pdfs.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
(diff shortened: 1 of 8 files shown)


jcayouette:
Yes this looks good to me. Should help with compatibility.

PR #4705, opened 2026-02-09
Add server and proxy migration documentation from 5.0 to 5.2

0rnela:
# Description

Add server migration path from 5.0 to 5.2.

# Target branches
- master

Check the changes **[here](https://w3.suse.de/~omaric/Docs/4migrations/docs/installation-and-upgrade/container-deployment/mlm/migrations/server/server-mlm-50-52.html)**. 

# Links
- This PR tracks issue ???


ktsamis (commented):
Besides the 2 small version numbers, this doc will be for beta 1 right? Because in alpha2 we have a problem with the migration and it needs a workaround https://github.com/uyuni-project/uyuni/issues/11444#issuecomment-3854955442. If it goes in alpha2 then we need this workaround, if it goes in beta 1 then it should be fixed by a PR and it won't be necessary.

ktsamis (approved):
I think it looks good, I see no issue

PR #4728, opened 2026-02-17
Add migration guides for server and proxy from 5.1 to 5.2

0rnela:
Add migration guides for server and proxy from 5.1 to 5.2


0rnela:
Files can be checked here: https://w3.suse.de/~omaric/Docs/51-to-52/

PR #4780, opened 2026-03-11
Fix typo in EOL clients partial

nodeg:
# Description

Fixes a small typo that I found while browsing the docs for EOL clients.

# Target branches

- 5.1
- 5.0


KucharczykL:
Thank you!

PR #4822, opened 2026-03-26
QE: fix redundant Tumbleweed in requirements table

Digvijay-x1:
# Description

I noticed that the "tumbleweed" is redundant ( written under both the "Software/Hardware" Column and "Recommendation" Column ). Having "Operating System" instead of "Tumleweed" in the "Software/Hardware" Column would match other generic names like CPU, RAM etc. 

This PR replaced the first-row label from {tumbleweed} to Operating System in both:
- Server Requirements table
- Proxy Requirements table

Before: 
image

After: 
image

# Target branches

- master ( This PR ) 
- 5.1
- 5.0
- 4.3

# Links

### DIFF
--- a/modules/installation-and-upgrade/pages/uyuni-install-requirements.adoc
+++ b/modules/installation-and-upgrade/pages/uyuni-install-requirements.adoc
@@ -26,7 +26,7 @@ Do not use NFS for storage because it does not support SELinux file labeling.
 | Details
 | Recommendation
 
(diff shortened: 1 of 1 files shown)


0rnela:
Thank you for your contribution. 
While we see your perspective, at the moment we are proceeding with using _Tumbleweed_ explicitly in the requirements instead of a more generic term.

PR #5042, opened 2026-06-29
5.1 - Procedure for uploading large files

0rnela:
# Description

Procedure added with the configuration for uploading files larger than 1GB.

# Target branches

- master
- 5.1

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/31035 / bug https://bugzilla.suse.com/show_bug.cgi?id=1268673


KucharczykL (approved):

en/modules/specialized-guides/pages/large-deployments/tuning.adoc
```diff
@@ -88,6 +86,39 @@ To re-tune your installation, you will need to use the dependency graph again.
 Start from the node where significant change has happened.
 
 
+== Upload files larger then 1GB
+
+.Procedure: Uploading files larger then 1GB
```
aaannz, line 91: I would not find this info when looking for a solution to mentioned bug.

This is apache2 parameter, so should be under parameters section:

- Description:
> This parameter specifies the number of bytes that are allowed in a request body. A value of 0 means unlimited.
- Location "Custom file: `/etc/apache2/conf.d/maxrequestbody.conf`"
- Value default: 1073741824 (1GB)
- Example: `LimitRequestBody 0`
- More information: https://httpd.apache.org/docs/current/mod/core.html#limitrequestbody
- Tune when:
> When following message appears in the logs: "Request entity too
large! The POST method does not allow the data transmitted, or the data volume exceeds the capacity limit."

PR #5165, opened 2026-07-28
Added procedure for recovery from a failed migration

0rnela:
# Description

From bug:
"If an administrator runs `mgradm upgrade` and the containerized database migration fails (for example, due to a "No space left on device" error), there is currently no documentation explaining how to recover. This leads to customer panic, uncleared storage volumes, and preventable support tickets"

# Target branches

- master
- 5.2

# Links
- This PR tracks bug https://bugzilla.suse.com/show_bug.cgi?id=1271596.

### DIFF
--- a/en/modules/installation-and-upgrade/pages/container-deployment/mlm/migrations/server/server-mlm-51-52.adoc
+++ b/en/modules/installation-and-upgrade/pages/container-deployment/mlm/migrations/server/server-mlm-51-52.adoc
@@ -261,13 +261,13 @@ endif::[]
 
 === Recover from a failed upgrade
 
-If the database migration fails during mgradm upgrade (for example due to "No space left on device", or an unexpected container exit), the original {postgresql} database should remain untouched and safe. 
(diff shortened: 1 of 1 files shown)


0rnela:
Once the procedure is approved, it will be converted to a snippet.

Here is what it says / looks like to begin with.

Screenshot from 2026-07-28 18-38-29

0rnela:
@cbosdo - I believe this is Hexagon's expertise.  I'd appreciate feedback when you get a chance. TIA.

mbussolotto (commented):
instead of this, I'd use `mgradm backup restore`. 
cc @aaannz

aaannz:
IMO this is assuming there is no prior backup. This is how to recover from failed upgrade.

But I don't know from where the reporter get `var-pgsql-migration-target`. This looks like AI generated snipped which is not correct.

mbussolotto:
oh , right, I thought `var-pgsql-walbackup` was somehow renamed in a refactoring :D

0rnela:
This will be addressed with separate PR where directories _data_old* will be documented.

PR #5177, opened 2026-08-06
master - Clarify availability of Salt's "virt" module in the Salt Bundle

KucharczykL:
# Description

This PR documents the availability of Salt's "virt" module inside the Salt Bundle (`venv-salt-minion`) and the required setup steps.

Specifically:
- Updates `contact-methods-saltbundle.adoc` to add a new section clarifying that the Salt `virt` module is not available inside the Salt Bundle by default, and explains the options (using classic `salt-minion` or installing `libvirt-libs` and `libvirt-python` using pip).
- Updates `system-types.adoc` and `vhm.adoc` to add `[NOTE]` blocks pointing to the Salt Bundle considerations when physical virtualization hosts are managed with `venv-salt-minion`.

# Target branches

- master (this PR)
- manager-5.2 https://github.com/uyuni-project/uyuni-docs/pull/5178
- manager-5.1 https://github.com/uyuni-project/uyuni-docs/pull/5179

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/31166
- Addresses Bugzilla issue https://bugzilla.suse.com/show_bug.cgi?id=1270694

### DIFF
--- a/en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
+++ b/en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
@@ -1,6 +1,6 @@
 [[contact-methods-saltbundle]]
 = Salt Bundle
-:revdate: 2025-07-14
+:revdate: 2026-08-06
 :page-revdate: {revdate}
 
 == What is Salt Bundle?
@@ -52,7 +52,8 @@ The {salt} state `util.mgr_switch_to_venv_minion` is available to switch from `s
 It is recommended to switch to `venv-salt-minion` in two steps to avoid trouble with shifting processes:
 
 .Procedure: Switching with `util.mgr_switch_to_venv_minion` state to `venv-salt-minion`
-
+[role=procedure]
+_____
 . Apply `util.mgr_switch_to_venv_minion` with no pillar specified first.
   This will result in the switch to `venv-salt-minion` with copying configuration files etc.
   It will not clean up the original `salt-minion` configurations and its packages.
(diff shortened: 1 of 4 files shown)


0rnela (approved):

en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -132,3 +135,26 @@ Environment=VENV_PIP_TARGET=/new/path/local/venv-salt-minion/pip
...
+[role=procedure]
+_____
+. Install the `libvirt-libs` package on the managed host system using the system's package manager.
+
+. Install the `libvirt-python` package inside the Salt Bundle environment:
++
+[source,bash]
+----
+salt <minion_id> pip.install libvirt-python
+----
```
meaksh, line 159: Hmmm, I think this is not really correct.

The classic Salt package is only available for some OSes, and even if you use the classic `salt-minion`, Uyuni (and MLM) will migrate the system to use the `venv-salt-minion` package automatically via highstate. So using classic salt-minion package is not really an option here.

Then, on SUSE & openSUSE systems, the `libvirt-python` library is already included in the Salt Bundle, so the user simply needs to install the `libvirt-libs` package in the system and that's it.

For non-SUSE & non-openSUSE systems, the `libvirt-libs` package is still required in the system, but in this case, the user would also need to extend their Salt Bundle to include the `libvirt-python` library inside of it. That is the case for calling `pip.install libvirt-python`.

I think we should adjust this. Thank in advance!

PR #5188, opened 2026-08-10
master - Clarify Salt virt module requirements on SUSE and non-SUSE hosts

KucharczykL:
# Description

This PR refines and corrects the documentation for Salt's "virt" module availability inside the Salt Bundle (`venv-salt-minion`) based on PR review feedback.

Specifically:
- Clarifies that using the classic `salt-minion` package is not a supported workaround because the systems are automatically migrated back to the Salt Bundle (`venv-salt-minion`) during highstate.
- Explains that on SUSE Linux Enterprise and openSUSE systems, `libvirt-python` is already pre-installed inside the Salt Bundle, so only the `libvirt-libs` system library needs to be manually installed on the host.
- Explains that on other operating systems (such as RHEL or Debian-based systems), users must install both `libvirt-libs` on the host and `libvirt-python` inside the Salt Bundle using `pip`.
- Updates `CHANGELOG.md` and page revision dates.

# Target branches

- master (this PR)
- manager-5.2 https://github.com/uyuni-project/uyuni-docs/pull/5189
- manager-5.1 https://github.com/uyuni-project/uyuni-docs/pull/5190

# Links
- Tracks issue https://github.com/SUSE/spacewalk/issues/31166
- Addresses Bugzilla issue https://bugzilla.suse.com/show_bug.cgi?id=1270694

### DIFF
--- a/en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
+++ b/en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
@@ -140,16 +140,31 @@ To ensure that such packages are available and functional after an update, it is
 [[contact-methods-saltbundle-virt]]
 == Availability of Salt's "virt" module inside the Salt Bundle
 
-The Salt `virt` module is not available by default inside the Salt Bundle (`venv-salt-minion`) because it requires Python `libvirt` bindings, which are not included in the bundle.
+The Salt `virt` module is not available by default inside the Salt Bundle (`venv-salt-minion`) because it requires system `libvirt` library dependencies, which are not included in the bundle.
 
-If you need to use the `virt` module (for example, on physical virtual hosts or hypervisors like KVM), you have two options:
+[NOTE]
+====
+Using the classic `salt-minion` package is not a supported workaround. If you use the classic `salt-minion`, {productname} will automatically migrate the system back to the Salt Bundle (`venv-salt-minion`) during highstate application.
+====
 
-* Install the classic `salt-minion` package instead of the Salt Bundle.
-* Or, if you want to keep using the Salt Bundle (`venv-salt-minion`), you must manually install the `libvirt-libs` system library on the managed host system, and then install the `libvirt-python` package in the `venv-salt-minion` environment using `pip`:
+To use the Salt `virt` module (for example, on physical virtual hosts or hypervisors like KVM) with the Salt Bundle, follow the instructions for your operating system:
+
+* **SUSE Linux Enterprise and openSUSE systems**:
++
+The `libvirt-python` package is already pre-installed inside the Salt Bundle on these systems. To enable the module, you only need to install the system library:
 +
 [role=procedure]
 _____
 . Install the `libvirt-libs` package on the managed host system using the system's package manager.
+_____
+
+* **Other operating systems (such as Red Hat Enterprise Linux or Debian-based systems)**:
++
+You must manually install the system library and then extend the Salt Bundle to include the Python bindings:
++
+[role=procedure]
+_____
+. Install the `libvirt-libs` (or equivalent system library package) on the managed host system using the system's package manager.
 
 . Install the `libvirt-python` package inside the Salt Bundle environment:
 +
--- a/en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
+++ b/en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
@@ -140,11 +140,11 @@ To ensure that such packages are available and functional after an update, it is
 [[contact-methods-saltbundle-virt]]
 == Availability of Salt's "virt" module inside the Salt Bundle
 
-The Salt `virt` module is available within the Salt Bundle (`venv-salt-minion`) on SUSE-based operating systems (SUSE Linux Enterprise Server, SUSE Linux Enterprise Micro, openSUSE Leap, and openSUSE Tumbleweed). To use the module, you must install the system `libvirt` library dependencies, which are not included in the bundle by default.
+The Salt `virt` module is available within the Salt Bundle (`venv-salt-minion`) on SUSE-based operating systems ({sles}, {slemicro}, {opensuse} Leap, and {opensuse} Tumbleweed). To use the module, you must install the system `libvirt` library dependencies, which are not included in the bundle by default.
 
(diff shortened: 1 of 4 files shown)


KucharczykL:
_home_lukas_work_git_uyuni-docs_build_en_docs_client-configuration_contact-methods-saltbundle html

en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -140,16 +140,31 @@ To ensure that such packages are available and functional after an update, it is
...
 == Availability of Salt's "virt" module inside the Salt Bundle
 
-The Salt `virt` module is not available by default inside the Salt Bundle (`venv-salt-minion`) because it requires Python `libvirt` bindings, which are not included in the bundle.
+The Salt `virt` module is not available by default inside the Salt Bundle (`venv-salt-minion`) because it requires system `libvirt` library dependencies, which are not included in the bundle.
 
-If you need to use the `virt` module (for example, on physical virtual hosts or hypervisors like KVM), you have two options:
+[NOTE]
+====
+Using the classic `salt-minion` package is not a supported workaround. If you use the classic `salt-minion`, {productname} will automatically migrate the system back to the Salt Bundle (`venv-salt-minion`) during highstate application.
+====
```
vzhestkov: This note looks redundant as the salt bundle is the only recommended and default now.

en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -140,16 +140,31 @@ To ensure that such packages are available and functional after an update, it is
...
 
-* Install the classic `salt-minion` package instead of the Salt Bundle.
-* Or, if you want to keep using the Salt Bundle (`venv-salt-minion`), you must manually install the `libvirt-libs` system library on the managed host system, and then install the `libvirt-python` package in the `venv-salt-minion` environment using `pip`:
+To use the Salt `virt` module (for example, on physical virtual hosts or hypervisors like KVM) with the Salt Bundle, follow the instructions for your operating system:
+
+* **SUSE Linux Enterprise and openSUSE systems**:
++
+The `libvirt-python` package is already pre-installed inside the Salt Bundle on these systems. To enable the module, you only need to install the system library:
```
vzhestkov: I don't really like wording of this part: `libvirt-python` package is already pre-installed
It's delivered with `venv-salt-minion` package, but there is no `libvirt-python` as a package in this case. Maybe better just drop this part:
```
The `libvirt-python` package is already pre-installed inside the Salt Bundle on these systems.
```

en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -140,16 +140,31 @@ To ensure that such packages are available and functional after an update, it is
...
 . Install the `libvirt-libs` package on the managed host system using the system's package manager.
+_____
+
+* **Other operating systems (such as Red Hat Enterprise Linux or Debian-based systems)**:
++
+You must manually install the system library and then extend the Salt Bundle to include the Python bindings:
++
+[role=procedure]
+_____
+. Install the `libvirt-libs` (or equivalent system library package) on the managed host system using the system's package manager.
```
vzhestkov: It's not true at all. We are shipping `libvirt` python module with the Salt Bundle for SUSE based distros only: SLES, SL Micro, openSUSE Leap and TW, but not for any other distros. And it's not that easy to extend the Salt Bundle to include `libvirt` python for specified OS as the module is strictly bind to the specific `libvirt-libs` binaries.

vzhestkov (changes requested):
Sorry for being so nitpicky, but I put some comments in the PR. The real problem with `libvirt` that it's not that clear which part of the implementation actually we are talking about as they called almost equally with python modules and libs on the OS, so yes it's also confusing for us, and not just for the end users, so we need to be careful with the wording.

KucharczykL:
@vzhestkov I have updated the PR, please take another look when you have the time!

0rnela (approved):

en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
```diff
@@ -140,21 +140,13 @@ To ensure that such packages are available and functional after an update, it is
 [[contact-methods-saltbundle-virt]]
 == Availability of Salt's "virt" module inside the Salt Bundle
 
-The Salt `virt` module is not available by default inside the Salt Bundle (`venv-salt-minion`) because it requires Python `libvirt` bindings, which are not included in the bundle.
+The Salt `virt` module is available within the Salt Bundle (`venv-salt-minion`) on SUSE-based operating systems ({sles}, {slemicro}, {opensuse} Leap, and {opensuse} Tumbleweed). To use the module, you must install the system `libvirt` library dependencies, which are not included in the bundle by default.
```
vzhestkov: `which are not included in the bundle by default.` - `libvirt` by its nature is very dependant on the system itself (the type, version etc.), and it's the reason we have to build our own `libvirt` python module to include it to the bundle (for a limited list of the clients, originally only for the clients where we offer VMs management with KVM, this feature was deprecated and dropped with 5.1). This end of the phrase is a bit eye scratchy for me by saying `by default` because we are only including binary linked dependencies to the bundle if there are no any other options to make core features working with the bundle, which is not fully true with this particular feature as now virt management is not really required and only used for detection if the VM running on the virtual manager host is also registered to the MLM/Uyuni, what is not needed by most of the customers (most probably). I'm not sure we have any customer who need it, most of the customers just complain that they see errors on trying to get something with `virt` salt module from the system. In this particular point for me it's better just to drop `by default` as it was before, because in this case we have no any other options.

vzhestkov (approved):
I left one big comment to the one particular phrase, but in general it looks good to me now, except repeating `must` in some parts, even if it's true it sounds like we are pushing to do it, maybe makes sense to make it bit softer, like: `If this functionality is required on the managed client, libvirt-libs package should be installed.` (BTW, in case of KVM with libvirt there is no chances to not to have it installed, so it should be present on any SUSE based KVM host)

en/modules/client-configuration/pages/contact-methods-saltbundle.adoc
KucharczykL: Done in https://github.com/uyuni-project/uyuni-docs/commit/04cda430fae3b7b257be7bb53c158ff8c0356c9a

PR #5216, opened 2026-08-24
master - Extract and relocate general sudo onboarding guidelines (bsc#1275794)

KucharczykL:
# Description

This PR extracts and relocates the duplicate passwordless `sudo` configuration guidelines out of OS-specific sections and into a central, global documentation page (`registration-sudo.adoc`).

Specifically:
- Adds a new global page `registration-sudo.adoc` under the "Client Registration > Registration Methods" hierarchy containing the single source of truth for `sudo` and `sudoers` setup.
- Updates the client pages for Ubuntu, Debian, Raspberry Pi OS, and SLE (supporting SLES 16/Agama) to point to the central page.

# Target branches

- master (this PR)
- manager-5.2 https://github.com/uyuni-project/uyuni-docs/pull/5217
- manager-5.1 https://github.com/uyuni-project/uyuni-docs/pull/5218

# Links
- This PR tracks issue https://github.com/SUSE/spacewalk/issues/31776
- Addresses Bugzilla issue https://bugzilla.suse.com/show_bug.cgi?id=1275794


0rnela (approved):

0rnela:
@KucharczykL - You also need to run script update-cfg-files in l10n-weblate and add to this PR before merging it.

0rnela (changes requested):
@KucharczykL - please run script update-cfg-files in l10n-weblate and add to this PR.

KucharczykL:
> @KucharczykL - You also need to run script update-cfg-files in l10n-weblate and add to this PR before merging it.

The script doesn't work for me in any of the branches. Maybe some toolchain changes?

0rnela:
I'm not aware of any I'm afraid.

en/modules/client-configuration/pages/registration-sudo.adoc
```diff
@@ -0,0 +1,40 @@
...
+_____
+. On the client, edit the `sudoers` file by running:
++
+----
+sudo visudo
+----
+. At the end of the file, add the following line. Replace `<user>` with the name of the user that is bootstrapping the client:
++
+----
+<user>  ALL=NOPASSWD: /usr/bin/python, /usr/bin/python2, /usr/bin/python3, /var/tmp/venv-salt-minion/bin/python
```
vzhestkov: In general looks good to me, but this line contains the entries which are not true anymore. Basically we could left just `/var/tmp/venv-salt-minion/bin/python` in the list as the Salt Bundle is the only supported method with Salt SSH now. For `SLE 15` maybe it makes some sense to left the note that `/usr/bin/python3` could be still used for very specific cases, but I would prefer to drop everything except the python from the bundle.

vzhestkov (approved):
Just one small correction inline, which maybe makes sense to include also.

And maybe it also makes sense to put the ref to the new section to all the supported clients, now it looks like Debian based and SLE were the only updated, but it should be true to all the clients we support.

PR #5224, opened 2026-08-28
master - Fix update-cfg-files when run outside the l10n-weblate directory

KucharczykL:
# Description

This PR fixes `l10n-weblate/update-cfg-files`, which failed with exit code 2 whenever it was invoked from anywhere other than the `l10n-weblate/` directory itself.

The `for I in *.cfg` glob expanded in the caller's working directory rather than the script directory. Run from the repository root it matched nothing, so `basename '*.cfg' .cfg` yielded `*`, which the enclosing `for` list then re-expanded to every entry in the repository root:

```
$ ./l10n-weblate/update-cfg-files
grep: l10n-weblate/CHANGELOG.md.cfg: No such file or directory
grep: l10n-weblate/Dockerfile.bci.cfg: No such file or directory
grep: l10n-weblate/go.mod.cfg: No such file or directory
...
```

It also left roughly 27 stray `l10n-weblate/*.cfg.tmp` files behind.

Specifically:

- `l10n-weblate/update-cfg-files`: anchor the working directory to `$CURRENT_DIR` before the loop, so the `*.cfg` glob always expands inside `l10n-weblate/`. Everything else in the script was already anchored via `BASH_SOURCE`.
- `l10n-weblate/update-cfg-files`: the header comment now says the script can be run from any directory.
- `CHANGELOG.md`: one entry for the fix.

Verified on a clean `master` checkout — running the script from the repository root, from `l10n-weblate/`, and from an unrelated directory each exits 0, leaves no `.cfg.tmp` files, and produces no diff in the `.cfg` files. Adding a throwaway `.adoc` page under `en/modules/client-configuration/pages/` and re-running from the repository root correctly appends its `[type: asciidoc]` entry, which was not possible before.

No change to any `.cfg` content is included here.

# Target branches

- master (this PR)
- manager-5.2 — not affected for now; the script is byte-identical there and carries the same bug, but holding off pending confirmation that the Weblate workflow is still in use
- manager-5.1 — same as above

`manager-5.0` and `manager-4.3` carry the same bug in the pre-migration variant of the script (`modules/` instead of `en/modules/`) and are out of scope.

# Links

- Noticed while responding to review feedback on https://github.com/uyuni-project/uyuni-docs/pull/5216, which asked for `update-cfg-files` to be run before merging.


jcayouette:
I do not think we are using weblate on master, and for 5.2, 5.1 we are using AI translations and are no longer using the weblate directory. We feed English to Vistatek and they work their magic then submit the fork back to use with our translated content directly.

KucharczykL:
> I do not think we are using weblate on master, and for 5.2, 5.1 we are using AI translations and are no longer using the weblate directory. We feed English to Vistatek and they work their magic then submit the fork back to use with our translated content directly.

@jcayouette Should the script be removed then? Ornela suggested me to run it in one of my PRs recently.

KucharczykL:
I looked at the current state - 5.2 is still on Weblate, at least `update_translation_files` runs every push. master keeps failing because of the protected branch stuff.

So I'd say I'll merge this one and we can move on to other things, which might remove it later but at least the footgun will be gone.

PR #5239, opened 2026-08-31
master - Add CI check enforcing revdate on changed AsciiDoc pages

KucharczykL:
# Description

Makes it possible to check if revdates have been updated properly:
- CI check that works only on files edited in a particular PR
- new Taskfile targets: `validate:revdate` (either using uv or regular Python) and `container:validate:revdate`

The actual checks are: revdate well-formedness, lacking a paired page-revdate and revdate age (default is 7 days).

# Target branches

- master (this PR)

# Links
- This PR supersedes https://github.com/uyuni-project/uyuni-docs/pull/4906


KucharczykL:
Some context for reviewers on what this will feel like in practice.

The check only looks at pages a PR actually changes, but if you run it across the whole repo today it reports 525 of 565 pages as stale. That is not a bug in the check: every one of those 525 is a staleness error, and there are zero missing or malformed revdates left after this PR, so the backlog is entirely "the revdate was never bumped when the page was touched".

Almost all of it comes from two mass mechanical commits, #4918 (toolchain migration) and #5052 (backtick sweep), which rewrote hundreds of pages without touching their revdates. The grace period is seven days, so a normal edit-and-bump workflow never trips it.

The practical effect is that the first person to edit one of those pages will be asked to bump its revdate to the day they edited it, which is what they should be doing anyway. The debt gets paid off one page at a time by the people already in the file, rather than in a single sweep that would set 525 pages to the same meaningless date.

Two related things I found but deliberately left alone, in case someone wants to pick them up:

- `en/modules/installation-and-upgrade/partials/snippet-database-backup-volume.adoc` is a partial that has both a level-1 title and a `:revdate:`. An attribute entry inside an included file overrides the same attribute in the including page, so that revdate silently wins over the revdate of whichever page includes it.
- All eight `_attributes.adoc` still set `:partialsdir: {moduledir}/pages/_partials`, which no longer exists. Harmless today because nothing references `{partialsdir}`.

